PR #299, opened 2015-01-09
Define identity of a web app (closes #272)

### LINKED ISSUE #272: Define identity of a web app.
* what identifies an app? An origin?  
 * how does one update the app? 
 * what happens if the scope changes?

marcoscaceres:


marcoscaceres:
@kenchris I changed the identity to be both the start URL and the manifest URL.

PR #302, opened 2015-01-14
Clarified opening deep link (closes #301)

### LINKED ISSUE #301: Clarify launching from a deep link.
The spec currently says:

> When an application context is created, the user agent MUST immediately navigate to the start URL with replacement enabled.

This doesn't handle the case where an app needs to open to a URL that is within its scope (i.e., to handle "deep linking").

marcoscaceres:


marcoscaceres:
@benfrancis, better?

benfrancis:
Looks good to me.

kenchris:
How are we handling conflicts? Or old apps with out of date scope

marcoscaceres:
No, someone needs to implement the spec as is and work out the kinks. I don't want use to be theorizing here about how this is all going to work for it to all fail spectacularly when someone tries to implement this.

PR #311, opened 2015-02-13
starring is giving/being a star, staring is looking at something intently

hillbrad:


marcoscaceres:
:star: your a star, @hillbrad :star:

marcoscaceres:
Ah crap, *you're*... I give up.

hillbrad:
:)

Issue #312, opened 2015-02-13
"manifest" link relation type should be registered per RFC 5988


dret:
It would be great to see the new link relation type be registered with IANA. If that helps, I'd be more than happy to add that to the "IANA Considerations" section.

marcoscaceres:
On Friday, February 13, 2015, Erik Wilde <notifications@github.com> wrote:

> It would be great to see the new link relation type be registered with
> IANA. If that helps, I'd be more than happy to add that to the "IANA
> Considerations" section.
>

That would be great! Thanks for your help.

dret:
ok, will do sometime next week, i guess. it'll simply take the standard registration template from RFC 5988 and add a section to the "IANA Considerations". the result will look similar to http://tools.ietf.org/html/rfc6892#section-4

marcoscaceres:
Thanks. If you find you are too busy, I'm happy to give it a go and can ping you to check over the PR.

marcoscaceres:
Sent request to IANA too.

marcoscaceres:
(well, to ietf)

PR #314, opened 2015-02-15
adding link relation registration with IANA

dret:
probably needs some wording improvements, but a first stab at resolving https://github.com/w3c/manifest/issues/312


marcoscaceres:
@dret, this is great. Thank you! There are a few minor things, but I'll just fix them locally.

PR #316, opened 2015-02-16
Removed JSON-LD text (closes #313)

### LINKED ISSUE #313: JSON-LD view
just out of curiosity: it seems that the processing model for the manifest data at least sometimes requires a sequence of items to be represented (as opposed to a set). at least this is how i read how icons must be processed (http://www.w3.org/TR/appmanifest/#icons-member). i am wondering if this can even work for the JSON-LD "view" that's proposed later on. RDF is notoriously bad with sequences, and afaict the example shown would lose the sequence that is necessary to correctly implement the algorithm defined in the spec. am i reading it wrong or missing something?

marcoscaceres:


kenchris:
i like it

PR #344, opened 2015-03-26
Add related_applications and prefer_related_applications.

mounirlamouri:
Fixes issue 326.

### DIFF
--- a/index.html
+++ b/index.html
@@ -315,7 +315,7 @@
         <li>Otherwise, return <code>false</code>.
         </li>
       </ol>
-      <div class="issue" title="🐒 Monkey patch">
+      <div class="issue" title="&#128018; Monkey patch">
         <p>
           Enforcing the navigation scope depends on [[!HTML]]'s navigate
           algorithm. As such, the following algorithm monkey patches [[!HTML]].
@@ -1638,16 +1638,18 @@ Content-Type: application/manifest+json
           <code title="">related_applications</code> member
         </h3>
         <p>
+          A <dfn>related application</dfn> is an application accessible to the
+          underlying application platform that has a relationship with the
+          web application <a>associated with a manifest</a>.
+        <p>
           The <dfn id=
           "member-related_applications"><code>related_applications</code></dfn>
-          member lists all the applications related to the web application.
-        </p>
-        <p>
-          The <code>related_applications</code> member is as an indication of a
-          relationship of the web application and the listed applications. This
-          relationship is unidirectional and unless a listed application claims
-          the same relationship, the <a>user agent</a> MUST NOT assume a
-          bi-directional endorsement.
+          member lists <a title="related application">related applications</a>
+          and serves as an indication of such a relationship between web
+          application and <a>related application</a>s. This relationship is
+          unidirectional and unless a listed application claims the same
+          relationship, the <a>user agent</a> MUST NOT assume a bi-directional
+          endorsement.
         </p>
         <p>
           Example of usages of the <code>related_applications</code> could be a
@@ -1660,7 +1662,7 @@ Content-Type: application/manifest+json
           The <dfn>steps for processing the <code>related_applications</code>
           member</dfn> are given by the following algorithm. The algorithm
           takes a manifest as an argument. This algorithm returns a list of
-          applications objects <var>applications</var>, which can be empty.
+          application objects <var>applications</var>, which can be empty.
         </p>
         <ol>
           <li>Let <var>applications</var> be an empty list.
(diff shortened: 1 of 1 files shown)


benfrancis:
I believe that you have the best of intentions with this proposal, but in my opinion it represents a very real risk to the World Wide Web as we know it. This could very easily be used by some implementers to significantly contribute to the fragmentation of the web into multiple parallel walled gardens of content for their own commercial gains, putting us back decades.

I'm struggling to understand how adding advertisements for web-alternative software applications and the commercial stores they can be acquired from to the "manifest for web application" specification falls within the charter of the Web Applications Working Group [1], or the requirements of this specification [2].

1. http://www.w3.org/2014/06/webapps-charter.html
2. http://w3c-webmob.github.io/installable-webapps/#requirements

anssiko:
@mounirlamouri Did a quick review. Here's a PR with some proposals with `mounirlamouri:related_applications` as the base: https://github.com/mounirlamouri/manifest/pull/1

@benfrancis Do you think you could propose some text to be added to the spec around this feature to address your concern, so that authors would be informed of correct usage? I think this is an attempt to address the very real [slam door problem](https://github.com/slightlyoff/AppInstallImprovements/blob/master/explainer.md#offering-related-applications). Assuming the Web Platform will be competitive with native in the near future this feature would actually not be needed for future web applications. However, today, the "slam doors" are a real problem that drive people away from the Web. This is an attempt to mitigate that. That said I see your concerns re misuse.

benfrancis:
@anssiko wrote:
> Do you think you could propose some text to be added to the spec around this feature to address your concern, so that authors would be informed of correct usage?

I think one of my main concerns is eroding user choice. You could take the optimistic view that this will help rid the web of door-slam advertisements for web-alternative apps, and that in order to use this feature developers must also add a web app manifest to their web site, thereby helping increase manifest usage. But these benefits are negated if the user agent then parses the boolean prefer_related_applications property and decides to show the user only an alternative application, not the web application.

A UX designer of a user agent which implements the specification may reasonably conclude that offering the user a choice would be confusing, so design a UI which only offers the option preferred by the developer. A side effect of this would be the gradual fragmentation of the web, without users even realising that they're not actually using the web, but a parallel web-alternative system.

Without the related_applications property users have the choice whether to install the web app from the browser, or to go to an app store to install a non-web alternative. With the property, the choice may be put in the hands of the app developer who can force them to use the version of the app which gives them the highest financial return.

I'm not sure what changes to the wording could help prevent this from happening?

skddc:
> With the property, the choice may be put in the hands of the app developer who can force them to use the version of the app which gives them the highest financial return.

This is actually more of an issue where the vendor has incentive, because as an app provider/developer, the 30% App Store fee can really hurt and be a worse option for them than handling payment themselves. However, they'll still want to use the related apps, because they have to adhere to what the user wants, even if they don't prefer it themselves. At that point the vendor has an incentive to draw the user to their App Store instead of the app provider's Web option.

benfrancis:
Exactly. The slam-door advertisements are a result of the fact that users have a choice, they're a crude attempt at influencing that choice. If getting rid of those ads also means getting rid of the choice, then I don't think the web, developers or users have necessarily benefited at all.

anssiko:
We cannot prevent the user agent implementers from doing such bad UI/UX decisions. However, what we can do spec-wise is to provide non-normative "Implementation Considerations". Imposing detailed normative requirements on the UI has in practice proven to be hard (or impossible), see http://www.w3.org/TR/wsc-ui/ We cannot anticipate all the possible UIs, form factors, input methods, modalities and so forth.

Another way to look at this is to let the market decide, i.e. believe the users will gravitate toward an implementation that works for them the best. If their browser discriminates the Web that they'd prefer, users switch to a competing browser.

All this assuming there are actually multiple implementers who are interested in this feature, since otherwise this would not make it through the interop phase, and would be better spec'd as a proprietary extension.

marcoscaceres:
@mounirlamouri ready to push green button?

marcoscaceres:
@mounirlamouri just realized you are missing when to process these members as part of the "steps for processing a manifest". It should be:
```HTML
<li>
Let <var>related applications</var> of parsed manifest be the result of running the <a>steps for processing the <code>related_applications</code> member</a> with manifest as the argument. 
</li>
```
And same for the other new member...

mounirlamouri:
@anssiko I applied your comments (merged your PR).
@marcoscaceres thanks for the reminder. Fixed.

@benfrancis I head your and understand your concerns but we are not doing anything crazy here: web pages are already implementing door slams. Some web sites will not even allow users to use their mobile website. I do not think adding fields in the Manifest is going to make this worse. The only risk is that UA would aggressively suggest installing native apps instead of visiting the websites but there are already proprietary ways to advertize these apps and I do not think UAs have been doing that. Also, the Manifest clearly want that advertizment to be the non-default behaviour.

PR #349, opened 2015-04-02
Changed file extension of web manifest (closes #346)

### LINKED ISSUE #346: The use of the `.manifest` file extension may be problematic / confusing ?
From: http://w3c.github.io/manifest/#h-media-type-registration

```
File extension(s):
    .json, .manifest
```

`.manifest` was the recommended file extension for the [Application Cache manifest file](https://html.spec.whatwg.org/multipage/browsers.html#manifests), and then, it was changed to  `.appcache` in order to [_"avoid clashing with Microsoft's unregistered `application/manifest` type"_](https://html5.org/r/5812). 

So, taking into consideration also the [legacy](https://www.google.com/search?hl=en&q=%22addtype+text%2Fcache-manifest+manifest%22) aspects, recommending `.manifest` as the file extension creates even more "clashing", and this can complicate things when it comes to server configs (e.g.: especially as in some of the cases, things like the media type [matter](https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache#Referencing_a_cache_manifest_file)).

__Note:__ As opposed to `.manifest`, in the case of `.json`, IMHO, things are better. Since for the manifest files for web apps the media type doesn't seem to matter (for now at least), and they are just JSON files, the configurations that most servers out there have should already work great for them (e.g.: serve them compressed, with the media type [`charset` parameter](http://www.w3.org/International/O-HTTP-charset#charset) set to `utf-8`, etc.).

marcoscaceres:
cc @alrra

### DIFF
--- a/index.html
+++ b/index.html
@@ -2275,7 +2275,7 @@ Content-Security-Policy: img-src icons.example.com;
                 File extension(s):
               </dt>
               <dd>
-                .json, .manifest
+                .json, .webmanifest
               </dd>
               <dt>
                 Macintosh file type code(s):


alrra:
:+1:

PR #351, opened 2015-04-06
Made using fallbacks normative (closes #350)

### LINKED ISSUE #350: Add meta tags equivs. for features were possible
I've implemented manifest on app.ft.com (it's not live yet), and my feedback is:

* The basic properties we needed (name and start URL) are redundant with numerous `<meta>` and `<link>` elements we have already (including proprietary ones such as `apple-mobile-web-app-title` and `msapplication-starturl`).

* JSON is unpleasant to work with. It's unsuitable for manual authoring due to lack of comments and draconian XML-like error handling, but having to use a serializer for just a couple of properties feels silly, especially that we've had code to serialize metadata already — in HTML.

* external file feels unnecessary given how tiny it is.

* Adding a few more `<meta>` elements is *not a problem at all* for us, because when they're surrounded by very similar markup they gzip to almost zero. We already emit `<meta>` and `<link>` elements for "legacy" specs, social media sites, and vendor-specific features.

So overall I'd strongly prefer an option to use `<meta>` and `<link>` equivalents for basic Manifest features, to reduce overhead (both on-the-wire overhead due to inlined gzipped data, and maintenance overhead thanks to reusing existing HTML-meta pipeline instead of creating a new JSON one). As a bonus, they'd be localizable #338.

marcoscaceres:
* Deriving metadata from the manifest's owner Document is now a normative recommendation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -158,7 +158,7 @@
           Example of using a <code>link</code> element to associate a website
           with a <a>manifest</a>. The example also shows how to use [[!HTML]]'s
           <code>link</code> and <code>meta</code> elements to give the web
-          application a fallback name and set of icons.
+          application a fallback short name and set of icons.
         </p>
         <pre class="example highlight html" title="linking to a manifest">
 &lt;!doctype&gt;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
> Should a web application have a different name depending on which page it is installed from? 

It's up to the developer.

PaulKinlan:
I generally feel ok with this as long as we treat the manifest as the authoritative representation and fallback when needed.

marcoscaceres:
@PaulKinlan wrote: 
> I generally feel ok with this as long as we treat the manifest as the authoritative representation and fallback when needed.

That's the plan.

PR #353, opened 2015-04-07
Obtaining a Manifest should follow usual CORS rules with credentials.

mounirlamouri:
Requesting to be same-origin with the document doesn't sound needed. If the Manifest is in another origin and it has some Access Controls set allowing it to be used, I think it should be fine to allow it to be used.

In the wild, we've seen a few websites adding there Manifest to cross origins places (like G+ using gstatic).

### DIFF
--- a/index.html
+++ b/index.html
@@ -771,16 +771,12 @@ if (standalone.matches) {
           <li>If <var>manifest link</var> is <code>null</code>, terminate this
           algorithm.
           </li>
-          <li>If the URL of <var>manifest link</var>'s <code>src</code>
-          attribute is not <a>same origin</a> as its owner
-          <code>Document</code>, then <a>issue a developer warning</a>
-          explaining that the manifest's URL needs to be <a>same origin</a> as
-          the owner <code>Document</code> and terminate this algorithm.
-          </li>
           <li>
-            <a>Obtain a resource</a> as per [[!HTML]] and let
-            <var>response</var> be the [[!FETCH]] <a href=
-            "https://fetch.spec.whatwg.org/#concept-response">response</a>. If
+            <a>Obtain a resource</a> with <var>mode</var> set to "<a href=
+            'https://html.spec.whatwg.org/multipage/infrastructure.html#attr-crossorigin-use-credentials'>Use
+            Credentials</a>" and let <var>response</var> be the [[!FETCH]]
+            <a href=
+            "http://fetch.spec.whatwg.org/#concept-response">response</a>. If
             the <a>obtain a resource</a> algorithm terminates prematurely or
             <var>response</var> is a network error, terminate this algorithm.
           </li>


marcoscaceres:
We are drafting a discussion paper around this issue. Please hold for a bit before merging.

marcoscaceres:
Broke you PR with another PR, sorry.

marcoscaceres:
Early draft of what we (Moz) want to discuss... we are hoping to set up a call about this with implementers.  

https://etherpad.mozilla.org/role-of-manifest

PaulKinlan:
Hmm, ok.  I can certainly talk to the point about the relationship of the manifest location and the page it was discovered on.  

It seems like we are treating the manifest as a complete package that is standalone (i.e, I can read the manifest.json and work out exactly what to do with it) yet there are a huge number of runtime resolutions happening that need the page it was linked from to exist, specifically the start_url.  

My biggest hurdle is that I believe we (the web) need manifest to be able to not be forced to live on the same host as where it was linked from.

benfrancis:
>  there are a huge number of runtime resolutions happening that need the page it was linked from to exist, specifically the start_url.

Can you expand on this a little? Obviously the start_url needs to exist in order to launch the app (that's true regardless of the relationship between the manifest and the page it was installed from). If the manifest is same-origin with the app then the start_url can just be resolved against the manifest URL, there's no need to involve the document it was installed from, which could be one of many URLs.

I would actually assume this resolution would happen at install time when processing the manifest, not at runtime. What other resolutions need to happen, and what prevents them being resolved against the manifest URL?

> My biggest hurdle is that I believe we (the web) need manifest to be able to not be forced to live on the same host as where it was linked from.

Are there more examples you can give of where this is problematic? Just trying to understand because this hasn't been a problem for Firefox OS so far.

jmajnert:
>Obviously the start_url needs to exist in order to launch the app (that's true regardless of the relationship between the manifest and the page it was installed from).

No. As is specced currently, start_url may not exist. The UA can use the page from which the manifest was linked to as the "entry point".

> If the manifest is same-origin with the app then the start_url can just be resolved against the manifest URL, there's no need to involve the document it was installed from, which could be one of many URLs.

Another solution is to allow start_url to be absolute URL.

PaulKinlan:
>  If the manifest is same-origin with the app then the start_url can just be resolved against the manifest URL, there's no need to involve the document it was installed from, which could be one of many URLs.

I get that, but at the same time you could have the full URL if needed and that solves a lot of the issues.

My concern is that we can't deploy this on to CDN's and we have a lot of history with developers for whom deploying to CDN is far far easier than storing files on a server.  We've had issue like this for domain verification on google.com/webmasters it is just a hard experience to get new files deployed on domain.  I can't say the individual partners and developers, but if you look at facebook.com, amazon.com etc everything static is hosted on a CDN.

Other things as well are an issue, I can't be on my http site and link to my manifest on https and have that launch the https version of the site.

benfrancis:
> if you look at facebook.com, amazon.com etc everything static is hosted on a CDN.

Hmm, OK. Facebook do already host a mozApps manifest at https://m.facebook.com/openwebapp/manifest.webapp - I have no idea how much of a challenge that was for them.

Same with:
* Twitter https://mobile.twitter.com/cache/twitter.webapp
* YouTube http://m.youtube.com/mozilla_youtube_webapp
* SoundCloud https://m.soundcloud.com/manifest.webapp

etc...

Overall I think hosting the manifest same-origin is a more scalable solution than the hoops I had to go through to verify my domain for an app I submitted to the Chrome Web Store, but I'm not doubting that it could be a challenge for some hosting setups.

benfrancis:
Let's discuss this in #360

index.html
```diff
@@ -771,16 +771,12 @@
...
-          explaining that the manifest's URL needs to be <a>same origin</a> as
-          the owner <code>Document</code> and terminate this algorithm.
-          </li>
           <li>
-            <a>Obtain a resource</a> as per [[!HTML]] and let
-            <var>response</var> be the [[!FETCH]] <a href=
-            "https://fetch.spec.whatwg.org/#concept-response">response</a>. If
+            <a>Obtain a resource</a> with <var>mode</var> set to "<a href=
+            'https://html.spec.whatwg.org/multipage/infrastructure.html#attr-crossorigin-use-credentials'>Use
+            Credentials</a>" and let <var>response</var> be the [[!FETCH]]
```
marcoscaceres: This doesn't match Fetch spec and references HTML instead.

marcoscaceres:
Rewriting this in terms of fetch.

PR #357, opened 2015-04-07
Fix links to "SchemaStore" GitHub repository

alrra:
--

It seems that the SchemaStore repository was moved from https://github.com/madskristensen/SchemaStore/ to https://github.com/SchemaStore/schemastore.


marcoscaceres:
Thanks once again :)

PR #358, opened 2015-04-08
Convert icon object to image object (close #327)

### LINKED ISSUE #327: Convert "Icon object" into a generic "image object"
The icon object can really just serve as a generic image object. This could then be reused in the future to, for instance, support splash screens.

marcoscaceres:
* this is just "editorial" change - it doesn't change any processing aspects of icons or anything like that.  
* generalizes icon to "image" so we can reuse image objects in other contexts, like for splash screens.


marcoscaceres:
Addressed feedback from @kenchris.

PR #362, opened 2015-04-17
Update the CSP reference; link directives

anssiko:
The `manifest-src`directive is defined in [CSP3], not in [CSP2], so I updated the CSP reference. Also linked the related directive definitions, fixed couple of typos.


marcoscaceres:
:+1: thanks @anssiko.

PR #369, opened 2015-04-27
Note about manifest and crawlers (closes #343)

### LINKED ISSUE #343: Add requirements when the Manifest is used by crawlers
One benefit of the Manifest is that it is discoverable by crawlers but I realise that we do not actually even mention that in the specification or use cases related to it. Also, we list no requirements for crawlers like folliwng robots.txt.

marcoscaceres:


marcoscaceres:
@mounirlamouri please review.

marcoscaceres:
@mounirlamouri, let me know if anything else should be added.

PR #370, opened 2015-04-27
Added background color for image object (close #366)

### LINKED ISSUE #366: image objects should have a background color
Use case: non-square container for icons and for splash-screens.

marcoscaceres:


marcoscaceres:
This can be used for both splash screen and icons. @alxlu, would appreciate your feedback.

PR #387, opened 2015-07-06
Add "content" background_color (closes #390)

### LINKED ISSUE #390: Add background color for splash screen.
Conditional on: https://github.com/w3c/manifest/pull/387

If we remove image object's `background_color`, we need a way of coloring the  backgrounds of splash screens.

marcoscaceres:
Currently, there doesn't seem to be a strong case to have background colors for individual icons. IMO, it would be better to just have a generic `icons_background_color`. If we see demand to override this on a per-icon level later, we should add `background_color` back.  Mozilla currently have no plans to support `background_color` on a per-image basis, but we want to support it on a generic basis as proposed in this PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -149,6 +149,7 @@
         "sizes": "128x128",
         "density": 2
       }],
+  "icons_background_color": "red",
   "splash_screens": [{
         "src": "splash/lowres",
         "sizes": "320x240"
@@ -970,6 +971,11 @@ Content-Type: application/manifest+json
           <var>manifest</var>, <var>manifest URL</var>, and "icons" as
           arguments.
           </li>
+          <li>Let <var>icons background color</var> of <var>parsed
+          manifest</var> be the result of running the <a>steps for processing
+          the <code>icons_background_color</code> member</a> with
+          <var>manifest</var>.
+          </li>
           <li>Let <var>splash screens</var> of <var>parsed manifest</var> be
           the result of running the <a>steps for processing an array of
           images</a> with <var>manifest</var>, <var>manifest URL</var>, and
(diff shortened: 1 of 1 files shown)


marcoscaceres:
[Moz bug 1180652](https://bugzilla.mozilla.org/show_bug.cgi?id=1180652)

benfrancis:
I agree that a background colour per icon is probably excessive, but I'm not sure I understand the difference between "theme_color" and "icons_background_color" after this change.

I think the "mask icon" debacle at WHATWG over Apple's (mis-)use of theme_color has shown that people generally consider "theme_color" as a background colour even though it isn't explicitly specified this way. Perhaps this spec should explicitly use "theme_color" as background colour, or just use "background_color" instead. "theme_color" would be more consistent with HTML.

marcoscaceres:
@benfrancis `theme_color` is used to color parts of the UI, which is different from just coloring in the background on which an icon sits. For example, Pinterest uses white for their MS live tile:

```HTML
<meta name="msapplication-TileColor" content="#ffffff">  
```
But their theme color is a darkish red: 
```HTML
<meta name="theme-color" content="#bd081c"> 
```

It wouldn't make sense, IMO, for them to sit their logo on a red background - because it would get lost as it's already red with a thin white outline. It also wouldn't make sense to use #fffff at the `theme_color`, because their brand color is "#bd081c". 

In short: It's better to have specific control over each design element - rather than forcing one to serve the purpose of another.

benfrancis:
Just FYI, we're planning on using the theme_color behind the icon in multiple places in Firefox OS. In the browser chrome, and as a background colour for "cards" (with an opacity of <1).

See https://wiki.mozilla.org/FirefoxOS/Pin_the_Web

marcoscaceres:
@benfrancis that sounds good, specially as we experiment with what that looks like. The spec says that the it's only a SHOULD that `icons_background_color` be used to cater for cases like the one you describe.

marcoscaceres:
Talked to @mounirlamouri, we agree that `background_color` on image objects complicates things - and neither of us want to implement it. So I will initially remove that. 

@mounirlamouri has an interesting proposal around just having a general "background_color" (which represents an applications background color), but I'll let him pitch it and how it would be used.

PR #389, opened 2015-07-06
added inline issues (closes #386)

### LINKED ISSUE #386: Add inline GH issues
Need to add the outstanding issues inline to the spec. Format is:

```
<div class=issue data-number=26></div>
<section id=issue-summary></section>
```

marcoscaceres:


kenchris:
Lgtm from Cascais :-)

marcoscaceres:
You BASTAAAARDD!! :sun_with_face: :bikini: :swimmer: :cocktail:

benfrancis:
Is it a common practice to put comments from issues inline in a spec? It seems like this is going to be a lot of work to maintain as the discussions develop. Just wondering if there's a way of doing this which doesn't create so much work for you as an editor @marcosc ?

marcoscaceres:
@benfrancis if it turns into a bit a of nightmare, then we can pull them out. Right now, there only a few.

PR #391, opened 2015-07-13
Remove background color on image object

marcoscaceres:
Reached agreement with @mounirlamouri on IRC to remove this. We will figure out a different approach for splash screen.  @kenchris, you approve?


kenchris:
I approve

PR #393, opened 2015-07-31
Fixed up titles and dfns

marcoscaceres:
Removes Respec warnings about abusing title attributes. Cleans up markup. No normative changes.


marcoscaceres:
making some fixes to respec to make this better and remove some redundant `data-lt` values.

PR #396, opened 2015-08-14
Add background_color member (close #394).

### LINKED ISSUE #394: Add background_color to represent the content's background color
I would like the Manifest to have a ```background_color``` field so a website could specify its background colour. The expected value is similar to ```theme_color```.

The use case for this is to be able to create a view for a web app with the right color before the app is loaded. For example, on a mobile device, if you add to your homescreen a website, it might take some time before it renders, in which case, a plain white screen wouldn't be ideal. This is different from a splashscreen because if a website is working offline, there might still be a delay between the user clicking and the website being rendered because of starting new processes and loading the data from the disk. In such situation, a splashscreen would be too heavy but showing the backrgound color early would be a smooth transition.

mounirlamouri:
@kenchris @marcoscaceres please take a look.

### DIFF
--- a/index.html
+++ b/index.html
@@ -164,7 +164,8 @@
   "start_url": "/racer/start.html",
   "display": "fullscreen",
   "orientation": "landscape",
-  "theme_color": "aliceblue"
+  "theme_color": "aliceblue",
+  "background_color": "red"
 }
 </pre>
       </section>
@@ -996,6 +997,11 @@ Content-Type: application/manifest+json
           <code>theme_color</code> member</a> with <var>manifest</var> as
           argument.
           </li>
+          <li>Let <var>background_color</var> of <var>parsed manifest</var> be
+          the result of running the <a>steps of processing the
+          <code>backrgound_color</code> member</a> with <var>manifest</var> as
+          argument.
+          </li>
           <li>Return <var>parsed manifest</var>.
           </li>
         </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1824,6 +1830,67 @@ <h3 id="applying">
...
       </section>
+      <section>
+        <h3>
+          <code title="">background_color</code> member
+        </h3>
+        <p>
+          The <dfn id=
+          "member-background_color"><code>background_color</code></dfn> member
+          describes the expected background color of the web application. It
+          will repeat what is already available in the application stylesheet
```
marcoscaceres: "will repeat" sounds like the UA should do something. This might confused devs. Just say "repeats".

index.html
```diff
@@ -1824,6 +1830,67 @@ <h3 id="applying">
...
+          but can be used by the <a>user agent</a> to draw the background color
+          of a web application for which the manifest is know before the files
+          are actually available, whether they are fetched from the network or
+          retrieved from the disk.
+        </p>
+        <p>
+          The <code>background_color</code> member is only meant to improve the
+          user experience while a web application is loading and should not be
```
marcoscaceres: "should not" needs to be "MUST NOT"

marcoscaceres:
r+ from me, bar the nits :) Thanks @mounirlamouri for drafting the text!

mounirlamouri:
Thanks for the review!

PR #403, opened 2015-09-19
Remove feature at risk for theme_color (closes #356)

### LINKED ISSUE #356: What's the use case for theme_color?
```theme_color``` in the Manifest sounds fairly odd. Discussing with @PaulKinlan and @marcoscaceres, my understanding is that it was meant as a defaul theme_color for the meta tag. However, this is not possible because the UA might want to have a theme_color value for the first paint but Manifest can't block the first paint.

I would understand that it could be used for UI-related features but I would prefer to avoid keeping things in the Manifest "in case of". Is there interest to use theme_color by an implementation?

marcoscaceres:


marcoscaceres:
@kenchris r?

marcoscaceres:
@kenchris r?

wizzwizz4:
:+1:

PR #406, opened 2015-10-07
Rename the spec to "Web App Manifest" to align with the established name

anssiko:
Rename the spec to "Web App Manifest" to align with the [established name][1].

(I don't always bikeshed about naming, but when I do, I rename the spec.)

[1]: https://www.google.com/search?q=%22web%20app%20manifest%22

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>
-      Manifest for a web application
+      Web App Manifest
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c-common" class=
     "remove">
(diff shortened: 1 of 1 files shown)


kenchris:
Lgtm

skddc:
+1

index.html
```diff
@@ -318,7 +318,7 @@
         <li>Otherwise, return <code>false</code>.
         </li>
       </ol>
-      <div class="issue" title="🐒 Monkey patch">
+      <div class="issue" title="&#128018; Monkey patch">
```
marcoscaceres: Why do you keep hating the monkey?

anssiko: TextMate keeps on hating the monkey.

marcoscaceres: We need to get you a copy of sublime ;)

PR #407, opened 2015-10-10
SVG icon sizes="any" (or without property "sizes")

laukstein:
SVG icon sizes="any" for any icon size, apply it always until specified icons with different sizes like 72x72, 48x48, etc.
The same logic also for "link" element.

Maybe we can even drop "sizes" property if the image type is SVG image/svg+xml. It is understandable that SVG image is scalable.
Speaking about "link" element, Apple earlier proposed with sizes="any" and later dropped that property, see spec https://developer.apple.com/library/safari/releasenotes/General/WhatsNewInSafari/Articles/Safari_9.html#//apple_ref/doc/uid/TP40014305-CH9-SW20. Has been shipped it Firefox 41 https://bugzilla.mozilla.org/show_bug.cgi?id=366324, Chrome still working on it https://code.google.com/p/chromium/issues/detail?id=294179. Related issue in https://github.com/whatwg/html/issues/110

### DIFF
--- a/index.html
+++ b/index.html
@@ -193,7 +193,8 @@
 &lt;meta name="application-name" content="Store Finder"&gt;
 &lt;link rel="icon" sizes="16x16 32x32 48x48" href="lo_def.ico"&gt;
 &lt;link rel="icon" sizes="512x512" href="hi_def.png"&gt;
-&lt;link rel="icon" sizes="any" href="icon.png"&gt;
+&lt;link rel="icon" sizes="any" href="icon.svg"&gt;
+&lt;link rel="icon" href="icon-repace.svg"&gt;
 </pre>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -193,6 +193,7 @@
 &lt;meta name="application-name" content="Store Finder"&gt;
 &lt;link rel="icon" sizes="16x16 32x32 48x48" href="lo_def.ico"&gt;
 &lt;link rel="icon" sizes="512x512" href="hi_def.png"&gt;
+&lt;link rel="icon" sizes="any" href="icon.png"&gt;
 </pre>
       </section>
     </section>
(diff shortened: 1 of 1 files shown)


mounirlamouri:
SVG images are vectorial but that doesn't mean that they can scale to any size easily. An SVG image optimized for say 32x32 might look fairly ugly if shown as a 318x318 icon.

laukstein:
@mounirlamouri, for that we can use CSS media queries in SVG, see @SaraSoueidan article http://tympanus.net/codrops/2014/08/19/making-svgs-responsive-with-css.

laukstein:
After thinking more, I expect without **sizes** must behavior as **any** size. So `sizes="any"` can be dropped.

laukstein:
sizes="any" already defined in https://html.spec.whatwg.org/#attr-link-sizes-any

marcoscaceres:
@laukstein I'm still a bit confused by what you are trying to propose. Should we just say that when sizes is missing, just default to "any"?

marcoscaceres:
@laukstein to be clear, we can't know with any certainty if an icon is SVG before we download it (even if it contains `type=`).

laukstein:
@marcoscaceres I propose to scale-fit SVG image and apply as a fallback to any appropriate image size (image and image size that isn't already defined), related issue https://github.com/w3c/manifest/issues/409. Property **sizes** here can be dropped.

marcoscaceres:
I'm closing this for now. We might end up coming back to this if it becomes an issue in practice.

laukstein:
@marcoscaceres, still no considerations for this?

marcoscaceres:
@laukstein, not yet... has it come up in the wild for you?

laukstein:
@marcoscaceres, I would prefer to use single scaleable SVG, instead of multiple size icons defined in HTML and in Manifest, etc.

Agagamand:
> I would prefer to use single scaleable SVG, instead of multiple size icons defined in HTML and in Manifest, etc.

I also.

PR #410, opened 2015-10-13
Migrate to the Web Platform WG

anssiko:
Given the [Web Platform WG][1] was started 10 Oct [terminating][2] the WebApps WG, I guess we want to update the boilerplate to match. This spec is listed in the [WPWG's deliverables][3].

[1]: https://www.w3.org/WebPlatform/WG/
[2]: https://lists.w3.org/Archives/Public/public-webapps/2015OctDec/0058.html
[3]: http://www.w3.org/2015/10/webplatform-charter.html#deliverables

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,10 +39,10 @@
             companyURL: "http://intel.com",
             w3cid: 41974
         }],
-        wg: 'Web Applications (WebApps) Working Group',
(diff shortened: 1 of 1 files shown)


kenchris:
lgtm

PR #413, opened 2015-11-21
Proper desc, cleanup

chaals:
the element is called desc. And should describe the image.

Also, assorted random trimming of excess precision...


marcoscaceres:
@chaals, thanks!

PR #414, opened 2015-11-24
grouping, precision

chaals:
including fixing the bugs I added


marcoscaceres:
thanks again @chaals :+1:

PR #420, opened 2016-01-27
Enable RDFa

halindrome:
The ReSpec configuration for this specification disables RDFa generation.  RDFa generation is stable and mature in ReSpec and provides for more ready processing of W3C specifications by knowledge engines.  Please enable RDFA generation in your specification.

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,7 +11,7 @@
     <script class='remove'>
 //make tidy happy
     var respecConfig = {
-        doRDFa: false,
+        doRDFa: true,
         previousPublishDate: "2013-12-17",
         specStatus: 'WD',
         shortName: 'appmanifest',


marcoscaceres:
The Editors' choice is to disable RDFa for this spec. We kindly ask that you please respect our choice to do so.

joshuastobaugh57:
I am learning on my own, what am I supposed to put in a reply e mail of this type

-----Original Message-----

marcoscaceres:
@joshuastobaugh57, no probs. You have two options: 

1. just write, "Ok, thanks for your consideration."
1. Make a case for adding it by writing something like, "I would kindly ask that you reconsider your position ..." and present a case as to why would should. You should, for instance, state exactly which "knowledge engines" are using this spec and for what purpose - and who is that of value to in particular. There needs to be strong evidence of value that the Editors can see (i.e., no empty promises, or "one day", or any such nonesense ... we need to *see it*, access it, or be able to somehow use it).

marcoscaceres:
(fixed a couple of typos above)

halindrome:
Thank you for your response @marcoscaceres .  Obviously the most interesting use case for embedding semantic information on our specifications is search engine indexing.  There are other compelling knowledge engine cases, but I imagine you would not find those compelling.  I will circle back with the Schema.org people and see if I can't help 1) improve the semantic markup that ReSpec emits for RDFa, and 2) help convince you that it is both USEFUL and USED already in the marketplace.

marcoscaceres:
@halindrome, I would be sold if Google (or one of the other engines) did something cool with the metadata. Right now "web manifest" returns top spot on Google and Bing - but maybe with RDFa, additional information about the document can be displayed? I'm really looking for stuff like that.

Also, if there has been progress on knowledge engines in the last few years - then I'd like to see some mainstream examples - and in particular, as it would apply to W3C specs. 

FYI, I'm not against metadata when it makes sense - but I'm really just looking for applicable stuff: I want to **see** the results of adding the data. I'm even working on project for Firefox where we intend to use the Opengraph data, plus Twitter's metadata, etc. See: https://github.com/mozilla/activity-streams/issues/36 . There, we intend to visualize Opengraph data, for instance.

PR #428, opened 2016-03-05
Remove reference to RFC5988.

### LINKED ISSUE #424: Is RFC5988 supported by implementers?
The last sentence of the spec's Abstract section mentions a document can be associated with a manifest "either directly in an HTML document or over the wire using RFC5988"  (https://tools.ietf.org/html/rfc5988) 

If I understand correctly, the directly in a HTML document would by via:
<link rel="manifest" href="..." >

And via RFC5988 would be a HTTP header:
Link: http://...; rel="manifest"

If referencing a manifest via RFC5988 is somewhat common and/or supported by implementers, I recommend we elaborate on this as a new section 5.2

If referencing a manifest via RFC5988 is very rare and/or not supported by implementers, I recommend considering removing the reference to this method in the Abstract section.

mounirlamouri:
As far as I know this is not implemented. Maybe someone has context about why it was added in the first place?

Closes #424.

### DIFF
--- a/index.html
+++ b/index.html
@@ -99,9 +99,7 @@
       <p>
         To associate documents of a web application with a manifest, this
         specification defines the <code>manifest</code> link type as a
-        declarative means for a document to be associated with a manifest
-        (either directly in a HTML document or over the wire using
-        [[RFC5988]]).
+        declarative means for a document to be associated with a manifest.
       </p>
     </section>
     <section id="sotd">


mounirlamouri:
@marcoscaceres PTAL

benfrancis:
I assume the idea is that if the link relation is specified in an HTTP header the manifest can be fetched immediately without having to wait for an HTML document to download. I think I remember some previous discussion about potential use cases related to signing, but as far as I know nobody has implemented the RFC5988 support yet.

marcoscaceres:
@mounirlamouri, it seemed like a good idea at the time :) I'm ok for you to merge this. We can always revive later if there is interest.

marcoscaceres:
@mounirlamouri, before you merge, can you please modify the commit message to include "(closes #424.)".

PR #433, opened 2016-03-22
Feat (dir member): Add dir member (closes #334)

### LINKED ISSUE #334: i18n: No direction metadata for 'name'; [I18N-ISSUE-416]
http://www.w3.org/International/track/issues/416

Raised by: Addison Phillips
http://www.w3.org/TR/2015/WD-appmanifest-20150212/#name-member

The 'name' manifest member does not provide a means of indicating the base direction of the name string. This is equivalent to extracting the 'dir' attribute appropriate for the `<title>` element's scope and may be necessary for correct rendering of bidirectional content.

For example, the author might need to set the base direction explicitly, since the Unicode paragraph detection algorithm can be fooled by a paragraph that starts with a strong LTR character, but is actually a RTL paragraph (or vice versa), eg. "نشاط التدويل is how you say 'i18n Activity' in Arabic."

It would be helpful to the author to be able to set the default base direction for a whole manifest file to rtl.

marcoscaceres:


marcoscaceres:
@r12a, could you possibly review this?

marcoscaceres:
@aphillips, a review from you would be much appreciated. In particular, I would like to just hook directly into the [[!BIDI]] algorithms if possible: that is, given a "dir", just run algorithm x...

index.html
```diff
@@ -136,6 +136,7 @@
         <pre class="example highlight json" title="typical manifest">
 {
   "lang": "en",
+  "dir": "ltr",
```
kenchris: Why was dir chosen instead of direction or directionality? Is this a common abbreviation for this? It made me think of directory :)

marcoscaceres: > On 22 Mar 2016, at 7:35 PM, Kenneth Rohde Christiansen <notifications@github.com> wrote:
> 
> In index.html:
> 
> > @@ -136,6 +136,7 @@
> >          <pre class="example highlight json" title="typical manifest">
> >  {
> >    "lang": "en",
> > +  "dir": "ltr",
> Why was dir chosen instead of direction or directionality? Is this a common abbreviation for this? It made me think of directory :)
> 
Similarly named attributes in the HTMLs and the XMLs.

kenchris: OK, makes sense then

index.html
```diff
@@ -1086,6 +1092,120 @@ <h3 id="applying">
       </p>
       <section>
         <h3>
+          <code>dir</code> member
+        </h3>
+        <p>
+          The <dfn><code>dir</code> member</dfn> specifies the <dfn>default
```
aphillips: I would remove the word "default" from "default base direction". The base direction is the starting paragraph direction for the Unicode bidi algorithms (UAX#9, see BD5).

index.html
```diff
@@ -1086,6 +1092,120 @@ <h3 id="applying">
...
+          </dt>
+          <dd>
+            <a data-lt=
+            "steps to programmatically determine the directionality of a member">
+            programmatically determined</a> from the value of the member.
+          </dd>
+        </dl>
+        <p>
+          When displaying the <a>directionality-capable members</a> to an
+          end-user, the user agent MUST use the <a>default base direction</a>
```
aphillips: again s/default base direction/base direction/

index.html
```diff
@@ -1086,6 +1092,120 @@ <h3 id="applying">
...
+          to correctly position the characters in text containing characters
```
aphillips: "correctly position the characters in text containing characters flowing from right to left" is not quite correct. It isn't necessarily about positioning, although one might refer to it that way.

Perhaps:
--
... to compute directional runs and layout or position text correctly in text containing mixed-direction sequences.
--

aphillips:
@marcoscaceres I added some comments to the diff. These are somewhat minor.

marcoscaceres:
@aphillips, this is good feedback. Thank you! I'll take that the changes needed are minor as a "good thing" :tm:

PR #439, opened 2016-03-31
Feat (splashscreens): Marked splashscreens as at risk

marcoscaceres:

PR #440, opened 2016-03-31
Feat (splashscreens): remove splash screen from spec (closes #9)

### LINKED ISSUE #9: Manifest format should support some way of including a splash screen
**This feature is at risk.** Unless we hear otherwise, this feature will be removed from the specification in the next few months.  

#### Why is this at risk?
In the past, various vendors had expressed an interest in having a splash screen for when an app boots up. However, at least one implementer claims that it's possible to achieve this effect by using `background_color` + `icon` members. Additionally, browsers engines are able to boot up in sub 200ms, making it possible for authors to create their own splash screen experiences without needing to rely on this.  

#### Use case
Because of the limited processing capabilities of some devices and the amount of time it takes to initialize a Web runtime, there can be a small - yet noticeably perceivable - delay between when the end-user starts an application and the application actually becoming available and interactive to the end-user.

Depending on the complexity of the application and the processing power of the device, this delay can range from a few milliseconds to a couple of seconds. When applications take more than a few seconds to load, it can be disorienting to end-users as nothing is displayed on the screen; leading to a poor user experience. Native applications overcome this issue by allowing developers to use a media file that is displayed the instant the application is launched. Such a file is commonly referred to as a "splash screen" or "launch image".

marcoscaceres:
P re-emtive removal of splashscreen. Don't merge until #439 has been settled and if implementers agree.

### DIFF
--- a/index.html
+++ b/index.html
@@ -997,11 +997,6 @@ Content-Type: application/manifest+json
           <var>manifest</var>, <var>manifest URL</var>, and "icons" as
           arguments.
           </li>
-          <li>Let <var>splash screens</var> of <var>parsed manifest</var> be
-          the result of running the <a>steps for processing an array of
-          images</a> with <var>manifest</var>, <var>manifest URL</var>, and
-          "splash_screens" as arguments.
-          </li>
           <li>Let <var>scope</var> of <var>parsed manifest</var> be the result
           of running the <a>steps for processing the <code>scope</code>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
Ok from Mozilla to remove splash screens.

kenchris:
lgtm

marcoscaceres:
Rebased.

PR #444, opened 2016-04-05
Feat(displaymodes): add priv/sec note (closes #402)

### LINKED ISSUE #402: Privacy Review: how to handle navigation to a fullscreen'd application?
It is actually more of a security issue than a privacy issue. If a user has an application installed and the UA wants them to use it instead of the regular browser experience when clicking on a link, it might be very easy for that application to spoof the user and make them believe they are on another website.

The specification should at least mention that issue and provide some guidance. Maybe show the origin in that case? Always allow the user to find the URL bar?

marcoscaceres:


marcoscaceres:
@kenchris, your magic touch please?

index.html
```diff
@@ -556,6 +556,30 @@ <h3 id="navigation-scope-security-considerations">
       </p>
       <section>
         <h3>
+          Privacy and security considerations
+        </h3>
+        <p>
+          When the <a>fullscreen</a> <a>display mode</a> is applied, it is
+          RECOMMENDED that the user agent still afford the end-user a means to
```
kenchris: it afford*s* but I am not sure afford is the right word here. Why not use 'allows'

index.html
```diff
@@ -556,6 +556,30 @@ <h3 id="navigation-scope-security-considerations">
...
+          access/view the origin, or URL, and icon of the Web application. This
```
kenchris: ... RECOMMENDED that the user agent provides the end-user a means to consult basic information about the web apps, such as the origin, complete URL, granted permissions and associated icon.

Such information shouldn't necessarily need to overlay the application itself, but can instead be shown in the task switcher, for instance by clicking on an info button.

kenchris: ... RECOMMENDED that the user agent provides the end-user a means to consult basic information about the web apps, such as the origin, complete URL, granted permissions and associated icon.

Such information shouldn't necessarily need to overlay the application itself, but can instead be shown in the task switcher, for instance by clicking on an info button.

kenchris: ... RECOMMENDED that the user agent provides the end-user a means to consult basic information about the web apps, such as the origin, complete URL, granted permissions and associated icon.

Such information shouldn't necessarily need to overlay the application itself, but can instead be shown in the task switcher, for instance by clicking on an info button.

kenchris: ... RECOMMENDED that the user agent provides the end-user a means to consult basic information about the web apps, such as the origin, complete URL, granted permissions and associated icon.

Such information shouldn't necessarily need to overlay the application itself, but can instead be shown in the task switcher, for instance by clicking on an info button.

index.html
```diff
@@ -556,6 +556,30 @@ <h3 id="navigation-scope-security-considerations">
...
+          does not necessarily mean overlaying the origin and icon over the
+          application itself. Rather, for example, the origin could be shown in
+          the OS's task manager or some such.
+        </p>
+        <p>
+          Additionally, when applying a manifest that changes the <a>display
```
kenchris: which sets the

index.html
```diff
@@ -556,6 +556,30 @@ <h3 id="navigation-scope-security-considerations">
...
+          mode</a> to anything except "<a>browser</a>", it is RECOMMENDED that
+          the user agent clearly indicate to the end-user that their are
+          leaving the normal browsing context of a web browser. Ideally,
+          launching or switching to a web application is performed in a manner
+          that is consistent with launching or switching to other applications
+          in the OS. For example, a long and obvious animated transition, or
+          speaking the text "Launching application X", or whatever is customary
```
kenchris: I would avoid the word 'whatever' it sounds like you dont give a shi... :)

index.html
```diff
@@ -556,6 +556,30 @@ <h3 id="navigation-scope-security-considerations">
...
+          on the host <abbr>OS</abbr>.
```
kenchris: Why not say platform instead of OS

index.html
kenchris: which sets the

index.html
kenchris: I would avoid the word 'whatever' it sounds like you dont give a shi... :)

index.html
kenchris: Why not say platform instead of OS

index.html
marcoscaceres: Argh, now I can't sound like a UX hipster :(

index.html
marcoscaceres: But I don't :) Fixing.

index.html
marcoscaceres: Used platform.

index.html
marcoscaceres: This presupposes access to the OS's task switcher - which, for instance, wouldn't be true for Firefox on Android. I'm just going to say that it's left up to the implementer to come up with something.

PR #451, opened 2016-04-14
Feat(density): remove density member (closes #450)

### LINKED ISSUE #450: Drop density descriptor from spec
There seems to be a lot of confusion about how to properly use the `density` member. It was designed to support discriminating against the user's screen: that is to say, if the end-user screen greater than or equal to X, then use this icon. 

When we started this effort a few years ago, this use case was driven by iOS developers that were having to create specific icons for 1x and 2x - not only for performance, but because 2x icons were getting scaled incorrectly in 1x displays. However, this is a bit of an edge case nowadays because most phones have +2x displays. 

Because of misuse and confusion around this feature, we are seeing a lot that is using density as a descriptor of the icon instead. E.g., a 36x36 pixel icon is "0.75"... which doesn't make any sense. Anyway, this gets worst for larger icons, where some developers have specified things like "density": "4" - where those icons would be fine to use on a 2x density display. 

As such, I propose we drop density from the spec.  

@mounirlamouri, @kenchris, @RobDolinMS, @boyofgreen, wdyt?

marcoscaceres:


marcoscaceres:
@mounirlamouri, r?

mounirlamouri:
:wave: :cry:

Issue #446, opened 2016-04-05
Specify how updates work


adrianhopebailie:
Is there a reason that the spec only defines a declarative way to request that a user agent obtain and process a manifest?

Likewise, it would be useful to have an explicit way to request that the user-agent update its version of the manifest.

marcoscaceres:
@adrianhopebailie yeah, we might need to add an explicit means to request an update to mitigate security issues: that is, we want to be able to show what's changed and give the end-user a chance to review or dismiss the update. 

Excuse the crappy mock-up I slapped together below, but want to illustrate the point:  

![screenshot 2016-04-06 14 16 44](https://cloud.githubusercontent.com/assets/870154/14305698/39373a10-fc02-11e5-9b5e-d06837ce4690.png)

The UA would also need to show more info, like if the display mode or start URL changed, etc. Basically, anything that could be a concern to a user.

marcoscaceres:
@adrianhopebailie, about "installing explicitly", most browser vendors would like to pursue the [installabiltiy signals](https://w3c.github.io/manifest/#installability-signals) route. If that doesn't work out, we might need to add an .requestInstall() API. However, we have resisted that option because we don't want install buttons all over the web.

adrianhopebailie:
What is the logic a user agent should apply to decide if the installation of one manifest is actually a new install or an update/replacement of an existing one? It would seem that this should be related to scope but I'm not sure if that's explicitly defined anywhere?

Put differently, why would you say this?

> we want to be able to show what's changed and give the end-user a chance to review or dismiss the update. 

If I visit the same URL a week apart I get no warnings that the content has changed since the last time I visited?

If the "effect" of a manifest is limited to functions that will not impact security or privacy then the normal controls for managing those risks should suffice surely?

marcoscaceres:
> If I visit the same URL a week apart I get no warnings that the content has changed since the last time I visited?

But you totally do. For instance, Safari shows you the following when you go to bbc.com/news: 

![screenshot_2016-04-08_17_57_08](https://cloud.githubusercontent.com/assets/870154/14377551/96d3cd16-fdb3-11e5-9dff-8242434e31f5.png)

And Chrome: 
![screenshot_2016-04-08_18_07_19](https://cloud.githubusercontent.com/assets/870154/14377785/f99b425c-fdb4-11e5-9c0c-94bf02058e6f.png)

[...]

PR #457, opened 2016-04-21
Fix(updates): remove text about updates (closes #384)

### LINKED ISSUE #384: Updating is under/incorrectly specified
The spec says that the user agent may "periodically check if the contents of a manifest has been modified (e.g., by honoring HTTP cache directives associated with the manifest or by checking for updates after the web application has been launched)."

Firstly, being able to periodically check the contents of the manifest requires that the manifest URL does not change. If the manifest URL changes, there's no way to know that any new manifest corresponds to the same app as the old manifest and the only way for the app's metadata to be updated is for the user to re-install the app.

The removal of the same-origin restriction on manifest URLs was intended mainly to allow CDNs to host the manifest on a separate origin to the app, but the above assumes that CDNs are well behaved in that they never change the URL of the manifest, or as a minimum provide a redirect. @slightlylate assures me that CDNs these days are well behaved and do not change URLs of resources. Does anyone have any additional data on that?

So assuming the manifest URL does not change, or at least provides redirects, the user agent can "periodically check the contents" of the manifest to see whether something has been updated. Then the spec says that "in the event that the members of the manifest have been updated, as determined by running the steps for processing a manifest and seeing if anything has changed, the user agent MAY update the metadata corresponding to the web application ".

Now during the "steps for processing the manifest" the spec says that the start_url must be checked to be same origin with the document URL, that is the document the app was installed from. The same is true for the scope property. If the user agent is just "periodically checking the contents" of the manifest, does it need to also keep a record of the URL of the document the app was installed from, in order to make these checks? Note that this URL may be different for every installation as an app can be installed from any page of the app.

So assuming that CDNs behave well, the manifest URL doesn't change, and we've stored the document URL, we can then periodically check the contents of the manifest. Does this work for authenticated manifest URLs too? How does the user agent get the necessary credentials to fetch the manifest? Is the check only done by the user agent when the user logs in and uses the app?

It sounds like the implementation in Chrome may also be based on some additional assumptions regarding a registered Service Worker scope, is that the case? @mounirlamouri ?

If the assumptions above are in fact assumptions made by the spec, do we need to write something in the spec about the fact that user agents do not expect the manifest URL to change, and that therefore CDNs are expected to keep the same URL for the manifest or provide a redirect?

marcoscaceres:
Removed the text about updating. Updating probably needs to happen explicitly on request by the developer via some API.

### DIFF
--- a/index.html
+++ b/index.html
@@ -911,8 +911,7 @@ if (standalone.matches) {
       <p>
         This section defines algorithms for <a href="#obtaining">obtaining</a>,
         <a href="#processing">processing</a>, and <a>applying</a> a
-        <a>manifest</a>, and gives recommendations to implementers on how to
-        react when the manifest is <a href="#updating">updated</a>.
+        <a>manifest</a>.
       </p>
       <section id="obtaining">
         <h3>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@kenchris r?

marcoscaceres:
In summary, it just removes the text till we find an appropriate solution for updating. 

@benfrancis, does this work for you? In particular, as we flesh out #446?

kenchris:
lgtm

marcoscaceres:
Ok, let's keep chatting in  #446

PR #458, opened 2016-04-21
Chore(extensions): add link to registry (closes #452)

### LINKED ISSUE #452: Possible Registry for Extension Properties
The current draft has a very useful extension point that is supposed to catch properties that are not specified in the draft itself. One possible way to make this a bit more managed would be to establish a registry where new properties can be registered and then become well-known. This way other specs extending what can be represented in a webapp manifest could do that in a more managed way.
https://tools.ietf.org/html/draft-wilde-registries talks about the general pattern of having a registry, and if there were such a registry, then the nect question would be how to establish one. There are several design options for this, but this issue is simply about the question of whether having a registry would be a useful thing to do.

marcoscaceres:


marcoscaceres:
@dret, wdyt?

dret:
thanks for making those changes, and of course i like the idea of a registry for extensions!
but as you know, i think that there are better options to implementing a registry pattern than just putting up a wiki somewhere and assuming that it will self-regulate. but that's a different discussion to have.
in the specs where we use extensibility and registries, we take great care to specify the processing model so that extensions are always optional, must allow for meaningful processing even when a client does not support the extensions, and are self contained in the sense that extensions can never change the semantics of anything outside of their own scope. this guidance helps to discourage the creation of extensions that are breaking changes instead of evolutionary ones.
i have wanted to write about this for a little while, actually, because i think that this is a pattern that doesn't get as much recognition as it should. let me write up a little blog post about this, and then i'll link to it.

marcoscaceres:
@dret, that all sounds great and I agree with the general direction. If you have thoughts on improving the processing model, I'm all ears. I'd like to start with the wiki regardless. I don't think that we need anything more heavy-weight for now. We should build that up over time if we need to. 

So, is it ok to merge?

index.html
```diff
@@ -3121,6 +3127,14 @@ <h3 id="applying">
             prefix.
           </p>
           <p>
+            We encourage implementors to add proprietary extensions to our
+            <a href=
+            "https://github.com/w3c/manifest/wiki/Extensions-Registry">Extensions
+            Registry</a>. This allows the community to track what extension to
```
dret: "... to track what extensions have been defined and documented..." ?

index.html
```diff
@@ -3121,6 +3127,14 @@ <h3 id="applying">
...
+            be made and periodically consider those extensions for
+            standardization.
```
dret: the idea of extensions usually is that (a) they are "standards" in the sense that they must be documented and this documentation should be linked from the registry. making sure that there (a) is documentation, and (b) what has been documented does not violate the processing model of the foundation spec is something that often is handled by some sort of feedback/review when registrations are made. whether the documentation is or is intended to become a standard in itself is not so important, as long as it is documented and follows the extension model.

index.html
```diff
@@ -3096,6 +3096,12 @@ <h3 id="applying">
           consistent.
         </p>
         <p>
+          To allow the community can easily find extensions, please add your
+          extensions to the <a href=
+          "https://github.com/w3c/manifest/wiki/Extensions-Registry">Extensions
+          Registry</a>.
```
dret: is the idea of "standard" and "proprietary" extensions similar to that of what IETF sometimes calls "experimental" and "permanent", with a different level of scrutiny when it comes to documentation and stability?

marcoscaceres: No. "standard" ones could be more "experimental" than proprietary things, and vice versa: the point is to distinguish things that a group of implementers are working on with the aim to standardize, and things that ship by a single vendor but there might not be any appetite to standardize for whatever reason. 

Having said that, for each category, those could be marked as stable or experimental.

index.html
marcoscaceres: Something like the above would be great as part of the introduction of the wiki page.

index.html
marcoscaceres: added

index.html
dret: > No. "standard" ones could be more "experimental" than proprietary
> things, and vice versa: the point is to distinguish things that a group
> of implementers are working on with the aim to standardize, and things
> that ship by a single vendor but there might not be any appetite to
> standardize for whatever reason.

so you're grouping by origin rather than maturity/community? that's fine 
as well, as long as it's clear and some process is in place to sort 
things accordingly.

index.html
dret: > Something like the above would be great as part of the introduction of
> the wiki page.

any hope that people actually read it and put it in as intended, instead 
of just picking the first table and adding it there? if there's 
structure, it would be good to have some model of how it is maintained.

dret:
here's the blog post on robust extensibility i was mentioning earlier: http://dret.typepad.com/dretblog/2016/04/robust-extensibility.html
it talks about the necessity to have well-defined extension points and a processing model for them. i will follow-up with a post on how to achieve this in an evolvable way, also discussing registries and the various options for them.

marcoscaceres:
@dret, would you be willing to be the person in charge of handling extensibility and the registry for this project?

dret:
On 2016-04-28 23:30, Marcos Cáceres wrote:
> @dret <https://github.com/dret>, would you be willing to be the person
> in charge of handling extensibility and the registry for this project?

handling the registry in terms of running it, or handling the 
extensibility and registry aspects of the spec and work towards a 
solution? assuming it's the latter and that there's editorial 
recognition, it would be great to contribute to the effort.

dret:
On 2016-04-29 12:03, Erik Wilde wrote:
> On 2016-04-28 23:30, Marcos Cáceres wrote:
>> @dret <https://github.com/dret>, would you be willing to be the person
>> in charge of handling extensibility and the registry for this project?
>
> handling the registry in terms of running it, or handling the
> extensibility and registry aspects of the spec and work towards a
> solution? assuming it's the latter and that there's editorial
> recognition, it would be great to contribute to the effort.

any update on this one? just trying to better understand the question 
and definitely interested in how this is going to be addressed in terms 
of API design, and an evolving identifier space. thanks!

marcoscaceres:
> On 30 Apr 2016, at 5:03 AM, Erik Wilde <notifications@github.com> wrote:
> 
> On 2016-04-28 23:30, Marcos Cáceres wrote:
> > @dret <https://github.com/dret>, would you be willing to be the person
> > in charge of handling extensibility and the registry for this project?
> 
> handling the registry in terms of running it, or handling the 
> extensibility and registry aspects of the spec and work towards a 
> solution? assuming it's the latter and that there's editorial 
> recognition, it would be great to contribute to the effort.
 
Apologies for the delay. 

Yes. Absolutely - but it means taking on editorial responsibility, and reviewing changes, etc. 

We recognize the people listed as Editors because they've put in significant hours of work diligently reviewing, and often implementing, the spec (I.e., one needs to earn their stripes by editing and showing a long term commitment to the project... Akin to, say, becoming a Blink or WebKit contributor as recognize by those communities). 

So, the place to start is by reviewing the spec, making PRs, filing relevant bugs, making sure the registry is up-to-date etc. 

Make sense?

dret:
On 2016-05-06 23:02, Marcos Cáceres wrote:
> Yes. Absolutely - but it means taking on editorial responsibility, and
> reviewing changes, etc.

sounds good to me, i'd like to do that.

> We recognize the people listed as Editors because they've put in
> significant hours of work diligently reviewing, and often implementing,
> the spec (I.e., one needs to earn their stripes by editing and showing a
> long term commitment to the project... Akin to, say, becoming a Blink or
> WebKit contributor as recognize by those communities).

sure thing. put in the work, get the recognition for doing it.

> So, the place to start is by reviewing the spec, making PRs, filing
> relevant bugs, making sure the registry is up-to-date etc.
> Make sense?

it does make sense, as a general approach. but i think we have a bit 
differing opinions on what works well as a robust/effective approach to 
manage a registry. you may remember from other specs that i find the 
"let's create a wiki and be done with it" approach not necessarily the 
best one. but i guess that's where the editorial fun starts, right?

marcoscaceres:
> On 10 May 2016, at 4:29 AM, Erik Wilde <notifications@github.com> wrote:
> 
> On 2016-05-06 23:02, Marcos Cáceres wrote:
> > Yes. Absolutely - but it means taking on editorial responsibility, and
> > reviewing changes, etc.
> 
> sounds good to me, i'd like to do that.
> 
> > We recognize the people listed as Editors because they've put in
> > significant hours of work diligently reviewing, and often implementing,
> > the spec (I.e., one needs to earn their stripes by editing and showing a
> > long term commitment to the project... Akin to, say, becoming a Blink or
> > WebKit contributor as recognize by those communities).
> 
> sure thing. put in the work, get the recognition for doing it.
> 
> > So, the place to start is by reviewing the spec, making PRs, filing
> > relevant bugs, making sure the registry is up-to-date etc.
> > Make sense?
> 
> it does make sense, as a general approach. but i think we have a bit 
> differing opinions on what works well as a robust/effective approach to 
> manage a registry. you may remember from other specs that i find the 
> "let's create a wiki and be done with it" approach not necessarily the 
> best one. but i guess that's where the editorial fun starts, right?

Exactly:) looking forward to collaborating further.

PR #459, opened 2016-04-21
Feat(image objects): clarify how to fetch (closes #443)

### LINKED ISSUE #443: Clarify how image objects should be fetched
Chrome stable has this bug - https://sw-cors-demo.herokuapp.com/

1. Image would be requested
1. SW would cache the request & response
1. "Add to homescreen" pressed
1. Icon fails to load

This is because the icon was fetched with mode "cors", and the service worker responded with an opaque response.

Chrome has since fixed this issue, but this spec should detail how images should be fetched to avoid other UAs making the same mistake.

Fetching should be similar to https://html.spec.whatwg.org/multipage/embedded-content.html#update-the-image-data - as in mode "no-cors".

marcoscaceres:
@jakearchibald WIP, let me know if this is ok?

### DIFF
--- a/index.html
+++ b/index.html
@@ -2142,6 +2142,43 @@ Content-Type: application/manifest+json
       </p>
       <section>
         <h3>
+          Fetching image objects
+        </h3>
+        <p>
+          To <a>fetch</a> an <a>image object</a>, the user agent MUST run the
+          <dfn>steps to fetch an image object</dfn>. The algorithm takes an
+          <a>image object</a> and the <var>manifest's URL</var>. It returns a
+          <a>Response</a>:
+        </p>
+        <ol>
+          <li>Let <var>request</var> be a new <a>Request</a>.
+          </li>
+          <li>Set the following properties of <var>request</var>:
+            <ol>
+              <li>
+                <var>url</var> is set to image object's URL.
+              </li>
+              <li>
+                <var>initiator</var> is "<code>manifest</code>".
+              </li>
+              <li>
+                <var>type</var> is "<code>image</code>".
+              </li>
+              <li>
+                <var>cors</var> is "<code>no-cors</code>".
+              </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2142,6 +2142,39 @@ <h3 id="applying">
       </p>
       <section>
         <h3>
+          Fetching image objects
+        </h3>
+        <p>
+          To <a>fetch</a> and <a>image object</a>, the user agent MUST run the
```
kenchris: an not and

jakearchibald:
This looks great to me. The only bit I'm unsure of is if `client` should be set or not. Null seems fine since that allows the image to be fetched without a client present, but would like a second opinion from @annevk.

annevk:
If you don't provide a client you should probably provide a referrer and maybe some other data. I guess it might makes sense to not have a client here, since it's a bit separate from the page. Not entirely sure, haven't thought about manifests too much. Still not really sold on them.

marcoscaceres:
> If you don't provide a client you should probably provide a referrer 

Good point about referrer. Added that to be the manifest's URL, obviously. Will leave the others as is for now and see how we go implementation wise.

annevk:
Referrer policy also needs to come from somewhere. I guess if you don't have a client you'll need some integration with the referrer policy specification then...

marcoscaceres:
@annevk, we do always have a reference back to the Document and the Link element that kicked off the request. Can we use that as a means to get a suitable client (e.g., the window object)?

marcoscaceres:
opened https://github.com/w3c/manifest/issues/465

annevk:
Yeah, maybe that's better. Although it's a little weird since you can be associated with multiple documents, right? And that document might disappear and then another one might appear? Would that mean you get a new policy?

How does CSP work today for manifest resources?

PR #462, opened 2016-04-22
Fix(processing): treat enums as case insensitive

marcoscaceres:
* display, orientation, and dir are now case insensitive
 * values are normalized to lower-case on return


mounirlamouri:
:+1:

PR #464, opened 2016-04-22
Fix(orientation): return undefined (closes #463)

### LINKED ISSUE #463: The curious (special) case of orientation
@mounirlamouri I can't remember why, but we made processing the `orientation` member default to  the empty string instead of underfined:

>   If Type(value) is not "undefined", issue a developer warning that the type is not supported.
>   Return the empty string.

However, in Gecko, we want to take the normalized web manifest and convert it into a WebIDL dictionary and make orientation adhere to the `OrientationLockType` of The Screen Orientation API.  

Would it be ok to change the spec to no longer return the empty string and just return undefined? Does that impact you at all?

marcoscaceres:


mounirlamouri:
That's fine. Though, FTR, Chrome's parser is using an internal lock type call "default".

PR #467, opened 2016-05-18
Feat(image object): set client to Document (closes #465)

### LINKED ISSUE #465: Need to specify client of image objects when fetching
@annevk wrote, re: obtaining image objects: 

> Referrer policy also needs to come from somewhere. I guess if you don't have a client you'll need some integration with the referrer policy specification then...

marcoscaceres:


marcoscaceres:
@kenchris r? merge if ok.

marcoscaceres:
@annevk, this ok for now?

annevk:
A client is an environment settings object.

annevk:
I don't think it addresses the more fundamental issues. Also, did you test to see that this is what implementations actually do here?

marcoscaceres:
> On 18 May 2016, at 8:45 PM, Anne van Kesteren <notifications@github.com> wrote:
> 
> A client is an environment settings object.
> 
Which can be a Document, no? That's what HTML says.

annevk:
It's not?

marcoscaceres:
> On 18 May 2016, at 10:17 PM, Anne van Kesteren <notifications@github.com> wrote:
> 
> It's not?
> 
It says: 
A responsible document
A Document that is assigned responsibility for actions taken by the scripts that use this environment settings object.

Environment settings object then just links back to itself. Top-notch editorial work right there, I might add (note also bug after the example, where the text about loops should be part of the preceding definition item).

annevk:
I'm not sure why you merged this, since it's still broken. That an environment settings object has a responsible document, doesn't mean the two are 1:1.

marcoscaceres:
I honestly don't have any idea what you are saying, so instead of going around in circles please just send a PR.

annevk:
I guess I'm similarly confused by this document as expressed in #465 so I'm not sure how to fix this. This is unlikely to match what anyone implements though.

annevk:
And it's basically a type error as far as standards integration goes...

PR #468, opened 2016-05-18
Feat(application's name): how to get the app's name (closes #466)

### LINKED ISSUE #466: short_name to fall back to name & vice versa
I'm seeing developers with manifests where `name` and `short_name` are duplicates, which is ok, but shouldn't be required.

It'd be nice if the `name` and `short_name` members explicitly fell back to each other. This will hopefully avoid an implementation giving up on displaying a name because `short_name` was empty, yet `name` was defined & already as short as it could be, eg "Facebook".

marcoscaceres:


marcoscaceres:
@kenchris, give it a once over and then we can check if Jake approves.

mounirlamouri:
lgtm, though I would not mention splash screen and homescreen directly but mention the space available for the name instead.

marcoscaceres:
@mounirlamouri thanks! updated. I want to have a concrete example tho - but used "spaced available".

marcoscaceres:
err, I mean space available.

marcoscaceres:
@jakearchibald, r?

jakearchibald:
Looks good to me!

PR #474, opened 2016-06-28
Fix(installation): clarify installation event (closes #473)

### LINKED ISSUE #473: Tidy up the install event specification
I'm looking at implementing the `install` event in Chrome and there are a few things that are implied but not explicit in the [Installation Events](https://w3c.github.io/manifest/#installation-events) section.

1. It doesn't explicitly state the type/interface of the event object (which I assume is [Event](https://dom.spec.whatwg.org/#event) because there's no special metadata).
2. The [Installation](https://w3c.github.io/manifest/#installation) algorithm states when to fire the `"install"` event at the *window* object, but this isn't mentioned or linked up in the [Installation Events](https://w3c.github.io/manifest/#installation-events) section. It isn't explicit that there's a connection between the `"install"` event being fired in Section 2.4 and the `oninstall` IDL attribute. (This section doesn't ever mention the circumstances under which the event is dispatched, despite it being mentioned previously.)

Bascially, you can figure out what's going on from reading the example, but I feel it isn't concrete in the spec, the way [events in Service Workers](https://www.w3.org/TR/service-workers/#document-context-events) spec are defined.

Also, I'm not really familiar with the way events are normally added to the Window object (please correct me if ignorant), but I didn't think defining a mixin interface was necessary. In WebIDL, can't you just write:

```WebIDL
partial interface Window {
  attribute EventHandler oninstall;
}
```
to extend the Window object without it having to implement a new interface?

marcoscaceres:
* Drop mixin in favor of partial interface
 * Clarify that installation is an `Event`. 
 * Link to steps to fire the event

### DIFF
--- a/index.html
+++ b/index.html
@@ -374,19 +374,17 @@
       </h2>
       <section>
         <h3>
-          <code>AppInstallEventHandlers</code> mixin
+          Extensions to the <code>Window</code> object
         </h3>
         <p>
-          The <a><code>AppInstallEventHandlers</code></a> mixin defines the
-          event handler attributes on which events relating to the
+          The following extensions to the <code>Window</code> object specify
+          the event handler attributes on which events relating to the
           <a>installation</a> of a web application are fired.
         </p>
         <pre class="idl">
-          [NoInterfaceObject, exposed=(Window)]
-          interface AppInstallEventHandlers {
+          partial interface Window {
             attribute EventHandler oninstall;
           };
-          Window implements AppInstallEventHandlers;
         </pre>
         <div class="example">
           <p>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@mgiuca, for your reviewing pleasure...

index.html
```diff
@@ -410,8 +408,12 @@ <h3 id="installation-sec">
             <code>oninstall</code> attribute
           </h4>
           <p>
-            <code>oninstall</code> is an <a>event handler IDL attribute</a> for
-            the "<a>install</a>" event type.
+            The <dfn for="Window">oninstall</dfn> is an <a>event handler IDL
```
mgiuca: Drop the "The". Just `oninstall`, it's cleaner :)

mgiuca:
lgtm with a minor grammar nit. Thanks for such a quick turnaround.

PR #485, opened 2016-08-09
refactor: install steps

marcoscaceres:


index.html
```diff
@@ -309,32 +309,28 @@
...
-          the manifest</a>.
-          </li>
-          <li>If <a>obtaining the manifest</a> results in an error, a user
-          agent can, at this point, fall back to using the <a>top-level
-          browsing context</a>' <code>Document</code>'s metadata to create an
-          <a>installation process</a>.
-          </li>
-          <li>If the <a>installation succeeded</a>, <a>fire an event</a> named
-          <code>install</code> at the <var>window</var> object.
+          <li>Then, <a>in parallel</a>:
```
kenchris: There is only one <ol> so why in parallel?

marcoscaceres: it's done off main thread.

kenchris: ah, in parallel with main thread.

lgtm then

marcoscaceres: 👍

PR #487, opened 2016-08-12
[Example] Make Racer 2k example from future

RobDolinMS:
Since the description says the Super Racer game is "The ultimate futuristic racing game from the future!" it may be more appropriate for it to be "Super Racer 3000" (future) instead of "Super Racer 2000" (16 years ago.)


kenchris:
Lgtm

Issue #405, opened 2015-09-26
Support for platform hints on icons


owencm:
Looking at Slack's website the other day I noticed just how different icon standards are for different platforms:

<img width="1101" alt="screen shot 2015-09-25 at 5 29 22 pm" src="https://cloud.githubusercontent.com/assets/279720/10114742/31a137b4-63ab-11e5-8af5-bfcbd9f91472.png">

I wonder if we should support the ability for sites to optionally hint a specific icon is more suitable to a certain platform. For example the manifest could provide default icons in different sizes in the way today, and additionally versions with a new property like `suggested_platforms` containing a list of strings corresponding to platforms that icon would be well suited for.

I'm not sure of the best way for us to standardize (or leave unstandardized?) platform strings though so UAs know which OS to apply them on - naively we could define a short list and add ones as they request to be added, but that seems potentially difficult. Also it's unclear to me whether we'd want to support versioning (e.g. 'win10' vs 'win' etc).

marcoscaceres:
@owencm thanks for the additional input + example! This is a duplicate of https://github.com/w3c/manifest/issues/361. Please take a look. 

FWIW, we are planning on adding `platform`. Just waiting for Gecko to catch up.

PR #490, opened 2016-08-25
feat(screenshots): Add screenshots member (closes #477)

### LINKED ISSUE #477: Add screenshots object
Many catalogs / stores of apps include one or more screenshots of an app.

While image dimensions and aspect ratios may vary, it would be useful for the spec to define a place to look for screenshots (similar to the icons object)

This could be an optional member of the manifest, but including a standard place to look for screenshots would be useful in building catalogs of web apps such as https://pwa.rocks/

marcoscaceres:


marcoscaceres:
@patrickkettner, @RobDolinMS, wdyt?

patrickkettner:
🎉 🎉 🎉 

looks great, two thoughts before the big green button gets pushed

- should there be an optional `platform` or `store` property. Like if Apple only wanted to show iOS style screenshots, or have xbox specific screenshots for the windows store, etc.

- do we want to lock in screenshot as the noun here, in case we want to support video or other content in the future? I would assume the answer is yes, I just want to make sure before we etch it into stone [:

marcoscaceres:
> should there be an optional platform or store property. Like if Apple only wanted to show iOS style screenshots, or have xbox specific screenshots for the windows store, etc.

See #405 ... I think I'll just add `platform` and reuse the existing semantics of platform in the spec till we sort something out.  

> do we want to lock in screenshot as the noun here, in case we want to support video or other content in the future? I would assume the answer is yes, I just want to make sure before we etch it into stone [:

I think using screenshot for both video and static images is ok. People know what they mean and can be differentiated by `type`.

patrickkettner:
LGTM

marcoscaceres:
@patrickkettner thanks for the review! 

@kenchris, if you think it's ok - please merge.

index.html
```diff
@@ -156,7 +156,16 @@
...
   "orientation": "landscape",
   "theme_color": "aliceblue",
-  "background_color": "red"
+  "background_color": "red",
+  "screenshots": [{
+    "src": "screenshots/in-game-1x.jpg",
+    "sizes": "640x480",
+    "type": "image/jpeg"
+  },{
+    "src": "screenshots/in-game-2x.jpg",
```
kenchris: Don't you want one of the examples to use "platform"?

marcoscaceres: yeah, that's probably a good idea... but I'd have to use a fake one.

PR #491, opened 2016-08-25
feat(platform): allow platform on image obj (closes #361)

### LINKED ISSUE #361: Ability to define platform-specific icon/splash style
Different platforms have different guidelines regarding their icon style, and an app may need to follow them to look "native". 

Wrong style, even if slightly wrong, can create "uncanny valley", e.g. multiple OSes use square icons with rounded corners, but with a different radius. Icon with a wrong corner radius looks like an amateurish knock-off.

The spec currently only matches icons on size and density, but not style/theme, so sites would need to resort to user-agent sniffing if they wanted to serve system-specific icons.

A theoretical solution available in HTML, with help of [Moz's nonstandard media queries](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Media_queries), is:

    <link rel=icon href="generic.png">
    <link rel=icon media="all and -moz-windows-theme:luna-blue" href="xp.png">
    <link rel=icon media="all and -moz-windows-theme:aero" href="vista.png">

marcoscaceres:


marcoscaceres:
cc @pornel, @patrickkettner

index.html
```diff
@@ -2501,6 +2501,71 @@ <h3 id="applying">
     </section>
     <section>
       <h2>
+        Multi-purpose members
+      </h2>
+      <p>
+        This section specifies members that can be used on multiple objects.
+        Each member specifies with object a member can be used with.
```
kenchris: which?

index.html
```diff
@@ -2501,6 +2501,71 @@ <h3 id="applying">
...
+        <p>
+          A <dfn>platform</dfn> represents a software distribution ecosystem or
+          possibly an operating system.
+        </p>
+        <p class="note">
+          This specification does not define the particular values for a the
+          <a><code>platform</code> member</a>. However, the working group
+          maintains a <a href=
+          "https://github.com/w3c/manifest/wiki/Platforms">list of known
+          platform values</a> in our wiki.
```
kenchris: Should we add some encouragement about not adding too many of these? or announcing them on our mailing list?

kenchris:
lgtm

index.html
marcoscaceres: I suspect there will only be like 3 or 4. If it gets out of hand, we should add something.

index.html
marcoscaceres: D'oh, missed this one.

marcoscaceres: fixing locally.

PR #493, opened 2016-08-25
WIP: purpose

marcoscaceres:
@beverloo how about something like:
https://rawgit.com/w3c/manifest/purpose/index.html#purpose-member

Still rough... but looking for feedback.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2209,6 +2209,14 @@ Content-Type: application/manifest+json
         to display on the end-user's device or most closely matches the
         end-user's preferences.
       </p>
+      <p>
+        User agents may modify an <a>image object</a> to better match the
+        platform’s visual style before displaying it to the user, for example
+        by rounding the corners or painting it in a specific color. It is
+        recommended that developers prepare their image objects for such
+        scenarios to avoid losing important information through, e.g., change
+        of color or clipped corners.
+      </p>
       <section>
         <h3>
           Fetching image objects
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2185,6 +2185,14 @@ <h3 id="applying">
         to display on the end-user's device or most closely matches the
         end-user's preferences.
       </p>
+      <p>
+        User agents may modify an <a>image object</a> to better match the
+        platform’s visual style before displaying it to the user, for example
+        by rounding the corners or painting it in a specific color. It is
```
kenchris: applying filters?

index.html
```diff
@@ -2185,6 +2185,14 @@ <h3 id="applying">
...
+        recommended that developers prepare their image objects for such
+        scenarios to avoid losing important information through, e.g., change
+        in of color or clipped corners.
```
kenchris: in of?

index.html
```diff
@@ -2276,22 +2284,133 @@ <h3 id="applying">
       </section>
       <section>
         <h3>
+          <code>purpose</code> member
+        </h3>
+        <p>
+          The <dfn><code>purpose member</code></dfn> is an <a>unordered set of
+          unique space-separated tokens</a> that are <a>ASCII
+          case-insensitive</a>. The allowed values the <a>icon purposes</a>.
```
kenchris: The allowed values the <a>icon purposes</a>. I don't understand. Missing "are:"?

index.html
```diff
@@ -2276,22 +2284,133 @@ <h3 id="applying">
...
+        </p>
+        <p>
+          When an <a>image object</a> is used as an <a>icon</a>, a developer
+          can specify that the image is intended to serve some special
+          <dfn>purpose</dfn> in the context of the host OS (i.e., for better
+          integration).
```
kenchris: UAs are fit to ignore these values?

index.html
```diff
@@ -2276,22 +2284,133 @@ <h3 id="applying">
...
+        <p>
+          The <dfn>icon purposes</dfn> are as follows:
+        </p>
+        <dl>
+          <dt>
+            <dfn data-lt="badge purpose">badge</dfn>:
+          </dt>
+          <dd>
+            A user agent can present this icon where space constraints and/or
+            color requirements differ from those of the application icon.
```
kenchris: Badge icons are intended to be applied on top of other image assets such as profile photos, in such cases where these might be more important than the app icon, e.g. when receiving a notification.

index.html
```diff
@@ -2276,22 +2284,133 @@ <h3 id="applying">
...
+          <li>If <var>type</var> is not "string", then:
+            <ol>
+              <li>If <var>type</var> is not "<code>undefined</code>", <a>issue
+              a developer warning</a> that the type is unsupported.
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          </li>
+          <li>Let <var>keywords</var> be the result of <a>splitting</a>
```
kenchris: It is JSON, why don't we do an array.... remembering viewport meta, devs might not get this right, ie... double space, comma, semi colon etc

marcoscaceres: yeah, you are probably right.

marcoscaceres: I gave this more thought, and I think I want to stick with a string here - because it's rare that something will have multiple purpose (and we only define one purpose). Should probably only use arrays for complex types.

index.html
marcoscaceres: I'd like to see the implementation and uses a bit more before adding this text. I've only seen the Android and Safari cases, and those seem to show the badges on their own.

index.html
marcoscaceres: I'm changing the wording to "can hint"

marcoscaceres:
Fixed commit message

PR #496, opened 2016-08-26
fix(display modes): refine standalone (closes #495)

### LINKED ISSUE #495: Standalone description not accurate
Currently our description of display:standalone says:
"In this mode, the user agent will exclude UI elements for controlling navigation, but can include other UI elements such as a status bar."

In fact, on all browsers I'm aware of on Android the UI for controlling navigation is just system UI that isn't obscured. Hence it may be clearer to say:

"In this mode, the user agent will exclude standard browser UI elements such as an URL bar, but can include other system UI elements such as a status bar or system back button."

marcoscaceres:


marcoscaceres:
Folks, wdyt? Does this match your expectations or standalone?

index.html
```diff
@@ -656,8 +656,9 @@ <h3 id="navigation-scope-security-considerations">
           Opens the web application to look and feel like a standalone native
           application. This can include the application having a different
           window, its own icon in the application launcher, etc. In this mode,
-          the user agent will exclude UI elements for controlling navigation,
-          but can include other UI elements such as a status bar.
+          the user agent will exclude standard browser UI elements such as an
+          URL bar, but can include other system UI elements such as a status
+          bar or system back button.
```
kenchris: "and/or" or just "and"

marcoscaceres: and/or

PR #499, opened 2016-09-13
[Appendix] Add crawl opt-out (closes #498)

### LINKED ISSUE #498: Describe how developers can block web crawlers
There may be a case where a web app developer does not want their web app to be crawled.  I was thinking proposing an additional element to support this, but realized the web already has a way to block crawlers.  

Below is some draft language, if this seems reasonable, please 👍 and I'll submit a PR.

> Like other web resources, a web app manifest should be accessible to any web browser or web crawler.
>  
> If a web app developer wants to block traffic from web crawlers, the developer MAY do so by including the web app manifest in a robots.txt file.  This is further described in the [robots.txt](http://www.robotstxt.org/) protocol.

RobDolinMS:
Adding an Appendix section of developer guidance including how developers can opt-out of web crawl.

Also adds a placeholder for some additional examples.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3597,5 +3597,32 @@ Content-Security-Policy: img-src icons.example.com
         specification via the HTML5Apps project.
       </p>
     </section>
+    <section id="developer-guidance"></section>
+    <section class="appendix">
+      <h2>
+        Developer Guidance
+      </h2>
+      <p>
+        This section provides non-normative guidance for developers of web app manifest files 
+      </p>
+      <h3>
+        Additional Examples
+      </h3>
+      <p>
+        TODO: Add additional examples
+      </p>
+      <h3>
+        Interaction With Web Crawlers
+      </h3>
+      <p>
+        Like other web resources, a web app manifest should be accessible to any web browser or web crawler.
+      </p>
+      <p>
+        If a web app developer wants to block traffic from web crawlers, 
+        the developer MAY do so by including the web app manifest in a robots.txt file. 
+        This is further described in the <a href="http://www.robotstxt.org/">robots.txt</a> protocol.
+        A web app developer could also use the <code>X-Robots-Tag</code> HTTP header.
+      </p>
+    </section>
   </body>
 </html>


index.html
```diff
@@ -3597,5 +3597,32 @@ <h3 id="applying">
...
+        Developer Guidance
+      </h2>
+      <p>
+        This section provides non-normative guidance for developers of web app manifest files 
+      </p>
+      <h3>
+        Additional Examples
+      </h3>
+      <p>
+        TODO: Add additional examples
```
kenchris: I think there is a special way to add the TODOs with respec

marcoscaceres:
I'd still prefer we do this in an explainer.md file. This stuff gets out of date pretty quick, so it's better to have it in file that we can update independently of the spec.

RobDolinMS:
Thanks @marcoscaceres.  I'll re-propose to add as a related explainer.md file instead of as an appendix.  

Thanks also to @MichaelChampion for explaining some of the normative vs. non-normative issues.

index.html
```diff
@@ -3597,5 +3597,32 @@ <h3 id="applying">
...
+      </p>
+      <h3>
+        Interaction With Web Crawlers
+      </h3>
+      <p>
+        Like other web resources, a web app manifest should be accessible to any web browser or web crawler.
+      </p>
+      <p>
+        If a web app developer wants to block traffic from web crawlers,
```
alrra: I personally think that  _`"wants to block traffic"`_ is not the correct wording here, and instead, something in the lines of _`"wants to inform"`_ should be used.

From http://www.robotstxt.org/faq/nosecurity.html: 

> .../robots.txt is not intended for access control, so don't try to use it as such. Think of it as a "No Entry" sign, not a locked door.

RobDolinMS:
Replaced by PR #505

index.html
RobDolinMS: Thanks @alrra.  I've replaced this PR with #499 and incorporated your suggestion of "block traffic" to "inform"

PR #503, opened 2016-10-07
BREAKING CHANGE: install event renamed appinstalled (closes #502)

### LINKED ISSUE #502: Rename "install" event to "appinstalled" to avoid SW event name clash
The Web App Manifest spec defines an event type [`"install"`](https://www.w3.org/TR/appmanifest/#oninstall-attribute) which is fired at the *window* object. The Service Workers spec defines an event type [`"install"`](https://www.w3.org/TR/service-workers/#service-worker-global-scope-install-event) which is fired at the `ServiceWorkerGlobalScope` object.

While these do not directly conflict (as they are fired at different objects), there is potential for confusion (for both web developers and browser developers). As one anecdote, in Chromium, the event objects for these two types of event are indistinguishable, so we have to take context into account (e.g., when logging the relative usage of these two features) to avoid conflating the two event types.

Since the concept of "installing" a service worker and "installing" a web app are totally unrelated, perhaps it is worth renaming this event to something different. Suggestions: `"register"`, `"installapp"`.

marcoscaceres:


marcoscaceres:
@mgiuca, PTAL.

index.html
```diff
@@ -419,37 +419,38 @@ <h3 id="installation-sec">
         </p>
         <pre class="idl">
           partial interface Window {
-            attribute EventHandler oninstall;
+            attribute EventHandler onappinstalled;
           };
         </pre>
         <div class="example">
           <p>
             This example shows two ways of handling the install event.
```
mgiuca: s/install/appinstalled

marcoscaceres: fixed

dominickng:
Seems fine to me with mgiuca's comment addressed.

patrickkettner:
lgtm

marcoscaceres:
Ok, updating Gecko now.

PR #505, opened 2016-10-12
[Primer] Create file and add text re: web cralwers

RobDolinMS:
This replaces PR #499


mgiuca:
nit: Replace "cralwers" with "crawlers" in the commit text.

PR #506, opened 2016-10-13
WIP: beforeinstallprompt

marcoscaceres:
Nowhere near ready... will hopefully finish tomorrow.

The existing install section needs a total rewrite :(

### DIFF
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -35,7 +35,6 @@
    *
    */
   class BeforeInstallPromptEvent extends Event {
-
     constructor(typeArg, eventInit) {
       // WebIDL Guard. Not in spec, as it's all handled by WebIDL.
       if (arguments.length === 0) {
@@ -59,7 +58,7 @@
         userChoiceResolver: null, // Implicit in spec
       };
 
-      internal.userChoice = new Promise((resolve) => {
+      internal.userChoice = new Promise(resolve => {
         if (eventInit && "userChoice" in eventInit) {
           return resolve(eventInit.userChoice);
         }
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -1,132 +1,154 @@
-/*globals BeforeInstallPromptEvent, DOMException*/
+/*globals DOMException*/
 "use strict";
-const internalSlots = new WeakMap();
-const installProcesses = [];
-const AppBannerPromptOutcome = new Set([
-  "accepted",
-  "dismissed",
-]);
+{
+  const internalSlots = new WeakMap();
+  const installProcesses = [];
+  const AppBannerPromptOutcome = new Set([
+    "accepted",
+    "dismissed",
+  ]);
 
-async function requestInstallPrompt(event, element) {
-  const internal = internalSlots.get(event);
-  internal.didPrompt = true;
-  const promptOutcome = await showInstallPrompt(element);
-  internal.userChoiceResolver(promptOutcome);
-}
+  /**
+   * Emulates the UI thread firing the event.
+   * @param  {BeforeInstallPromptEvent} event
+   * @param  {Element} element A DOM element that simulates the pop-up.
+   * @return {Promise<void>} Resolves when user makes a choice.
+   */
+  async function requestInstallPrompt(event, element) {
+    const internal = internalSlots.get(event);
+    internal.didPrompt = true;
+    const promptOutcome = await showInstallPrompt(element);
+    internal.userChoiceResolver(promptOutcome);
+  }
 
-class BeforeInstallPromptEvent extends Event {
-  constructor(typeArg, eventInit) {
-    // WebIDL Guard. Not in spec, as it's all handled by WebIDL.
-    if (arguments.length === 0) {
-      throw new TypeError("Not enough arguments. Expected at least 1.");
-    }
-    const initType = typeof eventInit;
-    if (arguments.length === 2 && initType !== "undefined" && initType !== "object") {
-      throw new TypeError("Value can't be converted to a dictionary.");
-    }
-    super(typeArg, Object.assign({ cancelable: true }, eventInit));
+  /**
+   * Implementation of BeforeInstallPromptEvent.
+   *
+   */
+  class BeforeInstallPromptEvent extends Event {
 
-    if (eventInit && typeof eventInit.userChoice !== "undefined" && !AppBannerPromptOutcome.has(String(eventInit.userChoice))) {
-      const msg = `The provided value '${eventInit.userChoice}' is not a valid` +
-        "enum value of type AppBannerPromptOutcome.";
-      throw new TypeError(msg);
-    }
-    // End WebIDL guard.
-    const internal = {
-      didPrompt: false,
-      userChoiceResolver: null,
-      userChoice: null,
-    };
-
-    internal.userChoice = new Promise((resolve) => {
-      if (eventInit && "userChoice" in eventInit) {
-        return resolve(eventInit.userChoice);
+    constructor(typeArg, eventInit) {
+      // WebIDL Guard. Not in spec, as it's all handled by WebIDL.
+      if (arguments.length === 0) {
+        throw new TypeError("Not enough arguments. Expected at least 1.");
       }
-      internal.userChoiceResolver = resolve;
-    });
-    internalSlots.set(this, internal);
-  }
-  prompt() {
-    // if (this.isTrusted === false) {
-    //   const msg = "Untrusted events can't call prompt().";
-    //   throw new DOMException(msg, "NotAllowedError");
-    // }
+      const initType = typeof eventInit;
+      if (arguments.length === 2 && initType !== "undefined" && initType !== "object") {
+        throw new TypeError("Value can't be converted to a dictionary.");
+      }
+      super(typeArg, Object.assign({ cancelable: true }, eventInit));
 
-    if (this.defaultPrevented === false) {
-      const msg = ".prompt() needs to be called after .preventDefault()";
-      throw new DOMException(msg, "InvalidStateError");
+      if (eventInit && typeof eventInit.userChoice !== "undefined" && !AppBannerPromptOutcome.has(String(eventInit.userChoice))) {
+        const msg = `The provided value '${eventInit.userChoice}' is not a valid` +
+          "enum value of type AppBannerPromptOutcome.";
+        throw new TypeError(msg);
+      }
+      // End WebIDL guard.
+      const internal = {
+        didPrompt: false,
+        userChoice: null,
+        userChoiceResolver: null, // Implicit in spec
+      };
+
+      internal.userChoice = new Promise((resolve) => {
+        if (eventInit && "userChoice" in eventInit) {
+          return resolve(eventInit.userChoice);
+        }
+        internal.userChoiceResolver = resolve;
+      });
+      internalSlots.set(this, internal);
     }
 
-    if (internalSlots.get(this).didPrompt) {
-      const msg = ".prompt() can only be successfully called once.";
-      throw new DOMException(msg, "InvalidStateError");
+    prompt() {
+      if (this.isTrusted === false) {
+        const msg = "Untrusted events can't call prompt().";
+        throw new DOMException(msg, "NotAllowedError");
+      }
+
+      if (this.defaultPrevented === false) {
+        const msg = ".prompt() needs to be called after .preventDefault()";
+        throw new DOMException(msg, "InvalidStateError");
+      }
+
+      if (internalSlots.get(this).didPrompt) {
+        const msg = ".prompt() can only be successfully called once.";
+        throw new DOMException(msg, "InvalidStateError");
+      }
+      requestInstallPrompt(this);
     }
-    requestInstallPrompt(this);
+
+    get userChoice() {
+      return internalSlots.get(this).userChoice;
+    }
   }
 
-  get userChoice() {
-    return internalSlots.get(this).userChoice;
-  }
-}
-
-async function notifyBeforeInstallPrompt(element) {
-  await trackReadyState();
-  if (installProcesses.length) {
-    return;
-  }
-  const event = new BeforeInstallPromptEvent("beforeinstallprompt");
-  window.dispatchEvent(event);
-  if (!event.defaultPrevented) {
-    await requestInstallPrompt(event, element);
-  }
-}
-
-function trackReadyState() {
-  return new Promise((resolve) => {
+  async function notifyBeforeInstallPrompt(element) {
     if (document.readyState === "complete") {
-      return resolve();
+      await trackReadyState();
     }
-    document.addEventListener("readystatechange", () => {
-      if (document.readyState === "complete") {
-        resolve();
-      }
-    });
-  });
-}
-
-async function showInstallPrompt(button) {
-  if (button) {
-    button.disabled = true;
+    if (installProcesses.length) {
+      return;
+    }
+    const event = new BeforeInstallPromptEvent("beforeinstallprompt");
+    window.dispatchEvent(event);
+    if (!event.defaultPrevented) {
+      await requestInstallPrompt(event, element);
+    }
   }
-  await Promise.all(installProcesses);
-  const prompt = document.createElement("fieldset");
-  prompt.id = "installprompt";
-  prompt.innerHTML = `
-    <h2>Add to Home screen</h2>
-    <p>... the foo app ...</p>
-    <button id="cancel">CANCEL</button>
-    <button id="add">ADD</button>
-  `;
-  const cancel = prompt.querySelector("#cancel");
-  const add = prompt.querySelector("#add");
-  document.body.appendChild(prompt);
-  const p = new Promise((resolve) => {
-    add.addEventListener("click", () => {
-      resolve("accepted");
-      //emulate installation to home screen
-      setTimeout(() => {
-        window.dispatchEvent(new Event("appinstalled"));
-      }, 1000);
+
+  function trackReadyState() {
+    return new Promise((resolve) => {
+      document.addEventListener("readystatechange", () => {
+        if (document.readyState === "complete") {
+          resolve();
+        }
+      });
     });
-    cancel.addEventListener("click", () => resolve("dismissed"));
-  });
-  cancel.onclick = add.onclick = function() {
-    document.body.removeChild(prompt);
-    installProcesses.splice(installProcesses.findIndex(item => item === p), 1);
+  }
+
+  async function showInstallPrompt(button) {
     if (button) {
-      button.disabled = false;
+      button.disabled = true;
     }
-  };
-  installProcesses.push(p);
-  return p;
+    await Promise.all(installProcesses);
+    const prompt = document.createElement("fieldset");
+    prompt.id = "installprompt";
+    prompt.innerHTML = `
+      <h2>Add to Home screen</h2>
+      <p>... the foo app ...</p>
+      <button id="cancel">CANCEL</button>
+      <button id="add">ADD</button>
+    `;
+    const cancel = prompt.querySelector("#cancel");
+    const add = prompt.querySelector("#add");
+    document.body.appendChild(prompt);
+    const p = new Promise((resolve) => {
+      add.addEventListener("click", () => {
+        resolve("accepted");
+        //emulate installation to home screen
+        setTimeout(() => {
+          window.dispatchEvent(new Event("appinstalled"));
+        }, 1000);
+      });
+      cancel.addEventListener("click", () => resolve("dismissed"));
+    });
+    cancel.onclick = add.onclick = function() {
+      document.body.removeChild(prompt);
+      installProcesses.splice(installProcesses.findIndex(item => item === p), 1);
+      if (button) {
+        button.disabled = false;
+      }
+    };
+    installProcesses.push(p);
+    return p;
+  }
+
+  if(!window.BeforeInstallPromptEvent) {
+    window.BeforeInstallPromptEvent = BeforeInstallPromptEvent;
+  } else {
+    console.warn("Using browser's implementation of BeforeInstallPromptEvent.");
+  }
+
+  window.notifyBeforeInstallPrompt = notifyBeforeInstallPrompt;
+  window.showInstallPrompt = showInstallPrompt;
 }
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -26,7 +26,7 @@ class BeforeInstallPromptEvent extends Event {
     }
     // End WebIDL guard.
     const internal = {
-      prompState: "idle", // "prompting", "done"
+      didPrompt: false,
     };
 
     internal.userChoice = new Promise((resolve) => {
@@ -44,27 +44,24 @@ class BeforeInstallPromptEvent extends Event {
       const msg = "Untrusted events can't call prompt().";
       throw new DOMException(msg, "NotAllowedError");
     }
-    let msg = "";
-    switch (internalSlots.get(this).promptState) {
-      case "done":
-        msg = ".prompt() has expired.";
-        throw new DOMException(msg, "InvalidStateError");
-      case "prompting":
-        msg = "Already trying to prompt.";
-        throw new DOMException(msg, "InvalidStateError");
-      default:
-        if (this.defaultPrevented === false) {
-          msg = ".prompt() needs to be called after .preventDefault()";
-          throw new DOMException(msg, "InvalidStateError");
-        }
-        internalSlots.get(this).promptState = "prompting";
+
+    if (this.defaultPrevented === false) {
+      const msg = ".prompt() needs to be called after .preventDefault()";
+      throw new DOMException(msg, "InvalidStateError");
     }
 
+    if(internalSlots.get(this).didPrompt){
+      const msg = ".prompt() can only be succesfully called once.";
+      throw new DOMException(msg, "InvalidStateError"); 
+    }
+
+    internalSlots.get(this).didPrompt = true;
+    
     (async function task() {
-      const promptOutcome = await showInstallPrompt();
-      internalSlots.get(this).promptState = "done";
+      const promptOutcome = await showInstallPrompt();      
       internalSlots.get(this).userChoiceHandlers.resolve(promptOutcome);
     }.bind(this)())
+  
   }
 
   get userChoice() {
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -1,6 +1,5 @@
 /*globals DOMException*/
-"use strict";
-{
+"use strict"; {
   const internalSlots = new WeakMap();
   const installProcesses = [];
   const AppBannerPromptOutcome = new Set([
@@ -21,6 +20,16 @@
     internal.userChoiceResolver(promptOutcome);
   }
 
+  function hasValidUserChoice({ userChoice }) {
+    if (typeof userChoice === "undefined") {
+      return true;
+    }
+    if (!AppBannerPromptOutcome.has(String(userChoice))) {
+      return false;
+    }
+    return true;
+  }
+
   /**
    * Implementation of BeforeInstallPromptEvent.
    *
@@ -38,9 +47,9 @@
       }
       super(typeArg, Object.assign({ cancelable: true }, eventInit));
 
-      if (eventInit && typeof eventInit.userChoice !== "undefined" && !AppBannerPromptOutcome.has(String(eventInit.userChoice))) {
+      if (eventInit && !hasValidUserChoice(eventInit)) {
         const msg = `The provided value '${eventInit.userChoice}' is not a valid` +
-          "enum value of type AppBannerPromptOutcome.";
+          " enum value of type AppBannerPromptOutcome.";
         throw new TypeError(msg);
       }
       // End WebIDL guard.
@@ -83,8 +92,8 @@
   }
 
   async function notifyBeforeInstallPrompt(element) {
-    if (document.readyState === "complete") {
-      await trackReadyState();
+    if (document.readyState !== "complete") {
+      await waitUntilReadyStateComplete();
     }
     if (installProcesses.length) {
       return;
@@ -96,8 +105,8 @@
     }
   }
 
-  function trackReadyState() {
-    return new Promise((resolve) => {
+  function waitUntilReadyStateComplete() {
+    return new Promise(resolve => {
       document.addEventListener("readystatechange", () => {
         if (document.readyState === "complete") {
           resolve();
@@ -125,7 +134,7 @@
     const p = new Promise((resolve) => {
       add.addEventListener("click", () => {
         resolve("accepted");
-        //emulate installation to home screen
+        // Emulate installation to home screen
         setTimeout(() => {
           window.dispatchEvent(new Event("appinstalled"));
         }, 1000);
@@ -143,7 +152,7 @@
     return p;
   }
 
-  if(!window.BeforeInstallPromptEvent) {
+  if (!window.BeforeInstallPromptEvent) {
     window.BeforeInstallPromptEvent = BeforeInstallPromptEvent;
   } else {
     console.warn("Using browser's implementation of BeforeInstallPromptEvent.");
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -98,7 +98,7 @@ window.addEventListener("beforeinstallprompt", async function(event) {
   // Emulate blocking tasks
   await new Promise((res) => setTimeout(res, 1000));
   try {
-    await event.prompt();
+    event.prompt();
   } catch (err) {
     console.error(err);
   }
--- a/implementation/BeforeInstallPromptEvent.js
+++ b/implementation/BeforeInstallPromptEvent.js
@@ -29,9 +29,10 @@ class BeforeInstallPromptEvent extends Event {
       didPrompt: false,
     };
 
-    internal.userChoice = new Promise((resolve) => {
+    internal.userChoice = new Promise((resolve, reject) => {
       internal.userChoiceHandlers = {
         resolve,
+        reject,
       };
       if (eventInit && "userChoice" in eventInit) {
         resolve(eventInit.userChoice);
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -233,6 +233,59 @@
       </p>
       <section>
         <h3>
+          Installing via automated prompting
```
kenchris: Installing by prompting automatically

index.html
```diff
@@ -233,6 +233,59 @@
...
+        </h3>
+        <p>
+          User agents that support an <a>automated install prompt</a> MUST run
```
kenchris: UAs may support ... in which case it MUST run...

index.html
```diff
@@ -233,6 +233,59 @@
...
+          the following steps before initiating an install process:
+        </p>
+        <ol>
+          <li>If a <a>manually</a> instantiated <a>installation process</a> for
+          this web application, terminate this algorithm.
+          </li>
+          <li>Let <var>event</var> be a newly constructed
+          <a>BeforeInstallPromptEvent</a> event named "beforeinstallprompt",
+          which does not bubble and is cancellable.
```
kenchris: , and which is

index.html
```diff
@@ -233,6 +233,59 @@
...
+          </li>
+          <li>Let <var>prompt promise</var> be the promise corresponding to the
+          event's prompt() method.
+          </li>
+          <li>Let <var>user-choice promise</var> be the promise corresponding
+          to the event's <code>prompt()</code> method.
+          </li>
+          <li>Fire <var>event</var> at the <code>Window</code> object.
```
kenchris: on the?

index.html
```diff
@@ -233,6 +233,59 @@
...
+          </li>
+          <li>Queue a task on the Realm’s settings object’s responsible event
```
kenchris: should realm be uppercase? why?

index.html
```diff
@@ -233,6 +233,59 @@
...
+          loop:
+            <ol>
+              <li>If <var>event</var>'s <a>cancel flag</a> is set, terminate
+              this algorithm.
+              </li>
+              <li>Otherwise, attempt to present the end-user with a UI that
```
kenchris: why attempt? You need to do it right?

kenchris: or it may?

index.html
```diff
@@ -233,6 +233,59 @@
...
+              lets them choose if they would like to install the web
+              application.
+              </li>
+              <li>If the previous step succeeds, resolve <var>prompt
+              promise</var>. Otherwise, reject <var>prompt promise</var> with
+              an ""AbortError", or "OperationError", or "NotAllowedError", or
+              "UnknownError" (depending on why the operation failed).
```
kenchris: This may be UA determined?

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+};
+
+dictionary BeforeInstallPromptEventInit : EventInit {
+  AppBannerPromptOutcome userChoice;
+};
+      </pre>
+        <p>
+          The <dfn><code>BeforeInstallPromptEvent</code></dfn> is dispatched
+          prior to activating an <a>automated install prompt</a>, allowing a
+          developer to prevent the default action for an install prompt.
```
kenchris: instead of a

index.html
marcoscaceres: I'm not a huge fan of MAY ... MUST constructions. Though we need to make it clear that user agent MAY support automated install prompts elsewhere. It's part of the big rewrite this whole install section needs.

index.html
marcoscaceres: DOM spec uses "Fire Event X at Y", so will follow that.

marcoscaceres:
Thanks for the initial comments @kenchris! Much appreciated.

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+};
+
+[Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+interface BeforeInstallPromptEvent : Event {
+    readonly attribute Promise&lt;AppBannerPromptOutcome&gt; userChoice;
+    Promise&lt;void&gt; prompt();
+};
+
+dictionary BeforeInstallPromptEventInit : EventInit {
+  AppBannerPromptOutcome userChoice;
```
mgiuca: This is something you've added here (that isn't in Chrome), right?

TBH it's not clear to me why we want to allow web developers to construct their own `BeforeInstallPromptEvent` objects that aren't being delivered by the UA. There isn't much use to this. If we do want to allow it for completeness, we need to make sure that it's not possible to use these fabricated BIPE objects to show a prompt -- more on that later.

It's strange that we go to the effort of allowing the website to specify a `userChoice` here (to simulate a real BIPE) but we don't provide any mechanism for them to override the `prompt` method. I suppose they could just subclass `BeforeInstallPromptEvent`, but if we allow that, then they can also override `userChoice` in the subclass. So I still don't see the point of this.

But if we must let the client supply a userChoice here, why is it an `AppBannerPromptOutcome` and not a `Promise<AppBannerPromptOutcome>`? The way it is now, the client has to pass the user's actual choice (accepted or dismissed) to the constructor of the event, before the simulated prompt is even shown, and the promise gets immediately resolved during construction. Shouldn't this constructor take a promise that can be resolved at a later time?

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+          default action (via <code>.preventDefault()</code>) prevents the user
+          agent from presenting the automated install prompt.
+        </p>
+        <p>
+          To create an instance of the <code>BeforeInstallPromptEvent</code>
+          interface, run the <dfn>steps to construct a BeforeInstallPromptEvent
+          event</dfn>:
+        </p>
+        <ol>
+          <li>Let <var>userChoice</var> be null.
```
mgiuca: These meta-variable names are confusing. *userChoicePromise* is `userChoice` (the `Promise<AppBannerPromptOutcome>`), and *userChoice* is the outcome that the promise is resolved with (the `AppBannerPromptOutcome`). That's messing with my head.

How about rename *userChoice* to *promptOutcome* (the outcome the promise is resolved with), and rename *userChoicePromise* to *userChoice* (the promise).

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+          <ol>
+            <li>If the event's cancel flag is not set, return a promise
+            rejected with an "InvalidStateError".
+            </li>
+            <li>If [[\]], return a promise rejected with an
+            "InvalidStateError".
+            </li>
+            <li>Let <var>p</var> be a new promise.
+            </li>
+            <li>Run the following steps in parallel:
```
mgiuca: Missing steps. Should refer to the above "steps before initiating an install process".

index.html
```diff
@@ -233,6 +233,59 @@
...
       <section>
         <h3>
+          Installing via automated prompting
+        </h3>
+        <p>
+          User agents that support an <a>automated install prompt</a> MUST run
+          the following steps before initiating an install process:
+        </p>
+        <ol>
+          <li>If a <a>manually</a> instantiated <a>installation process</a> for
```
mgiuca: Why not just make this algorithm the "steps for an automatically instantiated installation process" and drop this first early-exit step?

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+            following steps:
+          </p>
+          <ol>
+            <li>If the event's cancel flag is not set, return a promise
+            rejected with an "InvalidStateError".
+            </li>
+            <li>If [[\]], return a promise rejected with an
+            "InvalidStateError".
+            </li>
+            <li>Let <var>p</var> be a new promise.
```
mgiuca: As discussed above, we need to make sure that the following code *fails*:

```js
var e = new BeforeInstallPromptEvent({});
e.preventDefault();
e.prompt()
    .then(() => console.log('prompt() succeeded'))
    .catch(e => console.error('prompt() failed: ' + e));
```

That is, you must not be able to fabricate a `BeforeInstallPromptEvent` object, call `preventDefault` on it (to set its cancel flag), then call `prompt` on it to show an install banner without the UA's permission.

As I said above, I'm not sure why we allow these events to be fabricated at all, but if we do, there needs to be a check that this event was created by the UA and not by the site itself.

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+            <li>Return <var>p</var>.
+            </li>
+          </ol>
+        </section>
+        <section>
+          <h4>
+            <code>userChoice</code> attribute
+          </h4>
+          <p>
+            The <dfn><code>userChoice</code> attribute</dfn>, when getting,
```
mgiuca: "when getting"? Can we reword this?

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+        </h3>
+        <pre class="idl">
+enum AppBannerPromptOutcome {
+  "accepted",
+  "dismissed"
+};
+
+[Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+interface BeforeInstallPromptEvent : Event {
+    readonly attribute Promise&lt;AppBannerPromptOutcome&gt; userChoice;
```
mgiuca: I find the fact that this is a promise attribute (as opposed to a function that returns a promise) really weird. I know it's what Chrome does. I wonder if it's too late to change it?

It causes a lot of weirdness because there are side-effects upon reading the attribute. For example, in Chrome a user-constructed `BeforeInstallPromptEvent` logs a console error just by the event object itself being printed to the console, because it tries to read the `userChoice` attribute. And [this](https://bugs.chromium.org/p/chromium/issues/detail?id=655877).

index.html
```diff
@@ -410,6 +463,94 @@ <h3 id="installation-sec">
...
+            </li>
+          </ol>
+        </section>
+        <section>
+          <h4>
+            <code>userChoice</code> attribute
+          </h4>
+          <p>
+            The <dfn><code>userChoice</code> attribute</dfn>, when getting,
+            returns a promise.
```
mgiuca: returns *userChoicePromise*.

I know that variable is kind scoped to the `BeforeInstallPromptEvent` constructor... hmm. I think we need some way to refer to these promises across the different algorithms. There's also weirdness that the "steps before initiating an install process" refer to these promises as "the promise corresponding to the event's *X* method", rather than by name.

index.html
```diff
@@ -441,6 +583,17 @@ <h3 id="installation-sec">
         </div>
         <section>
           <h4>
+            <code>onbeforeinstallprompt</code> attribute
```
mgiuca: Should this come after onappinstalled, to match the order they appear in the IDL?

index.html
marcoscaceres: Realm is basically the global object: 
https://tc39.github.io/ecma262/#sec-code-realms

index.html
```diff
@@ -233,6 +233,59 @@
...
+              <li>If <var>event</var>'s <a>cancel flag</a> is set, terminate
+              this algorithm.
+              </li>
+              <li>Otherwise, attempt to present the end-user with a UI that
+              lets them choose if they would like to install the web
+              application.
+              </li>
+              <li>If the previous step succeeds, resolve <var>prompt
+              promise</var>. Otherwise, reject <var>prompt promise</var> with
+              an ""AbortError", or "OperationError", or "NotAllowedError", or
```
marcoscaceres: Yep, will add all the links... don't want to add them yet as will have to rewrite this a few times yet.

index.html
marcoscaceres: Right now, yes. Still need to clarify what that is for? Let's discuss that in the bug.

marcoscaceres: Yes, it's actually really irritating :) Prompt should just return the user choice. 

This would be so much nicer:  

```
e.prompt().then(obj => obj.userChoice);
```

index.html
```diff
@@ -233,6 +233,59 @@
...
+              "UnknownError" (depending on why the operation failed).
+              </li>
+              <li>In parallel, await the end-user's decision.
+              </li>
+              <li>Let <var>user choice</var> be the result of the previous
+              step, represented by either the string "accepted" or "dismissed".
```
marcoscaceres: I'll fix all markup related issues after we are happy with the overall text.

index.html
marcoscaceres: Agree. They are terrible :) Will change them.

index.html
marcoscaceres: I've not checked this in yet, but it now says:

> If the event's trusted flag is not set, return a promise rejected with a "NotAllowedError", optionally warning the developer that only trusted events are allowed to perfrom this operation.

index.html
marcoscaceres: Yep. Coming soon.

index.html
marcoscaceres: Agree. This is painful. 

My proposal would be that BIP has one internal "prompt" promise. That makes the whole design much more sensical. That promise resolves after prompt() succeeds and returns some kind of `UserPromptResult` object as proposed above.

index.html
marcoscaceres: Yeah, will move it.

index.html
marcoscaceres: >  I wonder if it's too late to change it?

I certainly wouldn't be opposed. I guess we need some telemetry data on usage to change it - and some coordination with existing sites that are using this. 

> I find the fact that this is a promise attribute (as opposed to a function that returns a promise) really weird. I know it's what Chrome does. I wonder if it's too late to change it?

There are lots of these in the platform, FWIW. https://github.com/heycam/webidl/issues/186#issuecomment-253526030

index.html
marcoscaceres: fixed

index.html
marcoscaceres: > But if we must let the client supply a userChoice here, why is it an AppBannerPromptOutcome and not a Promise<AppBannerPromptOutcome>? The way it is now, the client has to pass the user's actual choice (accepted or dismissed) to the constructor of the event, before the simulated prompt is even shown, and the promise gets immediately resolved during construction. Shouldn't this constructor take a promise that can be resolved at a later time?

If someone is synthetically constructing BIP Events, they probably know what they are doing - so I would prefer just to keep it as simple as possible: i.e., they provide the value "userChoice" that it resolves. If they provide no value, the promise never resolves. If it comes up as an issue, we can set a default value.

index.html
marcoscaceres: @kenchris, I'm not sure exactly what you are wanting to change here.

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
+/*globals BeforeInstallPromptEvent, DOMException*/
```
mgiuca: Should this file be surrounded in a closure so it hides all the other variables?

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    document.addEventListener("readystatechange", () => {
+      if (document.readyState === "complete") {
+        resolve();
+      }
+    });
+  });
+}
+
+window.addEventListener("beforeinstallprompt", async function(event) {
```
mgiuca: It's weird that this code is here. I think of this file as being a polyfill for `BeforeInstallPromptEvent`, so it provides the event object as well as the methods for simulating the system sending those events (`notifyBeforeInstallPrompt`). It shouldn't also have a handler for the event which automatically calls preventDefault and prompt.

It's even weirder because `index.html` has its own BIP handler which may race with this one.

implementation/index.html
```diff
@@ -0,0 +1,256 @@
...
+  } {
+    const assertion = "is trusted";
+    console.assert(ev.isTrusted, assertion, ev);
+  } {
+    const assertion = "calling prompt successfully returns void";
+    console.assert(typeof ev.prompt() === "undefined", assertion);
+  } {
+    const assertion = "Even with success, prompt() can only be called once";
+    try {
+      ev.prompt();
```
mgiuca: Why does this call `prompt` without first calling `preventDefault`?

I'm confused because you have another handler (in BIPE.js) that calls preventDefault and prompt. This seems to be testing that that event was handled first, before this one... but as we discussed previously the order isn't guaranteed. I think it would be best to have just a single bipe handler (in this file, not the other one) that tests all of these things.

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
+/*globals BeforeInstallPromptEvent, DOMException*/
+"use strict";
+const internalSlots = new WeakMap();
```
marcoscaceres: heh, I'm not a fan of the reverse :) However, once we put this into it's own scope, as suggested in your first comment, it won't matter. Hopefully JS will get real private members in the future.

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
...
+const installProcesses = [];
+const AppBannerPromptOutcome = new Set([
+  "accepted",
+  "dismissed",
+]);
+
+class BeforeInstallPromptEvent extends Event {
+  constructor(typeArg, eventInit) {
+    // WebIDL Guard. Wont be in spec in spec as it's all handled by WebIDL.
```
marcoscaceres: fixed

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
...
+      throw new TypeError(msg);
+    }
+    // End WebIDL guard.
+    const internal = {
+      didPrompt: false,
+    };
+
+    internal.userChoice = new Promise((resolve) => {
+      internal.userChoiceHandlers = {
+        resolve,
```
marcoscaceres: So, before it held .resolve and .reject. But will change to your suggestion:
```
internal.userChoiceResolver = resolve;
```

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
...
+    return internalSlots.get(this).userChoice;
+  }
+}
+
+async function notifyBeforeInstallPrompt(element) {
+  await trackReadyState();
+  if (installProcesses.length) {
+    return;
+  }
+  const event = new BeforeInstallPromptEvent("beforeinstallprompt");
```
marcoscaceres: Yeah, I was thinking of doing that (I've just been commenting out the checks while I play with this). Obviously, trusted gets set automatically when the real even is created by the browser through the binding layer. 

Within a few weeks, we will have both Chrome and Firefox implementing this for real in C++, so let's keep the current sample implementation as is.

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,143 @@
...
+  window.dispatchEvent(event);
+  if (!event.defaultPrevented) {
+    await showInstallPrompt(element);
+  }
+}
+
+function trackReadyState() {
```
marcoscaceres: Technically you are correct, but as this doesn't use a "await" keyword anywhere, it's fine to leave looking like a sync function.

implementation/index.html
```diff
@@ -0,0 +1,256 @@
...
+} {
+  const assertion = "prompt() is a function";
+  const e = new BeforeInstallPromptEvent("");
+  console.assert(e.prompt instanceof Function, assertion);
+} {
+  const assertion = "prompt() takes no arguments";
+  const e = new BeforeInstallPromptEvent("");
+  console.assert(e.prompt.length === 0, assertion);
+} {
+  const assertion = "calling prompt with unstrusted event results in NotAllowedError error";
```
marcoscaceres: fixed

implementation/index.html
```diff
@@ -0,0 +1,256 @@
...
+    e.prompt();
+    console.assert(false, assertion);
+  } catch (err) {
+    console.assert(err instanceof DOMException, assertion);
+    console.assert(err.type = "InvalidStateError", assertion);
+  }
+}
+
+// install events
+window.addEventListener("install", (ev) => {
```
marcoscaceres: Heh. Got used to the old way 👍  fixed!

implementation/BeforeInstallPromptEvent.js
```diff
@@ -7,9 +7,16 @@ const AppBannerPromptOutcome = new Set([
...
+  const internal = internalSlots.get(event);
+  internal.didPrompt = true;
+  const promptOutcome = await showInstallPrompt(element);
+  internal.userChoiceResolver(promptOutcome);
+}
+
 class BeforeInstallPromptEvent extends Event {
   constructor(typeArg, eventInit) {
-    // WebIDL Guard. Wont be in spec in spec as it's all handled by WebIDL.
+    // WebIDL Guard. Not in spec, as it's all handled by WebIDL.
```
kenchris: Not mentioned in the spec,

implementation/BeforeInstallPromptEvent.js
```diff
@@ -27,41 +34,34 @@ class BeforeInstallPromptEvent extends Event {
...
+      internal.userChoiceResolver = resolve;
     });
     internalSlots.set(this, internal);
   }
   prompt() {
-    if (this.isTrusted === false) {
-      const msg = "Untrusted events can't call prompt().";
-      throw new DOMException(msg, "NotAllowedError");
-    }
+    // if (this.isTrusted === false) {
```
kenchris: remove?

marcoscaceres:
Ok, this is more or less ready to go now. Anyone want to be a hero and give the spec text one last look over? Maybe @patrickkettner for a fresh perspective (you can ignore the "implementation/" files - though feedback on those tests would be welcome).

@kenchris I think I've addressed all your feedback. 

I'm away this week at a training course and part of next week at JS Conf AU in Melbourne - but will try to check and makes any updates in the evening.

mgiuca:
I will have another look over it shortly (definitely today).

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+        </h3>
+        <pre class="idl">
+          [Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+          interface BeforeInstallPromptEvent : Event {
+              readonly attribute Promise&lt;AppBannerPromptOutcome&gt; userChoice;
+              Promise&lt;void&gt; prompt();
+          };
+
+          dictionary BeforeInstallPromptEventInit : EventInit {
+            AppBannerPromptOutcome userChoice;
```
mgiuca: There was a discussion about this on the previous review round that went unfinished:

[My previous statement](https://github.com/w3c/manifest/pull/506/files/4af07193642bf9ae5d3f002337c64b196b0db908#r84207025)

I still don't see the point of this `userChoice` constructor argument being not a promise, implying that you've already made the choice and now you only have to understand why*. It _might_ be useful if `userChoice` was a promise, allowing the person fabricating this event to allow the choice to be resolved at some future point. But even then, I see little point in actually supporting the event fabrication use case at all. This would all (speccing, implementation, testing) be a lot easier if we banned fabrication altogether, or if the Web Platform doesn't allow that, just letting it exist but not do much (and not specifically inventing features to support this useless use case).

If you can justify why someone (a web developer, not a UA implementor) would want to create a `BeforeInstallPromptEvent` with a `userChoice` -- even a promise -- then I'm all ears.

\* Bonus points if you get that reference.

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+          <p>
+            To <dfn>request to present an install prompt</dfn> with
+            <a>BeforeInstallPromptEvent</a> <var>event</var>:
+          </p>
+          <ol>
+            <li>Set <a>[[\didPrompt]]</a> of <var>event</var> to
+            <code>true</code>.
+            </li>
+            <li>
+              <a>In parallel</a>:
```
mgiuca: How can this be done in parallel? Step 2.2 is dependent on the results of Step 2.1.

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+              </ol>
+            </li>
+          </ol>
+        </section>
+        <section>
+          <h4>
+            <code>userChoice</code> attribute
+          </h4>
+          <p>
+            The <dfn>userChoice</dfn> attribute, when getting, return the
```
mgiuca: s/return/returns?

(Maybe? I really don't understand the grammar of this awkward sentence, but as you mentioned previously it is spec language.)

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+          </h4>
+          <p>
+            This example shows how one might prevent an automated install
+            prompt from showing until the user has finished a set of tasks.
+            Those tasks are represented as an array of promises, which the
+            application "awaits" to finish before an install prompt is
+            presented to the end-user.
+          </p>
+          <pre class="example" title="'beforeinstallprompt' in action">
+          const tasksThatPreventInstallation = [];
```
mgiuca: Can this be `[...]` to imply it's got stuff in it, rather than just being an empty array?

Another way to express this would be to show a single task rather than an array. Perhaps, `waitForUserClickOnInstall`?

```js
event.preventDefault();
await waitForUserClickOnInstall;
event.prompt();
```

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+      </section>
       <section>
         <h3>
           Extensions to the <code>Window</code> object
         </h3>
         <p>
           The following extensions to the <code>Window</code> object specify
           the <a>event handler attributes</a> on which events relating to the
-          <a>installation</a> of a web application are fired.
+          <a>installation</a> of a web application are <a>fired</a>.
```
mgiuca: This could be taken out of this PR (as mentioned above).

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
         <pre class="idl">
           partial interface Window {
             attribute EventHandler onappinstalled;
+            attribute EventHandler onbeforeinstallprompt;
           };
         </pre>
         <div class="example">
           <p>
-            This example shows two ways of handling the "appinstalled" event.
+            This example shows two ways of handling the "<a>appinstalled</a>"
```
mgiuca: This could be taken out of this PR (as mentioned above).

index.html
```diff
@@ -439,18 +671,30 @@ <h3 id="installation-sec">
...
         </div>
-        <section>
+        <section data-dfn-for="Window">
           <h4>
             <code>onappinstalled</code> attribute
           </h4>
           <p>
-            <dfn for="Window">onappinstalled</dfn> is an <a>event handler IDL
-            attribute</a> for the "<a>appinstalled</a>" event type. The
+            The <dfn>onappinstalled</dfn> is an <a>event handler IDL
```
mgiuca: This could be taken out of this PR (as mentioned above).

index.html
```diff
@@ -2319,10 +2563,9 @@ <h3 id="applying">
           purposes</a>.
         </p>
         <p>
-          When an <a>image object</a> is used as an <a>icon</a>, a developer
-          can hint that the image is intended to serve some special
-          <dfn>purpose</dfn> in the context of the host OS (i.e., for better
-          integration).
+          When an <a>image object</a> is used as an icon, a developer can hint
```
mgiuca: This could be taken out of this PR (as mentioned above).

index.html
```diff
@@ -2893,16 +3136,39 @@ <h3 id="applying">
...
+          flag</dfn></a>
+        </li>
+        <li>
           <a href="https://dom.spec.whatwg.org/#event"><dfn><code>Event</code>
           interface</dfn></a>
         </li>
         <li>
-          <a href="https://dom.spec.whatwg.org/#concept-event-fire"><dfn>Fire
-          an event</dfn></a>
+          <a href=
```
mgiuca: This could be taken out of this PR (as mentioned above).

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+    const promptOutcome = await showInstallPrompt(element);
+    internal.userChoiceResolver(promptOutcome);
+  }
+
+  /**
+   * Implementation of BeforeInstallPromptEvent.
+   *
+   */
+  class BeforeInstallPromptEvent extends Event {
+
```
kenchris: unneeded newline

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+    const event = new BeforeInstallPromptEvent("beforeinstallprompt");
+    window.dispatchEvent(event);
+    if (!event.defaultPrevented) {
+      await requestInstallPrompt(event, element);
+    }
+  }
+
+  function trackReadyState() {
+    return new Promise((resolve) => {
```
kenchris: unneeded ()

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+      document.addEventListener("readystatechange", () => {
+        if (document.readyState === "complete") {
+          resolve();
+        }
+      });
+    });
+  }
+
+  async function showInstallPrompt(button) {
```
kenchris: the functon calling this called the argument element, why not just call it button there as well

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+      <button id="cancel">CANCEL</button>
+      <button id="add">ADD</button>
+    `;
+    const cancel = prompt.querySelector("#cancel");
+    const add = prompt.querySelector("#add");
+    document.body.appendChild(prompt);
+    const p = new Promise((resolve) => {
+      add.addEventListener("click", () => {
+        resolve("accepted");
+        //emulate installation to home screen
```
kenchris: // Emulate...

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+      installProcesses.splice(installProcesses.findIndex(item => item === p), 1);
+      if (button) {
+        button.disabled = false;
+      }
+    };
+    installProcesses.push(p);
+    return p;
+  }
+
+  if(!window.BeforeInstallPromptEvent) {
```
kenchris: I prefer space after if, as it is not really a function

implementation/index.html
```diff
@@ -0,0 +1,296 @@
...
+}
+
+#installprompt {
+  animation: fadein 2s;
+  width: 10cm;
+  padding: 1cm;
+}
+</style>
+<button onclick="notifyBeforeInstallPrompt(this)">
+  Simulate install BIP
```
kenchris: Why not write out BIP here

implementation/index.html
```diff
@@ -0,0 +1,296 @@
...
+</button>
+<button onclick="showInstallPrompt(this)">
+  Simulate manual "Add to Home Screen"
```
kenchris: Maybe add a bit of explanation to what that means

implementation/index.html
```diff
@@ -0,0 +1,296 @@
...
+</button>
+<script>
+/*globals BeforeInstallPromptEvent, DOMException*/
+"use strict";
+//
```
kenchris: why this empty comment?

implementation/index.html
```diff
@@ -0,0 +1,296 @@
...
+{
+  const assertion = "No arguments should throw";
```
kenchris: no argument much throw? why would an argument throw? :-)

Or do you mean some thing like "Calling with no arguments should throw"?

implementation/index.html
```diff
@@ -0,0 +1,296 @@
...
+  try {
+    new BeforeInstallPromptEvent();
+    console.assert(false, assertion);
+  } catch (err) {
+    console.assert(true, assertion);
+  }
+} {
+  const assertion = "First argument stringifies";
```
kenchris: is being stringified?

kenchris: I don't think the argument will stringify anything... but it itself may be stringified

index.html
```diff
@@ -224,15 +224,82 @@
...
+        the user agent presents to the user when, for instance, there are
+        sufficient <a>installability signals</a> to warrant <a>installation</a>
+        of the web application.
+      </p>
+      <p>
+        In either case, when a user agent <dfn data-lt=
+        "present an install prompt">presents an install prompt</dfn>, the
+        end-user's choice is represented as on of the following values (either
+        "accepted" or "dismissed"). These values are represented in the API of
+        this specification via <a>AppBannerPromptOutcome</a> enum.
```
kenchris: via *the*

index.html
```diff
@@ -224,15 +224,82 @@
...
         <h3>
+          Installing by prompting automatically
+        </h3>
+        <p>
+          The <dfn>steps to notify before an automated install prompt</dfn> are
+          given by the following algorithm:
+        </p>
+        <ol>
+          <li>Wait until the <code>Document</code> of the <a>top-level browsing
+          context</a> to be <a>completely loaded</a>.
```
kenchris: *is* completely loaded

index.html
```diff
@@ -224,15 +224,82 @@
...
+          </li>
+          <li>If there is already an <a>installation process</a> being
```
kenchris: prompt instead of process here? its confusing

index.html
```diff
@@ -406,25 +468,195 @@ <h3 id="installation-sec">
...
+          set when the event is <a data-lt=
+          "construct a BeforeInstallPromptEvent">constructed</a>:
+        </p>
+        <dl>
+          <dt>
+            <dfn>[[\didPrompt]]</dfn>
+          </dt>
+          <dd>
+            A boolean, initially <code>false</code>. Represents whether the
+            event has been used to <a>present an install prompt</a> (either by
```
kenchris: why is the last part within ()? I would do a comma instead

index.html
kenchris: in parallel means async in spec lang

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+        if (eventInit && "userChoice" in eventInit) {
+          return resolve(eventInit.userChoice);
+        }
+        internal.userChoiceResolver = resolve;
+      });
+      internalSlots.set(this, internal);
+    }
+
+    prompt() {
+      if (this.isTrusted === false) {
```
marcoscaceres: Well, it's as functional as any other user-constructed event that relies on trust. You are correct that it's crippled - but that's again by design: you can still construct these events and fire them, but it's just "prompt()" that doesn't do anything - but that will be exactly the same for the real ones when constructed by developers. In that sense, what is currently implemented is correct per-spec. 

Note that I could, theoretically, take BeforeInstallPromptEvent.js and plug it into Gecko, because we support JS-based IDL bindings. So, in such an implementation I would get a "trusted" event.

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+      requestInstallPrompt(this);
+    }
+
+    get userChoice() {
+      return internalSlots.get(this).userChoice;
+    }
+  }
+
+  async function notifyBeforeInstallPrompt(element) {
+    if (document.readyState === "complete") {
```
marcoscaceres: woops!

implementation/BeforeInstallPromptEvent.js
```diff
@@ -0,0 +1,154 @@
...
+      return;
+    }
+    const event = new BeforeInstallPromptEvent("beforeinstallprompt");
+    window.dispatchEvent(event);
+    if (!event.defaultPrevented) {
+      await requestInstallPrompt(event, element);
+    }
+  }
+
+  function trackReadyState() {
```
marcoscaceres: Used waitUntilReadyStateComplete

implementation/BeforeInstallPromptEvent.js
marcoscaceres: fixed

marcoscaceres:
Started addressing feedback... assertions in tests are now inline with clearer fail explanations now.

kenchris:
So maybe we should just merge this and then do the last fixed? I guess @marcoscaceres will have to merge all the commits to remove those unrelated changes and split them into a separate PR. Or maybe do that first and rebase this patch set?

marcoscaceres:
So, we are discussing maybe simplifying this whole things:
https://bugs.chromium.org/p/chromium/issues/detail?id=658638#c6

marcoscaceres:
closing, resending in pieces...

Issue #161, opened 2014-02-25
Integration with service workers


marcoscaceres:
Specify the relationship between this spec and [service workers](https://github.com/slightlyoff/ServiceWorker/). 

Or get service worker to define the member and hook in through the extension point in the spec...

kenchris:
Well that would mean that the service worker might not have been installed when the app has been added to the launcher and thus might not work offline until launched second time.

With it in the manifest, the UA could make sure that the service worker is installed (which in return could depend on fetching of resources to make offline work etc) when added to the home screen and fail adding it in case the service worker couldn't be installed.

I think that it makes sense

marcoscaceres:
> Well that would mean that the service worker might not have been installed when the app has been added to the launcher and thus might not work offline until launched second time.

True. But there is no guarantee that an app will ever be installed at all, and the app should work offline regardless (because 1. the UA might not support manifests, and 2. the user declined to install it, but still wants to access the app when offline - for example, the user is on vacation, they visited a restaurant's website, restarted their phone, and now want to view the website but offline). 

In other words, installing an application should never be a precondition for getting offline functionality.  

> With it in the manifest, the UA could make sure that the service worker is installed (which in return could depend on fetching of resources to make offline work etc) when added to the home screen and fail adding it in case the service worker couldn't be installed. 
>
> I think that it makes sense

I agree. It makes sense. But this is predicated on some assumptions that might not hold in practice: the service worker might not actually provide any offline capabilities. The service worker, and its cache, might have been evicted. Having a service worker in the manifest shouldn't grant that service worker any additional rights (or longevity) when compared to a SW declared in script. Otherwise, we will just end up with everyone declaring SWs in the manifest and then problems if the manifest doesn't get used by the UA. 

The more I think about this, the more I want to separate them: I feel like what the manifest provides and what SWs provide are separate concerns.

kenchris:
The offline etc won't depend on the manifest support, it just means that the first launch after added to the home screen will already have the service worker ready and if the app developer decides, work offline, have push messages etc. Without this, the user might first have it on second launch.

Is there another way do fix that issue? Another option would be to have some event to be called when an app is added to the home screen, which then would allow the developer to ensure the SW is installed, or fail "installing" the app.

marcoscaceres:
> The offline etc won't depend on the manifest support, it just means that the first launch after added to the home screen will already have the service worker ready and if the app developer decides, work offline, have push messages etc. Without this, the user might first have it on second launch.

[...]

PR #507, opened 2016-10-20
Add support for 'serviceworkers' member

kenchris:
First attempt at fixing #161

Where do we add the actual registration steps?

### DIFF
--- a/index.html
+++ b/index.html
@@ -3061,7 +3061,7 @@ Content-Security-Policy: img-src icons.example.com
               <li>Return <code>undefined</code>.
               </li>
             </ol>
-          <li>Return <var>type</var>.
+          <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
           </li>
         </ol>
       </section>
@@ -3099,7 +3099,7 @@ Content-Security-Policy: img-src icons.example.com
               <li>Otherwise, return False</li>
             </ol>
           </li>
-          <li>Return <var>use_cache</var>.
+          <li>Otherwise, return <var>value</var>.
           </li>
         </ol>
       </section>
--- a/index.html
+++ b/index.html
@@ -338,13 +338,25 @@
             <ol>
               <li>Instantiate an <a>installation process</a>.
               </li>
-              <li>Let <a>manifest</a> be the result of <a>obtaining the
+              <li>Let <a>manifest</a> and <a>manifest URL</a> be the result of <a>obtaining the
               manifest</a>.
               </li>
               <li>If <a>obtaining the manifest</a> results in an error, a user
               agent can, at this point, fall back to using the <a>top-level
               browsing context</a>' <code>Document</code>'s metadata to
               populate an <a>installation process</a>' UI.
+              </li>
+              <li>If <a>obtaining the manifest</a> succeeds, and the
+              <var>service worker registration</var> of <a>manifest</a> is not
+              <code>undefined</code>, a user agent can at this point
+              <ul>
+                <li>Invoke <a>Start Register</a> with <var>scope</var> and <var>src</var>
+                members of the <var>service worker registration</var>, a new <var>promise</var>,
+                null, <a>manifest URL</a>, <code>"classic"</code> and <code>False</code>,
+                </li>
+              </ul>
+              in which case the state of the settled <var>promise</var> determine whether the
+              <a>installation succeeded</a> or not.
               </li>
               <li>If the <a>installation succeeded</a>, <a>queue a task</a> on
               the <a>application life-cycle task source</a> to <a>fire an
@@ -1239,8 +1251,8 @@ Content-Type: application/manifest+json
           <code>start_url</code> member</a> with <var>manifest</var>,
           <var>manifest URL,</var> and <var>document URL</var> as arguments.
           </li>
-          <li>Let <var>service worker</var> of <var>parsed manifest</var> be
-          the result of running the <a>steps for processing the
+          <li>Let <var>service worker registration</var> of <var>parsed manifest</var>
+          be the result of running the <a>steps for processing the
           <code>serviceworker</code> member</a> with <var>manifest</var>,
           <var>manifest URL</var>, and <var>serviceworker</var> as arguments.
           </li>
@@ -3154,7 +3166,11 @@ Content-Security-Policy: img-src icons.example.com
       </p>
       <p>
         The <a href="https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy">
-        <dfn>Is origin potentially trustworthy</dfn></a> is defined in [[!SECURE-CONTEXTS]]
+        <dfn>Is origin potentially trustworthy</dfn></a> is defined in [[!SECURE-CONTEXTS]].
+      </p>
+      <p>
+        The <a href="https://w3c.github.io/ServiceWorker/#start-register"><dfn>Start Register</dfn></a>
+        is defined in [[!SERVICE-WORKERS-1]].
       </p>
       <p>
         The <a href=
--- a/index.html
+++ b/index.html
@@ -349,13 +349,17 @@
               <li>If <a>obtaining the manifest</a> succeeds, and the result of running the
               <a>steps for processing the serviceworker member</a> with <a>manifest</a>
               returns a valid <var>registration</var>, a user agent can at this point
-              <ul>
+              <ol>
+                <li>Let <var>client</var> be the <a>top-level browsing context</a>'
+                <code>Document</code>'s
+                <a href="https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object">
+                relevant settings object</a>, or <code>null</code> if unavailable.
                 <li>Invoke <a>Start Register</a> with <var>scope</var> and <var>src</var>
                 members of the <var>registration</var>, a new <var>promise</var>,
-                null, <a>manifest URL</a>, plus the <var>type</var> and <var>use_cache</var>
+                <var>client</var>, <a>manifest URL</a>, plus the <var>type</var> and <var>use_cache</var>
                 members of the <var>registration</var>,
                 </li>
-              </ul>
+              </ol>
               in which case the state of the settled <var>promise</var> determine whether the
               <a>installation succeeded</a> or not.
               </li>
--- a/index.html
+++ b/index.html
@@ -2048,8 +2048,9 @@ Content-Type: application/manifest+json
           service worker as defined in [[!SERVICE-WORKERS-1]].
         </p>
         <p>
-          The <a><code>serviceworker</code> member</a> represents a service worker
-          registration. Other service worker registrations can be done, for instance
+          The <a><code>serviceworker</code> member</a> represents a <a
+          href="https://w3c.github.io/ServiceWorker/#service-worker-registration-concept">
+          service worker registration</a>. Other service worker registrations can be done, for instance
           by a script; if these have different scopes they will be considered separate
           registrations. If they have the same scope and script URL, they coalesce.
           If they have different script URLs, last one wins.
@@ -2057,42 +2058,44 @@ Content-Type: application/manifest+json
         <p>
           The <dfn>steps for processing the <code>serviceworker</code>
           member</dfn> are given by the following algorithm. The algorithm
-          takes a <a>manifest</a> <var>manifest</var> and <a>serviceworker object</a>
-          <var>serviceworker</var> as an argument. This algorithm returns a
-          serviceworker object <var>serviceworker</var>, which can be <code>undefined</code>.
+          takes a <a>manifest</a> <var>manifest</var>. This algorithm returns a
+          registration object <var>registration</var>, which can be <code>undefined</code>.
         </p>
         <ol>
+          <li>Let <var>unprocessed registration</var> be the result of calling the
+          <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
+          with argument "<code>serviceworker</code>".
           <li>Let <var>src</var> be the result of running the <a>steps
           for processing the <code>src</code> member of a service worker</a>
-          with <var>serviceworker</var> and <var>manifest URL</var>.
+          with <var>unprocessed registration</var> and <var>manifest URL</var>.
           </li>
           <li>If <var>src</var> is <code>undefined</code>, or if the result of
-          running <a href="https://w3c.github.io/webappsec/specs/powerfulfeatures/#is-origin-trustworthy">Is origin potentially trustworthy</a>
+          running <a>Is origin potentially trustworthy</a>
           with the origin of <var>src</var> is <code>Not Trusted</code>, return
           <code>undefined</code>.
           </li>
-          <li>Otherwise, let <var>serviceworker</var> be an object with
+          <li>Otherwise, let <var>registration</var> be an object with
           properties <code>src</code> and <code>scope</code>. All properties
           initially set to <code>undefined</code>.
           </li>
-          <li>Set <var>serviceworker</var>'s <code>src</code> property to be
+          <li>Set <var>registration</var>'s <code>src</code> property to be
           <var>src</var>.
           </li>
           <li>Let <var>type</var> be the result of running the <a>steps
           for processing the <code>scope</code> member of a service worker</a>
-          passing <var>serviceworker</var>.
+          passing <var>unprocessed registration</var>.
           </li>
           <li>If <var>scope</var> is not <code>undefined</code>,
-          set<var>serviceworker</var>'s <code>scope</code>
+          set <var>registration</var>'s <code>scope</code>
           property to be <var>scope</var>.
           </li>
-          <li>Return <var>serviceworker</var>.
+          <li>Return <var>registration</var>.
           </li>
         </ol>
         <div class="example">
           In the following example, the web application is listing
           a service worker for the <code>/foo</code> scope:
-<pre class="">
+<pre class="example">
 "serviceworker": {
   "src": "sw.js",
   "scope": "/foo"
@@ -2904,14 +2907,14 @@ Content-Security-Policy: img-src icons.example.com
         <p>
           The <dfn>steps for processing the <code>src</code> member of a
           service worker</dfn> are given by the following algorithm. The algorithm takes
-          a <a>serviceworker object</a> <var>serviceworker</var>, and a <a>URL</a> <var>manifest
+          a <a>serviceworker object</a> <var>registration</var>, and a <a>URL</a> <var>manifest
           URL</var>, which is the <a>URL</a> from which the
           <var>manifest</var> was fetched. This algorithm will return a
           <a>URL</a> or <code>undefined</code>.
         </p>
         <ol>
           <li>Let <var>value</var> be the result of calling the
-          <a>[[\GetOwnProperty]]</a> internal method of <var>serviceworker</var>
+          <a>[[\GetOwnProperty]]</a> internal method of <var>registration</var>
           passing " <code>src</code>" as the argument.
           </li>
           <li>Let <var>type</var> be <a>Type</a>(<var>value</var>).
@@ -2945,14 +2948,14 @@ Content-Security-Policy: img-src icons.example.com
         <p>
           The <dfn>steps for processing the <code>scope</code> member of a
           service worker</dfn> are given by the following algorithm. The algorithm takes
-          a <a>serviceworker object</a> <var>serviceworker</var>, and a <a>URL</a> <var>manifest
+          a <a>serviceworker object</a> <var>registration</var>, and a <a>URL</a> <var>manifest
           URL</var>, which is the <a>URL</a> from which the
           <var>manifest</var> was fetched. This algorithm will return a
           <a>URL</a> or <code>undefined</code>.
         </p>
         <ol>
           <li>Let <var>value</var> be the result of calling the
-          <a>[[\GetOwnProperty]]</a> internal method of <var>serviceworker</var>
+          <a>[[\GetOwnProperty]]</a> internal method of <var>registration</var>
           passing " <code>scope</code>" as the argument.
           </li>
           <li>Let <var>type</var> be <a>Type</a>(<var>value</var>).
@@ -3148,6 +3151,10 @@ Content-Security-Policy: img-src icons.example.com
         "https://w3c.github.io/webappsec/specs/content-security-policy/#default_src">
         <dfn><code>default-src</code></dfn></a> directives are defined in
         [[!CSP3]].
+      </p>
+      <p>
+        The <a href="https://w3c.github.io/webappsec-secure-contexts/#is-origin-trustworthy">
+        <dfn>Is origin potentially trustworthy</dfn></a> is defined in [[!SECURE-CONTEXTS]]
       </p>
       <p>
         The <a href=
--- a/index.html
+++ b/index.html
@@ -604,8 +604,8 @@ window.addEventListener("appinstalled", handleInstalled);
       </p>
       <p>
         A developer specifies the navigation scope via the
-        <a><code>scope</code> member</a>. In the case where the
-        <a><code>scope</code> member</a> is missing or in error, the navigation
+        <a data-lt="member-scope"><code>scope</code> member</a>. In the case where the
+        <a data-lt="member-scope"><code>scope</code> member</a> is missing or in error, the navigation
         scope is treated as <dfn>unbounded</dfn> (represented as the value
         <code>undefined</code>). In such a case, the manifest is applied to all
         URLs the application context is <a>navigated</a> to (see related
@@ -1238,6 +1238,11 @@ Content-Type: application/manifest+json
           result of running the <a>steps for processing the
           <code>start_url</code> member</a> with <var>manifest</var>,
           <var>manifest URL,</var> and <var>document URL</var> as arguments.
+          </li>
+          <li>Let <var>service worker</var> of <var>parsed manifest</var> be
+          the result of running the <a>steps for processing the
+          <code>serviceworker</code> member</a> with <var>manifest</var>,
+          <var>manifest URL</var>, and <var>serviceworker</var> as arguments.
           </li>
           <li>Let <var>display mode</var> of <var>parsed manifest</var> be the
           result of running the <a>steps for processing the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1937,6 +1942,39 @@ <h3 id="applying">
...
+          registrations. If they have the same scope and script URL, they coalesce.
+          If they have different script URLs, last one wins.
+        </p>
+        <p class="note">
+          The <code>serviceworkers</code> member is processed using the <a>steps for
+          processing an array of service workers</a>.
+        </p>
+        <div class="example">
+<pre class="">
+"serviceworkers": [{
```
kenchris: Your own example used plural and an array :-)

index.html
```diff
@@ -1937,6 +1942,39 @@ <h3 id="applying">
...
+  "src": "sw.js",
+  "scope": "/foo",
+  "options": {}
+}
```
RobDolinMS: @kenchris - Should like 1971 be `}]` to match line 1967?

benfredwells:
What is the purpose of adding this information? Is it intended that having a SW defined in the manifest would cause the SW to be registered, similar to calling `navigator.serviceWorker.register` in script?

If so it would be good to get this reviewed by some people familiar with SW registration.

kenchris:
@benfredwells it is just like calling register manually or adding a service worker link tag. Jake already approved this approach.

kenchris:
Can we make a decision on whether we want just one or multiple service workers in the manifest? @marcoscaceres @jakearchibald 

I don't care strongly, but I would like to know so that I can change the WIP patch

marcoscaceres:
Only 1 please.

kenchris:
@jakearchibald what is your take?

RobDolinMS:
@kenchris  - After discussion with the Edge PM for Service Workers, Ali Alabas (@AliAMS) and others at MSFT/Edge, we believe a single serviceworker object listed in the web app manifest will address our use cases; but don’t object to having an array of serviceworkers.  /cc @marcoscaceres @jakearchibald

benfredwells:
@kenchris thanks for the explanation.

SW registration, updating (e.g. if you change the value in the manifest) and removal is complicated and has already had lots of thought put into it. Are you planning on including algorithms for declarative SW registrations, updates and removals in this spec? Maybe it would be better to just have the syntax for the declarations here, and get the SW spec updated to cover the notion of declarative registration in general, and point to that.

RobDolinMS:
As far as the question about one ServiceWorker{ } or multiple ServiceWorkers[ ], what if we went with one to start and then looked for feedback / push-back from potential implementers?  

/cc @marcoscaceres @kenchris @benfredwells @jakearchibald

kenchris:
I talked to Jake, we will discuss it after the Chrome Dev Summit as he is currently quite busy. I will work on this after than then.

marcoscaceres:
@kenchris, thanks for the update. Looking forward to @jakearchibald's input here. My position remains we should only have one (does enough work to make sure the application can be launched).

RobDolinMS:
@kenchris Have you been able to sync with @jakearchibald about one ServiceWorker{ } or multiple ServiceWorkers[ ] ?  

Summarizing positions:

- @kenchris - prefers ServiceWorkers[ ] but ["I don't care strongly"](https://github.com/w3c/manifest/pull/507#issuecomment-256024401)
- @marcoscaceres - prefers ServiceWorker{ } - ["only 1 please"](https://github.com/w3c/manifest/pull/507#issuecomment-256029771)
- @AliAMS and @RobDolinMS - [OK with either](https://github.com/w3c/manifest/pull/507#issuecomment-256156776)
- @jakearchibald - TBD

kenchris:
No, didn't get to talk unfortunately.

RobDolinMS:
(I've sent an email ping to @jakearchibald with @marcoscaceres and @kenchris on CC)

jakearchibald:
I'm happy to go with one for simplicity here. Additional SWs can be in the document as @marcoscaceres says, or perhaps even `Link` headers on the manifest response.

Apologies for the slow reply!

RobDolin:
@marcoscaceres or @kenchris - Since @jakearchibald has given a thumbs-up to one serviceworker{ } would one of you want to modify the PR and merge?

RobDolinMS:
@marcoscaceres or @kenchris - Since @jakearchibald has given a thumbs-up to one serviceworker{ } would one of you want to modify the PR?

kenchris:
Yes, I can work on this

kenchris:
Now I have IPR issues as well, though I am in the Web Platforms WG as indicated by the Repository Manager

chaals:
@siusin ping… re @kenchris access…

siusin:
@kenchris -
Intel hasn't re-joined the group yet after we rechartered, so from an IPR perspective, the PRs would be flagged as non valid until the contributors sign up for the group again.

@chaals -
Of course, chairs can make a decision and get the PR merged in this case.

chaals:
@siusin @kenchris I am happy that unless Intel fails to rejoin by December 15, they are currently treated as a member and have the effective commitment, so this should be marked as OK. But it seems something went funny with my access to ash-nazg to make it happen. Please go ahead and merge

siusin:

siusin:

RobDolinMS:
>Only 1 please. 
'Looks like @kenchris has fixed the PR so there is only one serviceworker{ } and @siusin has gotten the IRP thumbs-up.

@marcoscaceres - Would you want to merge?

marcoscaceres:
@RobDolinMS not yet :) This is far from done.

kenchris:
Looking forward to comments from @marcoscaceres  :-)

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          <code>serviceworker</code> member
+        </h3>
+        <p>
+          The <dfn><code>serviceworker</code> member</dfn> describes a
+          service worker as defined in [[!SERVICE-WORKERS-1]].
+        </p>
+        <p>
+          The <a><code>serviceworker</code> member</a> represents a service worker
+          registration. Other service worker registrations can be done, for instance
+          by a script; if these have different scopes they will be considered separate
```
marcoscaceres: We should probably not re-specify the rules that are in the service worker spec; we should reference them.

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+        <h3>
+          <code>serviceworker</code> member
+        </h3>
+        <p>
+          The <dfn><code>serviceworker</code> member</dfn> describes a
+          service worker as defined in [[!SERVICE-WORKERS-1]].
+        </p>
+        <p>
+          The <a><code>serviceworker</code> member</a> represents a service worker
+          registration. Other service worker registrations can be done, for instance
```
marcoscaceres: Link to registration definition in SW

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          <var>serviceworker</var> as an argument. This algorithm returns a
+          serviceworker object <var>serviceworker</var>, which can be <code>undefined</code>.
+        </p>
+        <ol>
+          <li>Let <var>src</var> be the result of running the <a>steps
+          for processing the <code>src</code> member of a service worker</a>
+          with <var>serviceworker</var> and <var>manifest URL</var>.
+          </li>
+          <li>If <var>src</var> is <code>undefined</code>, or if the result of
+          running <a href="https://w3c.github.io/webappsec/specs/powerfulfeatures/#is-origin-trustworthy">Is origin potentially trustworthy</a>
```
marcoscaceres: Nit: throw in this URL in the dependencies section at the bottom of the document.

marcoscaceres: Won't the service worker registration attempt do this for us?

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          The <dfn>steps for processing the <code>serviceworker</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <a>manifest</a> <var>manifest</var> and <a>serviceworker object</a>
+          <var>serviceworker</var> as an argument. This algorithm returns a
+          serviceworker object <var>serviceworker</var>, which can be <code>undefined</code>.
+        </p>
+        <ol>
+          <li>Let <var>src</var> be the result of running the <a>steps
+          for processing the <code>src</code> member of a service worker</a>
+          with <var>serviceworker</var> and <var>manifest URL</var>.
```
marcoscaceres: Below, you are using `let service worker be` thus trashing this one? I don't think we need to send in a service worker here. Just process the members first, then "attempt to register the service worker"

marcoscaceres: Sorry, this comment is in the wrong place!

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          The <a><code>serviceworker</code> member</a> represents a service worker
+          registration. Other service worker registrations can be done, for instance
+          by a script; if these have different scopes they will be considered separate
+          registrations. If they have the same scope and script URL, they coalesce.
+          If they have different script URLs, last one wins.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>serviceworker</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <a>manifest</a> <var>manifest</var> and <a>serviceworker object</a>
```
marcoscaceres: Below, you are using `let service worker be...` thus trashing this one? I don't think we need to send in a service worker here. Just process the members first, then "attempt to register the service worker"

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+        </p>
+        <ol>
+          <li>Let <var>src</var> be the result of running the <a>steps
+          for processing the <code>src</code> member of a service worker</a>
+          with <var>serviceworker</var> and <var>manifest URL</var>.
+          </li>
+          <li>If <var>src</var> is <code>undefined</code>, or if the result of
+          running <a href="https://w3c.github.io/webappsec/specs/powerfulfeatures/#is-origin-trustworthy">Is origin potentially trustworthy</a>
+          with the origin of <var>src</var> is <code>Not Trusted</code>, return
+          <code>undefined</code>.
```
marcoscaceres: screwing up the security aspects of the registration is a pretty big error... so we should maybe pass this through and just allow service worker registration to fail (and kill the whole object then)

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          </li>
+          <li>Otherwise, let <var>serviceworker</var> be an object with
```
marcoscaceres: nit: ` let <var>serviceworker</var> ` might confuse people into thinking this the actual service worker... maybe "Let `registration properties` be..."

index.html
```diff
@@ -2033,6 +2038,71 @@ <h3 id="applying">
...
+          set<var>serviceworker</var>'s <code>scope</code>
+          property to be <var>scope</var>.
+          </li>
+          <li>Return <var>serviceworker</var>.
+          </li>
+        </ol>
+        <div class="example">
+          In the following example, the web application is listing
+          a service worker for the <code>/foo</code> scope:
+<pre class="">
```
marcoscaceres: nit: class=example?

marcoscaceres:
So, this is a good start. We obviously still need to add when registration occurs, what happens when registration fails, how to deal when there are clients and when there are no clients (i.e., the sync update case).

index.html
kenchris: done

index.html
kenchris: Ok, fixed

index.html
kenchris: :)

index.html
kenchris: done, will need to check

index.html
kenchris: ok, did something similar

index.html
kenchris: ok, needs to be applied elsewhere in the spec

kenchris:
Yes, any help on that is appreciated :-) I will see if I can get junkee to have a look and maybe help me.

jungkees:
@kenchris, [Start Register](https://w3c.github.io/ServiceWorker/#start-register) is a gate function that you can invoke to trigger a [Register](https://w3c.github.io/ServiceWorker/#start-register). Invoking Start Register makes a job and schedules it in the job queue defined in SW. Start Register is also supposed to take care of the security check and parsing the URLs for the given src and scope, etc. Basically, a register request is processed async given a promise (to Start Register).

I'm not quite sure, but you might want to invoke Start Register from [steps for processing the serviceworker member](https://rawgit.com/w3c/manifest/4a5f762812a15c81f7c2b7f4f37831410031f4da/index.html#dfn-steps-for-processing-the-serviceworker-member) I guess. Start Register returns nothing but the given promise to it resolves with a [service worker registration](https://w3c.github.io/ServiceWorker/#dfn-service-worker-registration) on success and rejects on failure. (See the example call site: In https://w3c.github.io/ServiceWorker/#link-element-processing, step 12 of the algorithm starting with "When a serviceworker link’s link element ..".)

This is basically what SW expects for a register request. I think we should look together if any changes are needed while you're working on this PR.

index.html
```diff
@@ -338,14 +338,26 @@
...
               browsing context</a>' <code>Document</code>'s metadata to
               populate an <a>installation process</a>' UI.
               </li>
+              <li>If <a>obtaining the manifest</a> succeeds, and the
+              <var>service worker registration</var> of <a>manifest</a> is not
+              <code>undefined</code>, a user agent can at this point
+              <ul>
+                <li>Invoke <a>Start Register</a> with <var>scope</var> and <var>src</var>
+                members of the <var>service worker registration</var>, a new <var>promise</var>,
+                null, <a>manifest URL</a>, <code>"classic"</code> and <code>False</code>,
```
jungkees: SW's RegistrationOptions dictionary has been extended: https://w3c.github.io/ServiceWorker/#dictdef-registrationoptions. I think both `"classic"` and `"module"` worker types should be covered and also `useCache` should be supported to be aligned with other registration paths. (`useCache` indicates whether the UA should set the cache mode to "no-cache" or "default" when fetching a service worker script from networks.)

jungkees: For the _client_ param to Start Register, I think the browsing context's Document's relevant settings object should be passed instead of null. The given client is used to queue a task to later to manipulate promise settlement and parse a fetched script resources later.

kenchris: OK, will look at that tomorrow then

index.html
```diff
@@ -2083,34 +2084,59 @@ <h3 id="applying">
...
+          <li>Let <var>type</var> be the result of running the <a>steps
+          for processing the <code>type</code> member of a service worker</a>
+          passing <var>unprocessed registration</var>.
+          </li>
+          <li>If <var>type</var> is not <code>undefined</code>
+          set <var>registration</var>'s <code>type</code>
+          property to be <var>type</var>.
+          </li>
+          <li>If <var>type</var> is <code>undefined</code> abort these steps and
+          return <code>undefined</code>.
```
jungkees: Service workers expect a default value for type member. If type is undefined here, please set the default value to "classic" when passing it to Start Register.

index.html
```diff
@@ -2083,34 +2084,59 @@ <h3 id="applying">
...
+          <li>Let <var>use cache</var> be the result of running the <a>steps
+          for processing the <code>use_cache</code> member of a service worker</a>
+          passing <var>unprocessed registration</var>.
+          </li>
+          <li>If <var>use cache</var> is not <code>undefined</code>,
+          set <var>registration</var>'s <code>use_cache</code>
+          property to be <var>use cache</var>.
+          </li>
+          <li>If <var>use cache</var> is <code>undefined</code> abort these steps
+          and return <code>undefined</code>.
```
jungkees: Likewise, If use_cache is undefined here, please set the default value to false when passing it to Start Register.

* Note that there's an ongoing discussion about the expected default behavior in https://github.com/w3c/ServiceWorker/issues/893. So, there's a chance SW may require true as default depending on the discussion there.

index.html
```diff
@@ -2083,34 +2084,59 @@ <h3 id="applying">
...
-          initially set to <code>undefined</code>.
+          properties <code>src</code>, <code>scope</code>, <code>type</code> and
+          <code>use_cache</code>. All properties initially set to
+          <code>undefined</code>.
           </li>
           <li>Set <var>registration</var>'s <code>src</code> property to be
           <var>src</var>.
           </li>
-          <li>Let <var>type</var> be the result of running the <a>steps
+          <li>Let <var>scope</var> be the result of running the <a>steps
```
jungkees: Similar to type and use_cache, SW expects a default value for scope too. If scope is undefined, please give null to Start Register as a default value. That makes Start Register set the scope to the same location as the given script url.

kenchris: When the value is undefined here, it means that there was a parsing error (check sub algorithms), and I thus don't attempt to register at all.

index.html
kenchris: That is set in the sub algorithm. Undefined here means that the sub algorithm didn't get a valid value

index.html
kenchris: same, ok Please keep me up to date :)

index.html
```diff
@@ -2033,6 +2051,99 @@ <h3 id="applying">
...
+          <li>Otherwise, let <var>registration</var> be an object with
+          properties <code>src</code>, <code>scope</code>, <code>type</code> and
+          <code>use_cache</code>. All properties initially set to
+          <code>undefined</code>.
+          </li>
+          <li>Set <var>registration</var>'s <code>src</code> property to be
+          <var>src</var>.
+          </li>
+          <li>Let <var>scope</var> be the result of running the <a>steps
+          for processing the <code>scope</code> member of a service worker</a>
```
jungkees: When this algorithm returns _undefined_, the _steps to install the web application_ gets to pass nothing to Start Register. In this case, shouldn't there be an error handling such that Start Register is not invoked?

index.html
jungkees: I understood the intention. Please check if it actually doesn't attempt to register in that case (i.e. when the sub algorithm returns _undefined_.) I left a comment here: https://github.com/w3c/manifest/pull/507#discussion_r92408719.

index.html
jungkees: Same as above. I saw you return false when no value is given. It looks good.

index.html
```diff
@@ -2989,6 +3015,88 @@ <h3 id="applying">
...
+            </ol>
+          </li>
+          <li>If <a>Trim</a>(value) is the empty string, then:
+            <ol>
+              <li>Issue a developer warning that the type is unsupported.
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          <li>Return <var>type</var>.
```
jungkees: Shouldn't the return value be _value_.[[Value]] instead?

index.html
```diff
@@ -2989,6 +3015,88 @@ <h3 id="applying">
...
+          <li>If <var>type</var> is not "boolean", then:
+            <ol>
+              <li>If <var>type</var> is not "<code>undefined</code>", issue a
+              developer warning that the type is unsupported and
+              return <code>undefined</code>.
+              </li>
+              <li>Otherwise, return False</li>
+            </ol>
+          </li>
+          <li>Return <var>use_cache</var>.
```
jungkees: Shouldn't the return value be _value_.[[Value]] instead?

index.html
kenchris: Let me check that, it should not be invoked no

index.html
kenchris: We have <li>Let <var>type</var> be <a>Type</a>(<var>value</var>). above. That is how @marcoscaceres did it elsewhere. I think that should be fine. Why the value.[[Value]]? looks like a private slot

index.html
kenchris: <li>If <a>obtaining the manifest</a> succeeds, and the
              <var>service worker registration</var> of <a>manifest</a> is not
              <code>undefined</code>, a user agent can at this point...

It is not invoked if it fails

index.html
jungkees: Ah.. I got it. It has to do with the prefix ! for an ECMAScript abstract operation. So, the _value_ here is the actual value for this member already. But then it still needs to return _value_ instead of _type_. Referring to other cases, it should return Trim(value) here I suppose.

jungkees:
@kenchris, when the above comments are addressed, I think the shape of the hooks between Web App Manifest and Service Workers would be OK. For the details of the spec language here, I defer to you and @marcoscaceres. I'll continue to help, of course.

index.html
kenchris: right, fixed

index.html
kenchris: correct

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+          The <dfn data-lt="serviceworker-type"><code>type</code> member</dfn> of a
+          <a>serviceworker object</a> is the service worker's
+          <a href="https://html.spec.whatwg.org/multipage/workers.html#workertype">worker type</a>.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>type</code> member of a
+          service worker</dfn> are given by the following algorithm. The algorithm takes
+          a <a>serviceworker object</a> <var>registration</var>, and a <a>URL</a> <var>manifest
+          URL</var>, which is the <a>URL</a> from which the
+          <var>manifest</var> was fetched. This algorithm will return a
```
marcoscaceres: Hmm... what if the manifest redirected during fetch... we might need to clarify that somewhere.

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+              developer warning that the type is unsupported and
+              return <code>undefined</code>.
+              </li>
+              <li>Return <code>"classic"</code>.
+              </li>
+            </ol>
+          </li>
+          <li>If <a>Trim</a>(value) is the empty string, then:
+            <ol>
+              <li>Issue a developer warning that the type is unsupported.
```
marcoscaceres: This seems odd... you are returning undefined here, but "classic" above?

marcoscaceres: I.e., should we recover to "classic" in error cases for backwards compat?

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
```
marcoscaceres: Also, should should this be restricted to just "classic" and "module"?

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+          </li>
+          <li>Let <var>type</var> be <a>Type</a>(<var>value</var>).
+          </li>
+          <li>If <var>type</var> is not "boolean", then:
+            <ol>
+              <li>If <var>type</var> is not "<code>undefined</code>", issue a
+              developer warning that the type is unsupported and
+              return <code>undefined</code>.
+              </li>
+              <li>Otherwise, return False</li>
```
marcoscaceres: nit: return `<code>false</code>`

index.html
```diff
@@ -338,14 +338,31 @@
...
               </li>
+              <li>If <a>obtaining the manifest</a> succeeds, and the result of running the
+              <a>steps for processing the serviceworker member</a> with <a>manifest</a>
+              returns a valid <var>registration</var>, a user agent can at this point
+              <ol>
+                <li>Let <var>client</var> be the <a>top-level browsing context</a>'
+                <code>Document</code>'s
+                <a href="https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object">
+                relevant settings object</a>, or <code>null</code> if unavailable.
+                <li>Invoke <a>Start Register</a> with <var>scope</var> and <var>src</var>
```
kenchris: I am not sure how to best solve this. Maybe you could solve in a follow up?

index.html
```diff
@@ -338,14 +338,31 @@
...
+                members of the <var>registration</var>, a new <var>promise</var>,
+                <var>client</var>, <a>manifest URL</a>, plus the <var>type</var> and <var>use_cache</var>
+                members of the <var>registration</var>,
+                </li>
+              </ol>
+              in which case the state of the settled <var>promise</var> determine whether the
```
kenchris: fixed

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
         </section>
       </section>
+
+      <section>
+        <h3>
+          <code>serviceworker</code> member
+        </h3>
+        <p>
+          The <dfn><code>serviceworker</code> member</dfn> describes a
+          service worker as defined in [[!SERVICE-WORKERS-1]].
```
kenchris: Not as far as I could see. @jungkees ?

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
+        </p>
+        <p>
+          The <a><code>serviceworker</code> member</a> represents a <a
+          href="https://w3c.github.io/ServiceWorker/#service-worker-registration-concept">
+          service worker registration</a>. Other service worker registrations can be done, for instance
```
kenchris: Made it a note now. @jungkees can you point me to the best place to refer to?

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
+          by a script; if these have different scopes they will be considered separate
+          registrations. If they have the same scope and script URL, they coalesce.
+          If they have different script URLs, last one wins.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>serviceworker</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <a>manifest</a> <var>manifest</var>. This algorithm returns a
+          registration object <var>registration</var>, which can be <code>undefined</code>.
```
kenchris: ok

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
+        <ol>
+          <li>Let <var>unprocessed registration</var> be the result of calling the
+          <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
+          with argument "<code>serviceworker</code>".
+          <li>Let <var>src</var> be the result of running the <a>steps
+          for processing the <code>src</code> member of a service worker</a>
+          with <var>unprocessed registration</var> and <var>manifest URL</var>.
+          </li>
+          <li>If <var>src</var> is <code>undefined</code>, or if the result of
+          running <a>Is origin potentially trustworthy</a>
```
kenchris: ok

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
+          </li>
+          <li>Otherwise, set <var>registration</var>'s <code>scope</code>
+          property to be <var>scope</var>.
+          </li>
+          <li>Let <var>type</var> be the result of running the <a>steps
+          for processing the <code>type</code> member of a service worker</a>
+          passing <var>unprocessed registration</var>.
+          </li>
+          <li>If <var>type</var> is <code>undefined</code> abort these steps and
+          return <code>undefined</code>.
```
kenchris: :)

index.html
```diff
@@ -2033,6 +2055,99 @@ <h3 id="applying">
...
+          property to be <var>use cache</var>.
+          </li>
+          <li>Return <var>registration</var>.
+          </li>
+        </ol>
+        <div class="example">
+          In the following example, the web application is listing
+          a service worker for the <code>/foo</code> scope, bypassing the user agent cache
+          when fetching the <code>"sw.js"</code> source:
+<pre class="example">
```
kenchris: ok

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+        registration for the web application.
+      </p>
+      <section>
+        <h3>
+          <code>src</code> member
+        </h3>
+        <p>
+          The <dfn data-lt="serviceworker-src"><code>src</code> member</dfn> of a
+          <a>serviceworker object</a> is a <a>URL</a> representing a
+          service worker.
```
kenchris: @jungkees ^?

index.html
```diff
@@ -2816,6 +2931,181 @@ <h3 id="applying">
...
+        </ol>
+      </section>
+      <section>
+        <h3>
+          <code>scope</code> member
+        </h3>
+        <p>
+          The <dfn data-lt="serviceworker-scope"><code>scope</code> member</dfn> of a
+          <a>serviceworker object</a> is the service worker's associated
+          <a href="https://w3c.github.io/ServiceWorker/#dfn-scope-url">scope URL</a>.
```
kenchris: done

index.html
kenchris: That is how it works for Link element. What do you think @jungkees ?

index.html
kenchris: Can you help with that? :-)

index.html
kenchris: Yes, "classic", "module" or undefined. Everything else should fail the whole registration as it does with link element

index.html
kenchris: k

kenchris:
So I can go ahead and merge?

marcoscaceres:
Yeah, why not :) ship it

GrosSacASac:
If I understand correctly, the service worker declaration inside the manifest has no consequence unless the web app is installed without directly visting the web app url.(with an web app store ?)

I ask because I have an app where the service worker is only installed after the users checks the checkbox USE OFFLINE. I don't want the manifest to install the service worker on normal page load without user consent first.

I came here because https://www.w3.org/TR/appmanifest/#serviceworker-member words are confusing.

marcoscaceres:
@GrosSacASac, moved your issue here: https://github.com/w3c/manifest/issues/640 so we don't lose track of it.

PR #517, opened 2016-11-07
feat(window): add window.onbeforeinstallprompt

marcoscaceres:
Part 1, will send bits that can go in easily... then link everything together. 

Can folks please review these as they come in.  Will try to keep them small and self contained.


marcoscaceres:
This hasn't changed since ##506, so merging.

PR #518, opened 2016-11-07
feat: define application life-cycle task source

marcoscaceres:
This is part 2 of BIP... application life-cycle task source is used to queue up the events.


marcoscaceres:
This was defined in #506 and hasn't changed since, so merging.

PR #519, opened 2016-11-07
feat: add AppBannerPromptOutcome enum

marcoscaceres:
Part 3, adds AppBannerPromptOutcome enum, which represents the outcome from showing an install prompt.

### DIFF
--- a/index.html
+++ b/index.html
@@ -231,6 +231,41 @@
         instance, there are sufficient <a>installability signals</a> to warrant
         <a>installation</a> of the web application.
       </p>
+      <section data-dfn-for="AppBannerPromptOutcome">
+        <h4>
+          <code>AppBannerPromptOutcome</code> enum
+        </h4>
+        <pre class="idl">
+          enum AppBannerPromptOutcome {
+            "accepted",
+            "dismissed"
+          };
+        </pre>
+        <p>
+          The <dfn>AppBannerPromptOutcome</dfn> enum's values represent the
+          outcomes from <a data-lt="presents an install prompt">presenting the
+          end-user with an install prompt</a>.
+        </p>
+        <dl data-dfn-for="AppBannerPromptOutcome">
+          <dt>
+            <dfn>accepted</dfn>:
+          </dt>
+          <dd>
+            The end-user indicated that they would like the user agent to
+            <a>install</a> the web application.
+            <p class="note">
+              This does not necessarily mean that the <a>installation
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -231,6 +231,43 @@
...
+          The <dfn>AppBannerPromptOutcome</dfn> enum's values represent the
+          outcomes from <a data-lt="presents an install prompt">presenting the
+          end-user with an install prompt</a>.
+        </p>
+        <dl data-dfn-for="AppBannerPromptOutcome">
+          <dt>
+            <dfn>accepted</dfn>:
+          </dt>
+          <dd>
+            The end-user indicated that the user agent MUST attempt to
```
mgiuca: This wording is a little strange. "The end-user indicated that <X>." This doesn't exactly convey that X should happen, since it's qualified by "the end-user indicated that". I don't think you should be mandating *what* the UA should do here; these are enum values after all, not algorithms. You should just describe what the enum represents (i.e., "The end-user indicated that they would like to install the web application.") Save the instruction for the UA for the "steps to install a web application" algorithm.

index.html
```diff
@@ -231,6 +231,43 @@
...
+            <p class="note">
+              This does not necessarily mean that the <a>installation
+              process</a> will <a data-lt="installation succeeded">succeed</a>.
+            </p>
+          </dd>
+          <dt>
+            <dfn>dismissed</dfn>:
+          </dt>
+          <dd>
+            The end-user dismissed the install prompt, thus the user agent MUST
```
mgiuca: Similarly, it doesn't make sense to dictate what the UA MUST do here, since it is an enum value. Just "The end-user dismissed the install prompt."

index.html
marcoscaceres: Agree. Will fix.

index.html
```diff
@@ -231,6 +231,41 @@
...
+          end-user with an install prompt</a>.
+        </p>
+        <dl data-dfn-for="AppBannerPromptOutcome">
+          <dt>
+            <dfn>accepted</dfn>:
+          </dt>
+          <dd>
+            The end-user indicated that they would like the user agent to
+            <a>install</a> the web application.
+            <p class="note">
```
mgiuca: In the same spirit as my comments yesterday, I don't think this note is necessary. It's not an instruction, it's an enum value indicating a user choice. There should be no misapprehension that this implies success.

marcoscaceres: ok, will remove.

marcoscaceres: This has now been removed.

PR #520, opened 2016-11-07
Feat(events): add BeforeInstallPromptEvent

marcoscaceres:
Defines `BeforeInstallPromptEvent`, related Dictionaries, and `.prompt()` method as recently discussed in the relevant bug.

### DIFF
--- a/index.html
+++ b/index.html
@@ -435,18 +435,17 @@
             <a>Queue a task</a> on the <a>application life-cycle task
             source</a> to do the following:
             <ol>
-              <li>Let <var>event</var> be a <a data-lt=
-              "construct a BeforeInstallPromptEvent">newly constructed</a> <a>
-                BeforeInstallPromptEvent</a> named
-                "<a>beforeinstallprompt</a>", which is cancelable.
+              <li>Let <var>target</var> be at <a>Window</a> object of the
+              <a>top-level browsing context</a>.
               </li>
-              <li>
-                <a>Fire</a> <var>event</var> at the <a>Window</a> object of the
-                <a>top-level browsing context</a>.
+              <li>Let <var>showPrompt</var> be the result of <a>firing an
+              event</a> (<var>event</var>) named "beforeinstallprompt", using
+              <a>BeforeInstallPromptEvent</a> at <var>target</var> with its
+              <code>cancelable</code> attribute initialized to true.
               </li>
-              <li>If <var>event</var>'s <a>canceled flag</a> is not set, then,
-              <a>in parallel</a>, <a>request to present an install prompt</a>
-              with <var>event</var>.
+              <li>If <var>showPrompt</var> is true, then, <a>in parallel</a>,
+              <a>request to present an install prompt</a> with
+              <var>event</var>.
               </li>
             </ol>
           </li>
@@ -526,6 +525,125 @@
         DOM events <a>fired</a> by this specification use the <dfn>application
         life-cycle task source</dfn>.
       </p>
+      <section data-dfn-for="BeforeInstallPromptEvent">
+        <h3>
+          <code>BeforeInstallPromptEvent</code> Interface
+        </h3>
+        <pre class="idl">
+          [Constructor]
+          interface BeforeInstallPromptEvent : Event {
+              Promise&lt;PromptResponseObject&gt; prompt();
+          };
+          dictionary PromptResponseObject {
+            AppBannerPromptOutcome userChoice;
+          };
+        </pre>
+        <p>
+          The <dfn>BeforeInstallPromptEvent</dfn> is dispatched prior to
+          activating an <a>automated install prompt</a>, allowing a developer
+          to prevent the default action for an install prompt.
+        </p>
+        <p>
+          Thus, the default action of the <a>BeforeInstallPromptEvent</a> is to
+          <a data-lt="presents an install prompt">present an automated install
+          prompt</a> to the end-user. Canceling the default action (via
+          <code>.preventDefault()</code>) prevents the user agent from
+          <a data-lt="presents an install prompt">presenting an automated
+          install prompt</a> until a later time (see
+          <a>BeforeInstallPromptEvent.prompt</a>() method).
+        </p>
+        <p>
+          An instance of a <a>BeforeInstallPromptEvent</a> has the following
+          internal slots:
+        </p>
+        <dl>
+          <dt>
+            <dfn>[[\didPrompt]]</dfn>
+          </dt>
+          <dd>
+            A boolean, initially <code>false</code>. Represents if this event
+            was used to <a>present an install prompt</a> to the end-user.
+          </dd>
+          <dt>
+            <dfn>[[\userResponsePromise]]</dfn>
+          </dt>
+          <dd>
+            A promise that represents the outcome of <a>presenting an install
+            prompt</a>.
+          </dd>
+        </dl>
+        <section>
+          <h4>
+            <code>prompt()</code> method
+          </h4>
+          <p>
+            The <dfn>prompt</dfn> method, when called, runs the following
+            steps:
+          </p>
+          <ol>
+            <li>Let <var>p</var> be a newly created promise.
+            </li>
+            <li>Resolve <var>p</var> with
+            <var>this</var>.<a>[[\userResponsePromise]]</a>.
+            </li>
+            <li>If <var>this</var>.<a>[[\userResponsePromise]]</a> is pending:
+              <ol>
+                <li>If this event's <code>isTrusted</code> attribute is <code>
+                  false</code>, reject
+                  <var>this</var>.<a>[[\userResponsePromise]]</a> with
+                  <a>NotAllowedError</a>, optionally informing the developer
+                  that untrusted events can't call <code>prompt()</code>.
+                </li>
+                <li>Else if <var>this</var>.<a>[[\didPrompt]]</a> is
+                <code>false</code>, then, <a>in parallel</a>, <a>request to
+                present an install prompt</a> with this event. Wait, possibly
+                indefinitely, for the end-user to make a choice.
+                </li>
+              </ol>
+            </li>
+            <li>Return <var>p</var>.
+            </li>
+          </ol>
+          <p>
+            To <dfn>request to present an install prompt</dfn> with
+            <a>BeforeInstallPromptEvent</a> <var>event</var>:
+          </p>
+          <ol>
+            <li>
+              <a>Present an install prompt</a> and let <var>outcome</var> be
+              the result.
+            </li>
+            <li>Let <var>responseObj</var> be a newly created
+            <a>PromptResponseObject</a> whose <code>userChoice</code> member is
+            the value of <var>outcome</var>.
+            </li>
+            <li>Resolve <var>event</var>.<a>[[\userResponsePromise]]</a> with
+            <var>responseObj</var>.
+            </li>
+          </ol>
+        </section>
+        <section class="informative">
+          <h4>
+            Usage example
+          </h4>
+          <p>
+            This example shows how one might prevent an automated install
+            prompt from showing until the user has finished a set of tasks.
+            Those tasks are represented as an array of promises, which the
+            application "awaits" to finish before an install prompt is
+            presented to the end-user.
+          </p>
+          <pre class="example" title="'beforeinstallprompt' in action">
+            window.addEventListener("beforeinstallprompt", async (event) =&gt; {
+              event.preventDefault();
+              // Wait for e.g., the user to request installation from inside the app.
+              await Promise.all(tasksThatPreventsInstallation);
+              const { userChoice } = await event.prompt();
+              console.info(`user choice was: ${userChoice}`);
+            });
+          </pre>
+        </section>
+      </section>
       <section>
         <h3>
           Extensions to the <code>Window</code> object
@@ -3507,7 +3625,7 @@ Content-Security-Policy: img-src icons.example.com
         <li>
           <a href=
           "https://dom.spec.whatwg.org/#concept-event-fire"><dfn data-lt=
-          "fire|fired">Fire an event</dfn></a>
+          "fire|fired|firing an event">Fire an event</dfn></a>
         </li>
       </ul>
       <p>
--- a/index.html
+++ b/index.html
@@ -435,13 +435,14 @@
             <a>Queue a task</a> on the <a>application life-cycle task
             source</a> to do the following:
             <ol>
-              <li>Let <var>target</var> be at <a>Window</a> object of the
-              <a>top-level browsing context</a>.
-              </li>
-              <li>Let <var>showPrompt</var> be the result of <a>firing an
-              event</a> (<var>event</var>) named "beforeinstallprompt", using
-              <a>BeforeInstallPromptEvent</a> at <var>target</var> with its
+              <li>Let <var>event</var> be a newly constructed
+              <a>BeforeInstallPromptEvent</a> named
+              <code>beforeinstallprompt</code>, with its
               <code>cancelable</code> attribute initialized to true.
+              </li>
+              <li>Let <var>showPrompt</var> be the result of <a>firing</a>
+              <var>event</var> at the <a>Window</a> object of the <a>top-level
+              browsing context</a>.
               </li>
               <li>If <var>showPrompt</var> is true, then, <a>in parallel</a>,
               <a>request to present an install prompt</a> with
@@ -450,6 +451,16 @@
             </ol>
           </li>
         </ol>
+        <div class="issue">
+          Implementations may wish to show a prompt if, and only if, the site
+          explicitly requests it via
+          <a><code>BeforeInstallPromptEvent.prompt()</code></a>, but not
+          automatically without the site's approval. Is this something we want
+          to leave to the discretion of the user agent? (This would require
+          changing the language here, as "steps to notify before an automated
+          install prompt" would become "steps to notify that an install prompt
+          is available".)
+        </div>
       </section>
       <section>
         <h3 id="installation-sec">
@@ -543,14 +554,23 @@
           activating an <a>automated install prompt</a>, allowing a developer
           to prevent the default action for an install prompt.
         </p>
-        <p>
-          Thus, the default action of the <a>BeforeInstallPromptEvent</a> is to
+        <div class="note">
+          The default action of the <a>BeforeInstallPromptEvent</a> is to
           <a data-lt="presents an install prompt">present an automated install
           prompt</a> to the end-user. Canceling the default action (via
-          <code>.preventDefault()</code>) prevents the user agent from
-          <a data-lt="presents an install prompt">presenting an automated
-          install prompt</a> until a later time (see
-          <a>BeforeInstallPromptEvent.prompt</a>() method).
+          <a data-cite=
+          "DOM#dom-event-preventdefault"><code>preventDefault</code></a>)
+          prevents the user agent from <a data-lt=
+          "presents an install prompt">presenting an automated install
+          prompt</a>. The user agent is free to run <a>steps to notify before
+          an automated install prompt</a> again at a later time.
+        </div>
+        <p>
+          The <dfn>PromptResponseObject</dfn> contains the result of calling
+          <a for="BeforeInstallPromptEvent"><code>prompt()</code></a>. It
+          contains one member, <dfn for=
+          "PromptResponseObject">userChoice</dfn>, which states the user's
+          chosen outcome.
         </p>
         <p>
           An instance of a <a>BeforeInstallPromptEvent</a> has the following
@@ -561,8 +581,8 @@
             <dfn>[[\didPrompt]]</dfn>
           </dt>
           <dd>
-            A boolean, initially <code>false</code>. Represents if this event
-            was used to <a>present an install prompt</a> to the end-user.
+            A boolean, initially <code>false</code>. Represents whether this
+            event was used to <a>present an install prompt</a> to the end-user.
           </dd>
           <dt>
             <dfn>[[\userResponsePromise]]</dfn>
@@ -581,27 +601,24 @@
             steps:
           </p>
           <ol>
-            <li>Let <var>p</var> be a newly created promise.
-            </li>
-            <li>Resolve <var>p</var> with
-            <var>this</var>.<a>[[\userResponsePromise]]</a>.
-            </li>
             <li>If <var>this</var>.<a>[[\userResponsePromise]]</a> is pending:
               <ol>
-                <li>If this event's <code>isTrusted</code> attribute is <code>
-                  false</code>, reject
-                  <var>this</var>.<a>[[\userResponsePromise]]</a> with
-                  <a>NotAllowedError</a>, optionally informing the developer
-                  that untrusted events can't call <code>prompt()</code>.
+                <li>If this event's <a data-cite=
+                "!DOM#dom-event-istrusted"><code>isTrusted</code></a> attribute
+                is <code>false</code>, reject
+                <var>this</var>.<a>[[\userResponsePromise]]</a> with
+                <a>NotAllowedError</a>, optionally informing the developer that
+                untrusted events can't call <code>prompt()</code>.
                 </li>
                 <li>Else if <var>this</var>.<a>[[\didPrompt]]</a> is
-                <code>false</code>, then, <a>in parallel</a>, <a>request to
+                <code>false</code>, set <var>this</var>.<a>[[\didPrompt]]</a>
+                to <code>true</code>, then <a>in parallel</a>, <a>request to
                 present an install prompt</a> with this event. Wait, possibly
                 indefinitely, for the end-user to make a choice.
                 </li>
               </ol>
             </li>
-            <li>Return <var>p</var>.
+            <li>Return <var>this</var>.<a>[[\userResponsePromise]]</a>.
             </li>
           </ol>
           <p>
@@ -613,12 +630,10 @@
               <a>Present an install prompt</a> and let <var>outcome</var> be
               the result.
             </li>
-            <li>Let <var>responseObj</var> be a newly created
-            <a>PromptResponseObject</a> whose <code>userChoice</code> member is
-            the value of <var>outcome</var>.
-            </li>
-            <li>Resolve <var>event</var>.<a>[[\userResponsePromise]]</a> with
-            <var>responseObj</var>.
+            <li>Resolve <var>event</var>.<a>[[\userResponsePromise]]</a> with a
+            newly created <a>PromptResponseObject</a> whose <a for=
+            "PromptResponseObject">userChoice</a> member is the value of <var>
+              outcome</var>.
             </li>
           </ol>
         </section>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
This is what the text is based on:
```
class BeforeInstallPromptEvent extends Event {
  constructor(typeArg, eventInit) {
    // WebIDL Guard. Not in spec, as it's all handled by WebIDL.
    if (arguments.length === 0) {
      throw new TypeError("Not enough arguments. Expected at least 1.");
    }
    const initType = typeof eventInit;
    if (arguments.length === 2 && initType !== "undefined" && initType !== "object") {
      throw new TypeError("Value can't be converted to a dictionary.");
    }
    super(typeArg, Object.assign({ cancelable: true }, eventInit));

    if (eventInit && !hasValidUserChoice(eventInit)) {
      const msg = `The provided value '${eventInit.userChoice}' is not a valid` +
        " enum value of type AppBannerPromptOutcome.";
      throw new TypeError(msg);
    }
    // End WebIDL guard.
    let outcome = null;
    const internal = {
      didPrompt: false,
    };
    internalSlots.set(this, internal);
    internal.userResponsePromise = new Promise((resolve, reject) => {
      const props = {
        "promptOutcome": {
          get() {
            return outcome;
          },
          set: function(value) {
            if (outcome) {
              return;
            }
            outcome = value;
            resolve(createUserResponseObject(this));
          }.bind(this), // 'this' is the BeforeInstallPromptEvent instance
        },
        "rejectPromptPromise": {
          value(domException) {
            reject(domException);
          },
        },
      };
      Object.defineProperties(internal, props);
    });
    if (eventInit && eventInit.userChoice) {
      internal.promptOutcome = eventInit.userChoice;
    }
  }
  /**
   * prompt() method
   *
   * @return {[Promise]} The internal [[\userResponsePromise]]
   * @see https://www.w3.org/TR/appmanifest/#prompt-method
   */
  async prompt() {
    const internal = internalSlots.get(this);
    if (internal.promptOutcome !== null) {
      return internal.userResponsePromise;
    }
    if (!this.isTrusted) {
      const msg = ".prompt() can only called by trusted events.";
      internal.rejectPromptPromise(new DOMException(msg, "NotAllowedError"));
    } else if (!internal.didPrompt) {
      await requestInstallPrompt(this);
    }
    return internal.userResponsePromise;
  }
}
```

marcoscaceres:
There is one more medium sized PR coming tomorrow that defines the final steps.

marcoscaceres:
Fixed a couple of little things, so updated PR...

index.html
```diff
@@ -447,6 +447,171 @@ <h3 id="installation-sec">
...
         life-cycle task source</dfn>.
       </p>
+      <section data-dfn-for="BeforeInstallPromptEvent">
+        <h3>
+          <code>BeforeInstallPromptEvent</code> Interface
+        </h3>
+        <pre class="idl">
+          [Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+          interface BeforeInstallPromptEvent : Event {
+              readonly attribute Promise&lt;AppBannerPromptOutcome&gt; userChoice;
```
marcoscaceres: Oops, forgot to remove... removing now

index.html
```diff
@@ -447,6 +447,170 @@ <h3 id="installation-sec">
...
+          [Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+          interface BeforeInstallPromptEvent : Event {
+              Promise&lt;UserResponseObject&gt; prompt();
+          };
+
+          dictionary BeforeInstallPromptEventInit : EventInit {
+            AppBannerPromptOutcome userChoice;
+          };
+
+          dictionary UserResponseObject {
```
mgiuca: Why not PromptResponseObject?

index.html
```diff
@@ -447,6 +447,170 @@ <h3 id="installation-sec">
...
+          Thus, the default action of the <a>BeforeInstallPromptEvent</a> is to
+          <a data-lt="presents an install prompt">present an automated install
+          prompt</a> to the end-user. Canceling the default action (via
+          <code>.preventDefault()</code>) prevents the user agent from
+          <a data-lt="presents an install prompt">presenting an automated
+          install prompt</a> until a later time (see
+          <a>BeforeInstallPromptEvent.prompt</a>() method).
+        </p>
+        <p>
+          The <a>BeforeInstallPromptEvent</a> has three internal slots, which
```
mgiuca: Drop "three"; that sort of language can get out of date without anyone noticing.

index.html
```diff
@@ -447,6 +447,170 @@ <h3 id="installation-sec">
...
+          <code>BeforeInstallPromptEvent</code> Interface
+        </h3>
+        <pre class="idl">
+          [Constructor(DOMString typeArg, optional BeforeInstallPromptEventInit eventInit)]
+          interface BeforeInstallPromptEvent : Event {
+              Promise&lt;UserResponseObject&gt; prompt();
+          };
+
+          dictionary BeforeInstallPromptEventInit : EventInit {
+            AppBannerPromptOutcome userChoice;
```
mgiuca: I'm still not a fan of allowing this as an argument (I don't see why this is ever useful and it's weird that it's an outcome and not a promise). I don't remember the outcome of the discussion last time I brought it up. I want to have a closer look tomorrow.

index.html
marcoscaceres: Like it! Will rename.

index.html
marcoscaceres: yeah, it already bit me once.

marcoscaceres:
> Looks pretty good. I have some quick comments, but I haven't had a deeper look at it and I want to do that tomorrow (on my workstation where I've got the repo checked out so I can build it and look at it in HTML). Can we wait until then to land this?

Absolutely. Sorry, I should have posted a link where you can view the rendered version: 
https://rawgit.com/w3c/manifest/BIP/index.html

marcoscaceres:
Rebased, so all the parts are now in place (be sure to "shift-refresh" if viewing the rawgit link).

index.html
```diff
@@ -490,6 +490,170 @@ <h3 id="installation-sec">
...
+          </dt>
+          <dd>
+            A boolean, initially <code>false</code>. Represents if this event
+            was used to <a>present an install prompt</a> to the end-user.
+          </dd>
+          <dt>
+            <dfn>[[\promptOutcome]]</dfn>
+          </dt>
+          <dd>
+            A <a>AppBannerPromptOutcome</a> enum value, initially set to
```
mgiuca: s/A/An

index.html
```diff
@@ -490,6 +490,170 @@ <h3 id="installation-sec">
...
+            <code>null</code>. Represents the outcome of <a>presenting an
+            install prompt</a>.
+          </dd>
+          <dt>
+            <dfn>[[\userResponsePromise]]</dfn>
+          </dt>
+          <dd>
+            A promise, which resolves with an <a>PromptResponseObject</a>,
```
mgiuca: s/an/a

index.html
```diff
@@ -490,6 +490,170 @@ <h3 id="installation-sec">
...
+            interface, run the <dfn data-lt=
+            "construct a BeforeInstallPromptEvent">steps to construct a
+            <code>BeforeInstallPromptEvent</code> event</dfn>:
+          </p>
+          <ol>
+            <li>Let <a>[[\didPrompt]]</a> be <code>false</code>.
+            </li>
+            <li>Let <a>[[\userResponsePromise]]</a> be a newly created promise.
+            </li>
+            <li>Let <a>[[\promptOutcome]]</a> be <code>null</code>.
```
mgiuca: I think promptOutcome is unnecessary.

If you search the remaining text for promptOutcome, note that it is always assigned to, never read from, except:

1. "If [[promptOutcome]] is not null" -- that's just using it as a Boolean not an outcome value.
2. "whose userChoice member is set to event's [[promptOutcome]]." -- promptOutcome was just set to a value immediately above, so it doesn't have to be a member of BIPE, it can just be a local variable to that algorithm.

Therefore, promptOutcome can be replaced with a Boolean. But I suspect we can simplify it further.

index.html
```diff
@@ -490,6 +490,170 @@ <h3 id="installation-sec">
...
+            The <dfn>prompt</dfn> method, when called, runs the following
+            steps:
+          </p>
+          <ol>
+            <li>Let <var>p</var> be a newly created promise.
+            </li>
+            <li>Run the following steps <a>in parallel</a>:
+              <ol>
+                <li>If <a>[[\promptOutcome]]</a> is not <code>null</code>,
+                resolve <var>p</var> with <a>[[\userResponsePromise]]</a> and
```
mgiuca: Wait, "resolve p with userResponsePromise"? That means to resolve a promise with a promise. I think you mean "resolve p with promptOutcome"... but I'm trying to eliminate this (see above).

Why is p a new promise at all? Why can't we just have prompt() return userResponsePromise?

Then we can kill promptOutcome entirely.

marcoscaceres: > Wait, "resolve p with userResponsePromise"? That means to resolve a promise with a promise. I think you mean "resolve p with promptOutcome"... but I'm trying to eliminate this (see above)

Resolving the promise with the promise unwraps the internal promise - so not to expose the internal promise... but I'm trying to simplify as you suggested. Also distracted :)

mgiuca: Oh so your intention was specifically to not expose the internal promise? (i.e, every time you call prompt() it's a different promise?)

Is there a good reason to do that?

marcoscaceres: Ok... still in shock... back to this... 

> Why is p a new promise at all? Why can't we just have prompt() return userResponsePromise?

It avoids exposing the internal promise. Then .prompt() can vend promises that all get resolved when the internal one does.

index.html
marcoscaceres: Brilliant! You are right... just updating the JS implementation. I'll try to post an update later tonight.

index.html
marcoscaceres: > Oh so your intention was specifically to not expose the internal promise? (i.e, every time you call prompt() it's a different promise?)

Yeah, I realized we could just do that after the discussion we had in the bug (the discussion about e.prompt() === e.prompt()). I think returning a new promise works nicely because in reality the internal promise just becomes an implementation detail. But it will be nice in the spec because it makes the behavior quite nice and clear, IMHO.

marcoscaceres:
@mgiuca, I updated .prompt() based on your feedback. I made the following changes:

 1. Remove redundant [[\promptOutcome]]  
 1. Refactored .prompt() to no longer rely on  [[\promptOutcome]]  
 1. Still resolving .prompt() with [[\userResponsePromise]]. 

I also updated the example implementation in https://github.com/w3c/manifest/pull/520#issuecomment-258789304

@dominic, if you have a few mins, I would really appreciate if you could give us some feedback on: https://rawgit.com/w3c/manifest/BIP/index.html#beforeinstallpromptevent-interface

The prototype implementation on which the spec text is based: https://github.com/w3c/manifest/pull/520#issuecomment-258789304

marcoscaceres:
err, I mean @domenic... fixed above.

domenic:
I'm a bit unclear why the promise fulfills with an object which is just a wrapper around userChoice, instead of with userChoice directly. Maybe it's future extensibility.

All the internal slots have an extra slash in them in the rendered output.

The constructor is a bit troublesome, since events all share the same constructor logic in most implementations, namely https://dom.spec.whatwg.org/#constructing-events. Maybe that logic needs to move to the prompt method? That seems like it will change things a decent bit.

"request to present an install prompt" isn't cross-linking correctly.

When referring to internal slots, it seems better to use obj.[[slotName]] notation. Currently obj is missing, which is a bit confusing since it differs: in prompt() it should be `this`, whereas in "request to present an install prompt" it should be _event_, I think.

index.html
```diff
@@ -490,6 +490,157 @@ <h3 id="installation-sec">
...
+          <dd>
+            A boolean, initially <code>false</code>. Represents if this event
+            was used to <a>present an install prompt</a> to the end-user.
+          </dd>
+          <dt>
+            <dfn>[[\userResponsePromise]]</dfn>
+          </dt>
+          <dd>
+            A promise, which resolves with a <a>PromptResponseObject</a>, which
+            represent the outcome of <a>presenting an install prompt</a>.
```
mgiuca: represents

index.html
```diff
@@ -490,6 +490,157 @@ <h3 id="installation-sec">
...
+            <li>Return <var>p</var>.
+            </li>
+          </ol>
+          <p>
+            To <dfn>request to present an install prompt</dfn> with
+            <a>BeforeInstallPromptEvent</a> <var>event</var>:
+          </p>
+          <ol>
+            <li>
+              <a>Present an install prompt</a> and <var>outcome</var> be the
```
mgiuca: and *let* outcome be the result

index.html
```diff
@@ -490,6 +490,157 @@ <h3 id="installation-sec">
...
+            This example shows how one might prevent an automated install
+            prompt from showing until the user has finished a set of tasks.
+            Those tasks are represented as an array of promises, which the
+            application "awaits" to finish before an install prompt is
+            presented to the end-user.
+          </p>
+          <pre class="example" title="'beforeinstallprompt' in action">
+            window.addEventListener("beforeinstallprompt", async (event) =&gt; {
+              event.preventDefault();
+              // await e.g., user composing an email...
```
mgiuca: This comment is a bit confusing. I think I understand it (if the user is composing an email in the application you would wait until they finish the task before prompting). But I think that's just a bit confusing of a use case and the primary use case for this is not to avoid interrupting the user, but to present in-context UI for prompting.

So how about:
```js
// Wait for e.g., the user to request installation from inside the app.
```

index.html
```diff
@@ -490,6 +490,157 @@ <h3 id="installation-sec">
...
+              await Promise.all(tasksThatPreventsInstallation);
+              const { userChoice } = await event.prompt();
+              console.info(`user selected: ${userChoice}`);
```
mgiuca: How about `"user choice was: ${uc}"` because the grammar would be weird of "user selected: accepted".

marcoscaceres:
@domenic wrote, 
> I'm a bit unclear why the promise fulfills with an object which is just a wrapper around userChoice, instead of with userChoice directly. Maybe it's future extensibility.

Yeah, it's for extensibility: The (non-standard) Chrome implementation currently has additional information being returned (e.g., "platforms", which could be "Web, Play Store,"), but we are still discussing if that will be included. It's likely we will add more things. 

> All the internal slots have an extra slash in them in the rendered output.

Will fix. 

> The constructor is a bit troublesome, since events all share the same constructor logic in most implementations, namely https://dom.spec.whatwg.org/#constructing-events. Maybe that logic needs to move to the prompt method? That seems like it will change things a decent bit.

Ah, yeah. Although the .prompt() logic can probably stay the same, and I can just remove the constructor prose. The internal slots can be set as private instance properties without involvement of the constructor (i.e., they are an implementation detail in a sense). 

> "request to present an install prompt" isn't cross-linking correctly.

Fixing.

> When referring to internal slots, it seems better to use obj.[[slotName]] notation. Currently obj is missing, which is a bit confusing since it differs: in prompt() it should be this, whereas in "request to present an install prompt" it should be event, I think.

Fixing.

marcoscaceres:
@mgiuca, seems you were right about not having `{userChoice}` be passed during construction. It's incompatible with DOM's event creation. My apologies.  
https://github.com/whatwg/dom/issues/372#issuecomment-259598448

On the upside, makes this much simpler.

marcoscaceres:
Ok, I've integrated all feedback. 

The final thing is if we stick with resolving with `PromptResponseObject` (safe option if we want to add additional things in the future, like .platforms) - or  we feel confident we won't and we just resolve prompt with `AppBannerPromptOutcome` (i.e., "dismissed", "accepted").

mgiuca:
I'd like to resolve with a `PromptResponseObject`. It's low cost and extensible into the future.

domenic:
The constructor should be `[Constructor(DOMString type, optional EventInit eventInitDict)]`. (Guidance incoming: https://github.com/whatwg/dom/issues/364.)

You should state the "newly created PromptResponseObject" is created in the relevant Realm of _event_. See https://github.com/heycam/webidl/issues/135. Most specs are bad about this but we're trying to be better going forward :).

RobDolinMS:
I'm on a thread with @NolanLawson @JacobRossi and Ali Alabbas about this in @MSEdge

marcoscaceres:
@RobDolinMS and Microsoft folks - currently discussing with Google as to how we could improve/simplify this even further. We would love to chat to you folks too.  @NolanLawson or @JacobRossi, if you are at Chrome Dev Summit, grab @slightlyoff to discuss further.

kenchris:
Any update on this?

RobDolinMS:
@marcoscaceres @kenchris Sorry for non-response on this issue; working with other MSFT stakeholders to get to a consensus response.  I hope to have something later this week.

RobDolinMS:
Met with other Microsoft stakeholders on this topic yesterday and we're a LGTM for this PR.

index.html
```diff
@@ -490,6 +489,125 @@ <h3 id="installation-sec">
...
+                <li>If this event's <code>isTrusted</code> attribute is <code>
+                  false</code>, reject
+                  <var>this</var>.<a>[[\userResponsePromise]]</a> with
+                  <a>NotAllowedError</a>, optionally informing the developer
+                  that untrusted events can't call <code>prompt()</code>.
+                </li>
+                <li>Else if <var>this</var>.<a>[[\didPrompt]]</a> is
+                <code>false</code>, then, <a>in parallel</a>, <a>request to
+                present an install prompt</a> with this event. Wait, possibly
+                indefinitely, for the end-user to make a choice.
```
mgiuca: Need to set didPrompt to true here (otherwise multiple prompts can be shown if this is called again before the first one resolves).

index.html
```diff
@@ -490,6 +489,125 @@ <h3 id="installation-sec">
...
+        <section>
+          <h4>
+            <code>prompt()</code> method
+          </h4>
+          <p>
+            The <dfn>prompt</dfn> method, when called, runs the following
+            steps:
+          </p>
+          <ol>
+            <li>Let <var>p</var> be a newly created promise.
```
mgiuca: Why create *p* and immediately resolve it with `[[userResponsePromise]]`?

Can't you simply delete 1 and 2 here, and in the final step, return `[[userResponsePromise]]`?

index.html
```diff
@@ -490,6 +489,125 @@ <h3 id="installation-sec">
...
+            prompt from showing until the user has finished a set of tasks.
+            Those tasks are represented as an array of promises, which the
+            application "awaits" to finish before an install prompt is
+            presented to the end-user.
+          </p>
+          <pre class="example" title="'beforeinstallprompt' in action">
+            window.addEventListener("beforeinstallprompt", async (event) =&gt; {
+              event.preventDefault();
+              // Wait for e.g., the user to request installation from inside the app.
+              await Promise.all(tasksThatPreventsInstallation);
```
mgiuca: I think I've said this before elsewhere but I'm not a huge fan of this example. It's quite unclear what "tasks that prevents installation" means, and it doesn't speak to the main use case we have in mind for this feature (which is not blocking on some task, but waiting for user input, e.g., clicking an "install me" button on a site).

Can we rework this to have the code be

1. Upon receipt of the beforeinstallprompt event, set a button in the site UI to "enabled"?
2. Await pressing the button.
3. Call `prompt()`.

marcoscaceres:
@mgiuca, thanks for the feedback. Will try to address the last few comments soon.

marcoscaceres:
@kenchris, actually, can you take this (only small changes left)? I'm bogged down with web payments stuff + have a f2f to attend next week.

mgiuca:
I can take it and work on it (also happy for Kenneth to take it).

marcoscaceres:
@mgiuca, if you can, that would be great!

kenchris:
If you need help, tell me

mgiuca:
OK, I will take this.

mgiuca:
(That commit 0a05f3e was a rebase.)

mgiuca:
I just did a pass over the beforeinstallprompt spec and pushed my changes (i.e., the delta between Marcos' last commit 0a05f3e and mine, 4bb2977).

Preview on [rawgit](https://rawgit.com/w3c/manifest/BIP/index.html).

Kenneth, do you want to have a look?

index.html
```diff
@@ -435,22 +435,32 @@
...
-              with <var>event</var>.
+              <li>If <var>showPrompt</var> is true, then, <a>in parallel</a>,
+              <a>request to present an install prompt</a> with
+              <var>event</var>.
               </li>
             </ol>
           </li>
         </ol>
+        <div class="issue">
+          Implementations may wish to show a prompt if and only if the site
```
kenchris: comma around if and only if?

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+          to prevent the default action for an install prompt.
+        </p>
+        <div class="note">
+          The default action of the <a>BeforeInstallPromptEvent</a> is to
+          <a data-lt="presents an install prompt">present an automated install
+          prompt</a> to the end-user. Canceling the default action (via
+          <a
+          data-cite="DOM#dom-event-preventdefault"><code>preventDefault</code></a>)
+          prevents the user agent from <a data-lt="presents an install
+          prompt">presenting an automated install prompt</a> until a later time
```
kenchris: *at* a later time

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+        <p>
+          An instance of a <a>BeforeInstallPromptEvent</a> has the following
+          internal slots:
+        </p>
+        <dl>
+          <dt>
+            <dfn>[[\didPrompt]]</dfn>
+          </dt>
+          <dd>
+            A boolean, initially <code>false</code>. Represents if this event
```
kenchris: represents whether? maybe if is fine... just sounds weird to me :)

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+          <p>
+            To <dfn>request to present an install prompt</dfn> with
+            <a>BeforeInstallPromptEvent</a> <var>event</var>:
+          </p>
+          <ol>
+            <li>
+              <a>Present an install prompt</a> and let <var>outcome</var> be
+              the result.
+            </li>
+            <li>Let <var>responseObj</var> be a newly created
```
kenchris: do we need the Obj here?

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+            </li>
+          </ol>
+        </section>
+        <section class="informative">
+          <h4>
+            Usage example
+          </h4>
+          <p>
+            This example shows how one might prevent an automated install
+            prompt from showing until the user clicks a button to install the
```
kenchris: do we need some examples to when it makes sense to delay?

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+            app. In this way, the site can control the installation experience.
+          </p>
+          <pre class="example"
+               title="Using beforeinstallprompt to present an install button">
+            window.addEventListener("beforeinstallprompt", async (event) =&gt; {
```
kenchris: do you need the () around event here? is that because of async... normally you dont with one argument

index.html
```diff
@@ -526,14 +536,150 @@ <h3 id="installation-sec">
...
+              // Suppress automatic prompting.
+              event.preventDefault();
+
+              // Show the (disabled-by-default) install button. This button
+              // resolves the installButtonClicked promise when clicked.
+              installButton.disabled = false;
+
+              // Wait for the user to click the button.
+              await installButtonClicked;
```
kenchris: I think it is a bit magic where this variable comes from

index.html
mgiuca: I think comma around "and only if" is correct.

(Why can't I do many comments together? Maybe this pull request was started so long ago that it's grandfathered out.)

index.html
mgiuca: Ah I think you misinterpreted this sentence. (Marcos wrote it and I just got confused myself now.) It's not trying to prevent prompting later. It's trying to prevent prompting *now*, but saying that the UA can prompt again later. I've rewritten it for clarity. WDYT?

index.html
mgiuca: Done.

index.html
mgiuca: Do you mean can we just combine these into a single step? Agree... done.

index.html
mgiuca: Do you mean to explain why it makes sense to use this rather than just letting it happen?

I reworded that second sentence thusly:
"In this way, the site can leave installation at the user's discretion (rather than prompting at an arbitrary time), whilst still providing a prominent UI to do so."

index.html
mgiuca: It isn't a syntax error (but I didn't confirm whether it works; I assume it does). Done.

index.html
mgiuca: Hmm yeah. async/await is nice but a bit magic here.

Refactored into a traditional addEventListener setup. The outer function no longer needs to be async.

index.html
kenchris: ok

mgiuca:
I addressed those comments.

Meta-review questions:

1. I see a bunch of people using htmltidy on their specs. I tried to use this but it complained about the special respec elements and refused to do anything. Is there a trick for getting it to work on respec documents?
2. Do you prefer me to squash all the commits into one before landing this (or perhaps Marcos' one + my changes), or keep the long tail of git commits?

PTAL. Thanks!

kenchris:
I think that squashing is nicer... But you could do one commit for Marcos work and one for your modifications

mgiuca:
> cannot use await now though (not async method...)

The inner function is async, and that's where I use await. I tested it, and it works. (The outer function does not need to be async since it just immediately does things including setting up a click handler.)

> I think that squashing is nicer... But you could do one commit for Marcos work and one for your modifications

Squashed into two commits.

Also, how about my HTMLTidy question?

> I see a bunch of people using htmltidy on their specs. I tried to use this but it complained about the special respec elements and refused to do anything. Is there a trick for getting it to work on respec documents?

Lastly, please wait for me to get some more people at Google to look at this before merging. Thanks!

marcoscaceres:
@mgiuca 
> I see a bunch of people using htmltidy on their specs. I tried to use this but it complained about the special respec elements and refused to do anything. Is there a trick for getting it to work on respec documents?

No, there should be no trick. Make sure you are using [HTML Tidy version 5.4.0](https://github.com/htacg/tidy-html5) (I think that is latest). 

 Then just:

```Bash
tidy -config tidyconfig.txt -o index.html index.html
```

mgiuca:
Thanks. I must've been using a very old version of tidy. (Had to build from GitHub source.)

Now tidied (and squashed that into my commit).

mgiuca:
FYI, I wrote up a doc with the things we're thinking about on Chrome before we close this out:
[Chrome change proposal](https://docs.google.com/document/d/1ONRGrtTqZj0mKWhVWJRWic9UfmsAvuZmy76s4-75y_M/edit)

This lists the things we'll have to change in Chrome to become compliant with this spec. I'm checking with some people internally whether we're OK to make these changes.

kenchris:
I cannot access that doc

mgiuca:
Sorry, fixed.

marcoscaceres:
@mgiuca, did you get an ok to change Blink to match? Should we merge?

mgiuca:
@marcoscaceres Almost... I am giving @slightlyoff one more day because I haven't heard back. Unless he objects, I think we will merge as-is. Thanks for your patience.

marcoscaceres:
np. No hurry.

mgiuca:
Hey,

Good you merged :+1:  :)

For the record, I am still having a back-and-forth with Alex. He had some concerns (not about the spec per se but the compatibility issues if we change Chrome). But I am happy this is finally landed. We will continue to figure out the migration course for Chrome, and if we have any issues with the spec we can open new issues instead of continuing this megathread.

THANKS!

PR #522, opened 2016-11-08
feat(installation): define when/how to fire BIP

marcoscaceres:
Specify how the browser automatically decides when to fire the BeforeInstallPromptEvent event.


marcoscaceres:
Actually, this is in the wrong place, sending again.

PR #524, opened 2016-11-08
feat(installation): define when/how to fire BIP

marcoscaceres:


marcoscaceres:
Ok, once this final one lands, then I'm going to do some editorial reorganization of the editorial section. But let's get this in and let's get implementing!!! ❤️

index.html
```diff
@@ -374,6 +360,67 @@
         </ol>
       </section>
       <section>
+        <h2>
+          Install prompts
+        </h2>
+        <p>
+          An end-user can <dfn data-lt="manual installation">manually</dfn>
```
kenchris: is that a requirement that that is possible?

marcoscaceres:
WARNING: IF THIS DOESN'T LAND BY TOMORROW TRUMP GETS AUTOMATICALLY ELECTED!

kenchris:
Does that means that if he gets elected, you drop this patch?

marcoscaceres:
> Does that means that if he gets elected, you drop this patch?

Will it matter at that point?

kenchris:
im fine with this (patch that is)

index.html
marcoscaceres: No, it's just a statement of fact.

marcoscaceres:
@mgiuca see https://github.com/w3c/manifest/pull/520

PR #530, opened 2016-11-18
feat: Add splashscreens member + purpose (closes #510)

### LINKED ISSUE #501: (Re-)Add Optional Splash Screens
This issue proposes resurrecting the `splash_screens` object that was removed with PR #440 

While UAs can fall-back to using one of the smaller images and a background color, offering the optional `splash screens` object gives web app developers an option to have the UA display a larger, more detailed image while their app loads.  

(Example from [prior spec version](http://www.w3.org/TR/2015/WD-appmanifest-20151124/))
```
  "splash_screens": [{
        "src": "splash/lowres",
        "sizes": "320x240"
      }, {
        "src": "splash/hd_small",
        "sizes": "1334x750"
      }, {
        "src": "splash/hd_hi",
        "sizes": "1920x1080",
        "density": 3
      }],

```

### LINKED ISSUE #510: Add splashscreen as a `purpose`
If we give regarding the icon purpose/type/kind, maybe we should add splashscreen in there to help UA pick an image if it's meant to be used for the splashscreen?

marcoscaceres:


index.html
```diff
@@ -1932,6 +1927,43 @@ <h3 id="applying">
...
+          A <dfn>splash screen</dfn> is a graphical representation (usually an
+          image on a background color) that is presented as the end user before
+          the user agent is able render content. A splash screen is intended to
+          convey to the end-user that the web application is being prepared for
+          them in the background.
+        </p>
+        <p>
+          When showing a <a>splash screen</a>, a user agent MUST announce to
+          assistive technologies that the web application is being loaded in
+          the background.
```
patrickhlauke: This general statement of intent looks good to me. Is this intended to mean the UA automatically announces this? I'm assuming in its current form, an app can't use the splash screen to show additional information (e.g. short status/progress messages of what it's loading/doing - e.g. "reticulating splines 67%" or similar ? Otherwise this would also need to be taken into account (and at that stage, it's more likely that authors will want/need to control what exactly is exposed to AT, e.g. using an `aria-live` region...

stevefaulkner: "a user agent MUST announce to assistive technologies" is unclear, suggest something in form of "users agents MUST expose the loading state of the application via the platform <a>Accessibility API</a>(s)"

kenchris:
latest changes lgtm

marcoscaceres:
@kenchris, thanks for checking. Waiting on @mounirlamouri, who originally requested this.

mounirlamouri:
Sorry, I entirely missed track of this. What's the benefit of adding a member **and** a purpose?

kenchris:
@marcoscaceres can you answer @mounirlamouri 's question?

RobDolinMS:
Rather than adding another object to the Manifest, what about just adding guidance for developers to include full screen sized icons on the icons[ ] object?

marcoscaceres:
Icons and splash screens are different things tho.

RobDolinMS:
>Icons and splash screens are different things tho.

@marcoscaceres That's definitely true; but if there is no splash screen, a larger icon centered on a background seems to be a common splash screen used by developers.  (This seems to be what Twitter and LinkedIn apps do.)

Rather than adding another object (which I'm hesitant to do since we don't want Manifests to be too complex), I'm thinking if we just encourage larger images in the `icons[ ]` array, this gives advanced web app developers a way to specify desired images to use for splashscreens; and gives new web app developers a good result from a simple manifest.

kenchris:
We need to solve images... Like with masking for round icons etc or the icons cannot be reused. The current situation leads to weird UX like Twitter Lite on the Pixel

marcoscaceres:
@kenchris, doest your adaptive icon proposal #563 solve the problem?

kenchris:
Yes that proposal fixes the issues that I mentioned about Twitter Lite and the Pixel

tomayac:
On iOS 11.3 Beta 2, Apple apparently do not follow the Chrome convention (hack?) of creating a splashscreen. However, Apple have a ```meta``` tag for the old-school iOS Web apps ([documentation](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW6)). The iOS Human Interface Guidelines have up-to-date (including iPhone X) [specs for the dimensions](https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/launch-screen/).

index.html
```diff
@@ -1930,6 +1925,43 @@ <h3>
...
       </section>
+      <section>
+        <h3>
+          <code>splashscreens</code> member
+        </h3>
+        <p>
+          A <dfn>splash screen</dfn> is a graphical representation (usually an
+          image on a background color) that is presented as the end user before
+          the user agent is able render content. A splash screen is intended to
+          convey to the end-user that the web application is being prepared for
```
cvan: `end-user` is `end user` above. can you choose the same one to be consistent?

index.html
```diff
@@ -1930,6 +1925,43 @@ <h3>
           </li>
         </ol>
       </section>
+      <section>
+        <h3>
+          <code>splashscreens</code> member
+        </h3>
+        <p>
+          A <dfn>splash screen</dfn> is a graphical representation (usually an
```
cvan: should this read `splashscreens` instead of `splash screen`?

the other member sections start [like this](https://github.com/w3c/manifest/pull/530/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR1945). is that an issue

index.html
```diff
@@ -1930,6 +1925,43 @@ <h3>
...
+          image on a background color) that is presented as the end user before
+          the user agent is able render content. A splash screen is intended to
```
cvan: a few typos caught:
- `presented as the end user ` -> `presented to the end user`
- `is able render` → `is able to render`

DongHyunSong:
I wonder when the UA decide to change the screen from splash screen to main page.

PR #533, opened 2016-12-02
[Explainer] Add very simple manifest example

RobDolinMS:
Manifests with just name and icons are quite common.  

While it's definitely not as good as a more detailed manifest, it's a step in the right direction of turning a web page into a web app.


RobDolinMS:
@marcoscaceres If there's something I need to do on the IPR front, please let me know.

marcoscaceres:

PR #540, opened 2017-01-03
Put backticks around link and meta tags

pepelsbey:
So they will render as code samples, not HTML


pepelsbey:
Otherwise they’re invisible

image

image

marcoscaceres:
Whoops! thanks.

PR #546, opened 2017-01-29
fix various things reverting back to gh-pages

marcoscaceres:


kenchris:
I would probably just leave out the t word :) You need to be special to deserve being mentioned in such an important spec. Maybe make the description more generic/global (ie. leave the US out - it is a global problem anyway)

marcoscaceres:
Agree, it was a bit of a rage response to the Muslim ban. I'm AFK, can you fix quickly?

marcoscaceres:
Fixed, btw.

PR #548, opened 2017-01-29
Removed country specificity

hexalys:
Take focus off of the U.S. in particular, to a more global stance or any other countries where the behavior may apply.


marcoscaceres:

PR #558, opened 2017-03-14
[Index | Example] Add Service Worker object

RobDolinMS:
Adds the service worker object to the complex example

@KenChris (or others) - would you please comment if you think scope should be:
 a) "/racer" - matching example in serviceworker section
 b) "/racer/" - matching the `scope` object earlier in the example

### DIFF
--- a/index.html
+++ b/index.html
@@ -155,6 +155,11 @@
   "orientation": "landscape",
   "theme_color": "aliceblue",
   "background_color": "red",
+  "serviceworker": {
+    "src": "sw.js",
+    "scope": "/racer",
+    "use_cache": false
+  },
   "screenshots": [{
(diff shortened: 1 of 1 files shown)


kenchris:
I would assume that they mean the same, assuming /races is a dictionary. If it is not, it won't work in any of the cases.

https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html

Does the SW spec say anything about this?

kenchris:
For the example, I think it makes more sense to use /races/ because that makes it clear it is a dictionary

RobDolinMS:
Thanks @kenchris.  I've updated the example your suggestion.  
If there is no additional feedback, please review and merge.  Thanks.

Issue #561, opened 2017-03-21
Study Android adaptive icons


kenchris:
https://developer.android.com/preview/features/adaptive-icons.html

@owencm @slightlyoff @marcoscaceres @RobDolinMS @anssiko

PR #562, opened 2017-03-22
WIP: Add support for adaptive icons

### LINKED ISSUE #555: Icon shapes and masking
I've been thinking about how Vanilla Android 7 has largely round icons (or at least some new Google-flavor Android does) and thus icons provided in the manifest, even with the `platform` property will look slightly strange on that OS. 

I heard some developer complain recently that they're not sure how to get the circular shape.

Taking a broader view on this problem, Samsung has their own icon mask shape, Xiaomi have their own icon shape etc.

Thus I wonder if a better solution is to have sites provide their icon as a large square with sufficient padding around the edges and then have the browser apply a mask appropriate for the OS it's running on.

Not sure how that could fit into the spec in terms of `platform` etc. Perhaps we could just use `platform: 'android_maskable'` or something, but wanted to raise it as an idea.

kenchris:
Adds support for adaptive icons with examples.

Fixes #555 and #561

### DIFF
--- a/index.html
+++ b/index.html
@@ -168,6 +168,11 @@
   "orientation": "landscape",
   "theme_color": "aliceblue",
   "background_color": "red",
+  "serviceworker": {
+    "src": "sw.js",
+    "scope": "/racer/",
+    "use_cache": false
+  },
   "screenshots": [{
     "src": "screenshots/in-game-1x.jpg",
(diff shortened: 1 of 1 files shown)


kenchris:
@owencm @anssiko @RobDolinMS Please take a look

PR #563, opened 2017-03-23
feat: add support for adaptive icons

kenchris:
Add support for adaptive icons

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/manifest/pull/563.html" title="Last updated on Dec 6, 2017, 8:58 PM GMT (5af1e02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/563/0707d4a...kenchris:5af1e02.html" title="Last updated on Dec 6, 2017, 8:58 PM GMT (5af1e02)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -303,6 +303,31 @@
           might be better suited for the space available underneath an icon).
         </p>
       </section>
+      <section data-link-for="WebAppManifest">
+        <h3>
+          Application's icon
+        </h3>
+        <p>
+          The application icon is derived from the <a>icons</a> member
+          array, depending on the size of the image resources and other meta data
+          such as the <a data-link-for="ImageResource">purpose</a> member.
+        </p>
+        <p>
+          <dfn>Adaptive icons</dfn> are icons which are prepared to support masks and
+          animations. Each icon consists of two equally sized square images.
+          A foreground image, which can be considered the main icon, and a
+          background image, which may and may not be used given the platform.
+        </p>
+        <p>
+          The inner 2/3 of the images appears within the mask. Thus, that part
+          is what makes up the final image, i.e. by default there will be a margin
+          of 1/6 of the size.
+        </p>
+        <p>
+          For special animations, each layer may move up to 1/3 of the icon size
+          in each of the directions.
+        </p>
+      </section>
       <section>
         <h3>
           Installation process
@@ -1829,6 +1854,14 @@
           next-most-appropriate icon as determined by examining the
           <a>ImageResource</a>'s members.
         </p>
+        <p class="note" data-link-for="ImageResource">
+          In order to use <a>adaptive icons</a>, you define the <a>foreground</a>
+          and <a>background</a> properties of a given <a>ImageResource</a>. These
+          are special properties only intented for icons only, and are thus ignored for
+          the <a data-link-for="WebAppManifest">screenshots</a> member.
+          A <a>src</a> can be added in order as a fallback for vendors which
+          do not support adaptive icons.
+        </p>
         <div class="example">
           <p>
             In the following example, the developer has made the following
@@ -1869,6 +1902,55 @@
                   "src": "icon/hd_hi.svg",
                   "sizes": "257x257"
                 }]
+            }
+          </pre>
+        </div>
+        <div class="example">
+          <p>
+            In the following example, the developer is defining <a>adaptive icons</a>
+            with a fallback for UA not supporting that.
+          </p>
+          <pre class="example json">
+            {
+              "icons": [
+                {
+                  "foreground": "icon/fg.png",
+                  "background": "icon/bg.png",
+                  "src": "icon/fallback.png",
+                  "sizes": "432x432",
+                  "type": "image/png"
+                }
+              ]
+            }
+          </pre>
+          <p>
+            Below the developer uses one common SVG image as background image for the
+            <a>adaptive icons</a>.
+          </p>
+          <pre class="example json">
+            {
+              "icons": [
+                {
+                  "foreground": "icon/fg@4.png",
+                  "sizes": "432x432",
+                  "type": "image/png"
+                },
+                {
+                  "foreground": "icon/fg@2.png",
+                  "sizes": "216x216",
+                  "type": "image/png"
+                },
+                {
+                  "foreground": "icon/fg.png",
+                  "sizes": "108x108",
+                  "type": "image/png"
+                },
+                {
+                  "background": "icon/shared-bg.svg",
+                  "sizes": "any",
+                  "type": "image/svg"
+                }
+              ]
             }
           </pre>
         </div>
@@ -2347,7 +2429,9 @@
       </h2>
       <pre class="idl">
         dictionary ImageResource {
-          required USVString src;
+          USVString foreground;
+          USVString background;
+          USVString src;          
           DOMString sizes;
           USVString type;
           USVString purpose;
@@ -2621,6 +2705,10 @@
           The <a>src</a> member of an <a>ImageResource</a> is a <a>URL</a> from
           which a user agent can fetch the image's data.
         </p>
+        <p class="note">
+          In the case <a>foreground</a> and <a>background</a> members are both present,
+          <a>src</a> serves as a fallback for UAs not supporting <a>adaptive icons</a>.
+        </p>
         <p>
           The steps for <dfn>processing the <code>src</code> member of an
           image</dfn> are given by the following algorithm. The algorithm takes
@@ -2640,6 +2728,54 @@
           </li>
         </ol>
       </section>
+
+      <section data-dfn-for="ImageResource" data-link-for="ImageResource">
+        <h3>
+          <dfn>foreground</dfn> and <dfn>background</dfn> member
+        </h3>
+        <p>
+          The <a>foreground</a> member and the <a>background</a> member of
+          an <a>ImageResource</a> is a <a>URL</a> from which a user agent
+          can fetch the data requires for <a>adaptive icons</a>.
+        </p>
+        <p>
+          The steps for <dfn>processing the <code>foreground</code> member of an
+          image</dfn> are given by the following algorithm. The algorithm takes
+          a <a>ImageResource</a> <var>icon</var>, and a <a>URL</a>
+          <var>manifest URL</var> , which is the <a>URL</a> from which the
+          <var>manifest</var> was fetched. This algorithm will return a
+          <a>URL</a> or <code>undefined</code>.
+        </p>
+        <ol>
+          <li>Let <var>value</var> be <var>image</var>["foreground"].
+          </li>
+          <li>If <a>Type</a>(<var>value</var>) it not <code>String</code>,
+          return <code>undefined</code>.
+          </li>
+          <li>Otherwise, <a>parse</a> <var>value</var> using <var>manifest
+          URL</var> as the base URL and return the result.
+          </li>
+        </ol>
+        <p>
+            The steps for <dfn>processing the <code>background</code> member of an
+            image</dfn> are given by the following algorithm. The algorithm takes
+            a <a>ImageResource</a> <var>icon</var>, and a <a>URL</a>
+            <var>manifest URL</var> , which is the <a>URL</a> from which the
+            <var>manifest</var> was fetched. This algorithm will return a
+            <a>URL</a> or <code>undefined</code>.
+          </p>
+          <ol>
+            <li>Let <var>value</var> be <var>image</var>["background"].
+            </li>
+            <li>If <a>Type</a>(<var>value</var>) it not <code>String</code>,
+            return <code>undefined</code>.
+            </li>
+            <li>Otherwise, <a>parse</a> <var>value</var> using <var>manifest
+            URL</var> as the base URL and return the result.
+            </li>
+          </ol>
+      </section>
+
       <section data-dfn-for="ImageResource">
         <h3>
           <dfn>type</dfn> member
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -318,6 +318,31 @@
...
         <h3>
+          Application's icon
+        </h3>
+        <p>
+          The application icon is derived from the <a><code>icons</code> member</a>
+          array, depending on the size of the image resources and other meta data
+          such as the <a>purpose member</a>.
+        </p>
+        <p>
+          <dfn>Adaptive icons</dfn> are icons which are prepared to support masks and
```
marcoscaceres: Nit: "which are prepared", change to active voice.

index.html
```diff
@@ -1799,6 +1824,14 @@ <h3 id="applying">
           next-most-appropriate icon as determined by examining the <a>image
           object's</a> members.
         </p>
+        <p class="note">
+          In order to use <a>adaptive icons</a>, you define the <code>foreground</code>
```
marcoscaceres: nit: `<a>foreground</a>` and same for background.

index.html
```diff
@@ -1799,6 +1824,14 @@ <h3 id="applying">
...
+          and <code>background</code> properties of <a>image objects</a>. These
+          are special properties only intented for icons and are thus ignored for
+          the <a><code>screenshots</code> member</a>.
+          A <code>src</code> can be added in order as a fallback for vendors which
```
marcoscaceres: "can be added in order as a fallback" doesn't read well.

index.html
```diff
@@ -1825,23 +1858,73 @@ <h3 id="applying">
...
+  ]
+}
 </pre>
         </div>
+        <div class="example">
+          <p>
+            In the following example, the developer is defining <a>adaptive icons</a>
+            with a fallback for UA not supporting that.
+          </p>
+          <pre class="">
```
marcoscaceres: class="json" (syntax highlighter will pick that up)

marcoscaceres:
(also, I didn't complete the review... just interested in who might support this right now :) )

index.html
kenchris: i.e. just drop "which are" ?

index.html
kenchris: without code?

index.html
kenchris: can be added as a fallback?

index.html
kenchris: ah great, that might be missing others places too

kenchris:
@owencm will look at this soon and has suggested/hinted about the idea before :-) The great thing is that this will also allow for creating Windows style and iOS masked icons easily (check details in #555)

marcoscaceres:
Ok, but want a firm commitment on implementer interest before we consider merging this. 

From Mozilla side, we abstain for now.

index.html
marcoscaceres: No, it should read "are icons which X prepends to support masks" - who is X? the browser, the OS, the developer? Otherwise, it's not clear who has the responsibility to do that.

index.html
marcoscaceres: Yeah, no `code`. ReSpec will automatically "code" it for you. If it doesn't, that's a bug somewhere else.

index.html
marcoscaceres: sgtm

marcoscaceres: Also, make sure the `<a>src</a>` and worst case we fix those later if the link breaks.

index.html
marcoscaceres: Yeah, we should clean that up. We should sit down together at TPAC and sped a day doing cleanup.

marcoscaceres:
Cc'ing @andreasbovens to see if we can get some traction on the Mozilla side.

kenchris:
Thanks for looking at this. I will update it when I have free cycles or when I get more feedback from owen, andreas etc

index.html
```diff
@@ -1799,6 +1824,14 @@ <h3 id="applying">
           next-most-appropriate icon as determined by examining the <a>image
           object's</a> members.
         </p>
+        <p class="note">
+          In order to use <a>adaptive icons</a>, you define the <code>foreground</code>
+          and <code>background</code> properties of <a>image objects</a>. These
+          are special properties only intented for icons and are thus ignored for
```
andreasbovens: "intended"

index.html
```diff
@@ -1825,23 +1858,73 @@ <h3 id="applying">
...
+      "background": "icon/bg.png",
+      "src": "icon/fallback.png",
+      "sizes": "432x432",
+      "type": "image/png"
+    }
+  ]
+}
+          </pre>
+          <p>
+            Below the developer uses one common SVG image as background image for the
```
andreasbovens: Comma needed after "Below".

index.html
```diff
@@ -2727,17 +2810,17 @@ <h3 id="applying">
           the image's data.
         </p>
         <p>
-          The <dfn>steps for processing the <code>src</code> member of an
-          image</dfn> are given by the following algorithm. The algorithm takes
-          a <a>image object</a> <var>icon</var>, and a <a>URL</a> <var>manifest
+          The <dfn>steps for processing an image source</dfn> are given by the
+          following algorithm. The algorithm takes a <var>member
+          name</var>, a <a>image object</a> <var>icon</var>, and a <a>URL</a> <var>manifest
```
andreasbovens: "an <a>image object</a> <var>icon</var>"

index.html
```diff
@@ -2760,6 +2843,17 @@ <h3 id="applying">
       </section>
       <section>
         <h3>
+          <code>foreground</code> and <code>background</code> member
+        </h3>
+        <p>
+          The <dfn data-lt="icon-foreground"><code>foreground</code> member</dfn> and
+          the <dfn data-lt="icon-background"><code>background</code> member</dfn> of an
+          <a>image object</a> are <a>URL</a>'s from which a user agent can fetch the
+          data requires for <a>adaptive icons</a>.
```
andreasbovens: `<a>URL</a>s`
`data required`

index.html
kenchris: are icons which follows a few set guidelines, allowing the user agent to apply masks and animations to the icons

owencm:
I haven't had a chance to do a detailed proof read of the text, but on a high level I'm supportive of adding this feature, and doing it via adding a background and foreground image alongside the "fallback" src sounds like the right approach to me.

marcoscaceres:
> I haven't had a chance to do a detailed proof read of the text, but on a high level I'm supportive of adding this feature, and doing it via adding a background and foreground image alongside the "fallback" src sounds like the right approach to me.

Thanks @owencm for the words of encouragement, tho I'm still looking for implementation commitment if we are to add this. 

@kenchris, any chance you could go back over this and address the outstanding bits of feedback (and rebase)? We can then have a solid proposal to keep shopping around.     

@andreasbovens, we should discuss this proposal next week... see if it's implementable and if we can get it onto the Fennec roadmap.

kenchris:
I am fine with rebasing the patch, though I might need to redo it after our big spec changes, but that is fine too.

I took @owencm words like a commitment to implement this, is that right?

kenchris:
OK, I rebased the PR :-)

marcoscaceres:
Something is not right. It includes a weird commit history?

kenchris:
Oh I see, what the heck happened... I followed the instructions right here from GitHib... ("or view command line instructions")

kenchris:
screenshot from 2017-12-06 21-36-45

Looks normally locally...

kenchris:
Pushed exactly the same way around and now it looks much better

marcoscaceres:
Excellent! Looks good now 🧐👍

kenchris:
I was trying to join the two commits.. but I don't see how... also, the PR still complains about conflicts!

marcoscaceres:
Everything looks ok from here, but I’ll give it a closer inspection later today

kenchris:
OK, thanks! Should be much better now! :-)

index.html
```diff
@@ -2347,7 +2429,9 @@ <h2>
       </h2>
       <pre class="idl">
         dictionary ImageResource {
-          required USVString src;
+          USVString foreground;
+          USVString background;
+          USVString src;
```
marcoscaceres: src must be required.

kenchris: That will break some cases though, like if you want to have one entry for a "any" scalable background only - in that case you don't want to require src

owencm:
Thanks both - just to clarify my statement, I'm supportive of the general direction, and think we'd like to support Adaptive Icons on Chrome (although I'm not sure of our timeline).

The general shape also looks good to me personally, but I've asked for one more LGTM from our  team to confirm that we're happy with intending to implement this specific API shape.

I hope that's helpful, sorry we've been so slow on this...

index.html
marcoscaceres: I have other use cases where we need src. 

Consider, if foreground is missing, “src” is used as foreground. If foreground is present it wins. 

If the above holds, what’s the use case for not requiring src?

kenchris: I had example like this, where background is inferred. When I discussed the feature with some devs, I got a request for that. Background is often very simple and not pixel optimized, so people like using a SVG for that.

```
+                {
 +                  "foreground": "icon/fg.png",
 +                  "sizes": "108x108",
 +                  "type": "image/png"
 +                },
 +                {
 +                  "background": "icon/shared-bg.svg",
 +                  "sizes": "any",
 +                  "type": "image/svg"
 +                }
 +              ]
```

kenchris:
Hi @owencm just wanted to hear if you have heard back from your team?

mgiuca:
Hi,

Owen just showed this to me. Sorry, I didn't know of its existence until now. @dominickng and I just had a detailed discussion about it.

Unfortunately I ran out of time to type up my thoughts on this today, and have to run now. We are supportive of doing this but there are some high-level conceptual issues, as well as syntax issues, that I think we should resolve before even considering the spec language. We should first decide what format we want the manifest to take, and what the semantics should be, *then* we can codify it in spec language. So I will avoid doing a code review on the text for now.

I will write up the details tomorrow.

mgiuca:
So there's quite an iceberg of complexity here, and unfortunately this reply is very long.

## High-level semantics

So before we even get into the Manifest syntax, we need to talk about the high-level semantics of adaptive icons.

### Separation of concerns

I think conceptually that we're dealing with two mostly unrelated issues here:
* Separating out the foreground/background layers of an icon, for the purpose of parallax or "pop" animations.
* Masking an icon, where the UA or OS crops out part of a full-bleed image to create an icon of uniform shape.

Either of these makes sense without the other, as shown in this table:

\- | Non-masking (developer-controlled shape) | Masking (OS-controlled shape)
--- | --- | ---
**Non-parallax (single image)** | Traditional icon | iOS
**Parallax (foreground + background)** | ? | Android O

Admittedly, the "non-masking + parallax" option makes less sense than the others, and we don't know of any platform that does parallax without masking, but it is still reasonable. I think it would be helpful if we think of these as two separate (but somewhat related) features.

Another thought I had is that while masking is clearly an important thing on both Android and iOS going forward, the parallax effect seems very much like a gimmick specific to Android that, while fun, is hardly noticeable (the animation almost always takes place *under your finger*) and may not be worth developing web tech for. So let's keep it in mind, but separate out these two concerns.

### Foreground without background

_"A foreground image, which can be considered the main icon, and a background image, which may and may not be used given the platform."_

So there are three possible ways for UAs to display an icon:

* a) just `src`,
* b) just `foreground`,
* c) `foreground` composited on `background`.

Basically, this (and the new examples) suggests that "`src` is the old deprecated way; a new site puts their full logo in `foreground`, with some other background in `background`, keeping `src` merely as a fallback." There are several problems with this:

* Now `foreground` has to be designed to look good by itself or on `background`.
* You can't make an icon where part of your logo is in `foreground`, and part in `background`.

Have a play with [these](https://adapticon.tooo.io/). Look how many of them play with slicing part of the logo in `foreground` and part in `background`, and consider how they would look if you just showed the foreground. Also look at [these examples](https://medium.com/google-design/designing-adaptive-icons-515af294c783).

* Google Maps would be just half the logo in `foreground`.
* Google Hangouts would just be the white quotation marks without any green.
* The Play Store logo in the latter article would just be a white square with a transparent triangle cut out of it.
* I made my own *totally not official* [adaptive Chrome logo](https://adapticon.tooo.io/#/bg=https://i.imgur.com/WbkYq4b.png/fg=https://i.imgur.com/dM1Jmzo.png) (based on [this SVG](https://vignette.wikia.nocookie.net/logopedia/images/c/ca/Google_Chrome_for_Android_Icon_2016.svg/revision/latest?cb=20160305053723)). The `foreground` is just the blue circle, which doesn't form a logo on its own.

Basically, we can't ask developers to design a `foreground` + `background` logo for masking, and then allow UAs to optionally strip out the background. We should require that UAs either use `foreground` and `background` together, or use `src` and ignore the adaptive icon.

### Masking and proportions

This is a huge source of complexity (and confusion), and I'll note up front that we could reduce a lot of the complexity here if we had masking, but not parallax.

_"The inner 2/3 of the images appears within the mask. Thus, that part is what makes up the final image, i.e. by default there will be a margin of 1/6 of the size. For special animations, each layer may move up to 1/3 of the icon size in each of the directions."_

This isn't nearly detailed enough, and also it's straight out of the Android Oreo spec. I'm wary of committing to the Android proportions, because they may not match how other platforms treat icons, and also Android may change in the future.

For some reason the official Android docs don't tell you this, but [this Medium article](https://medium.com/google-design/designing-adaptive-icons-515af294c783) does: the icon space has two key zones: the "dead zone" and the "safe zone".

* The "dead zone" is a square outer border of 1/6 of the icon size on each side. All pixels in this zone are guaranteed to *never* be seen in any mask, except when animating/parallaxing. The inverse of the dead zone (the pixels that *may* be seen) is a square of 2/3 of the width and height of the icon.
* The "safe zone" is a circle of 11/36 of the icon size in radius. All pixels in this zone are guaranteed to be seen in all masks.

Safe zone within a rounded square mask

We need to spec something similar, specifying both the dead and safe areas. This allows developers to design the full-square version of their logo within the dead zone (making sure any important parts are also within the safe zone), and then extending the edges of the image outwards to cover the dead zone. That's what I did with my Chrome logo experiment above:

Background:
Adaptive Chrome logo, background
Foreground:
Adaptive Chrome logo, foreground

We would also need to spec the "default" representation of an icon (i.e., what the icon looks like without parallax) as an algorithm, something like this:

> 1. Start with the background image.
> 2. Composite the foreground image onto the background.
> 3. Crop out 1/6 of the image size on each side (the dead zone).
> 4. You MAY apply a mask of any size, making any pixels that are more than 11/36ths of the image size away from the center (the safe zone) transparent. You MUST NOT make any pixel within the safe zone transparent. It is RECOMMENDED that you make all such pixels transparent (i.e., a circle).

The big unknown for me is, how do we specify the dead zone in a way that isn't locked to Android Oreo's specifications? What if a future platform comes along where the dead zone is 1/2 the width, instead of 1/3? We have to decide on a particular number for the dead zone. We can accommodate any platform with a smaller dead zone than what we choose, by cropping. We can't accommodate a platform with a bigger dead zone. Oreo has a fairly big dead zone, so we could just go with 1/3, but it is arbitrary and the only reason we chose it is because of Oreo.

If we were *only* doing maskable icons (not parallax), then not only would we not need separate background/foreground images, but we also wouldn't need to spec dead zone, only safe zone. This would remove steps 2 and 3 above, and avoid the arbitrary decision about how big to make the dead zone. Then the Chrome logo would just be this:

Maskable Chrome logo

Which frankly is much easier for developers to understand, and we wouldn't need to have 55% of the image cropped out in almost all cases.

## Manifest syntax

OK onto the actual details of the Manifest.

Dom and I discussed this and we're not really fans of just shoving the new adaptive icons alongside the existing `src`. For one thing, it means each adaptive ImageResource needs a corresponding `src`. What if you supply an adaptive ImageResource at a different size to the rest of your traditional images? What if you want to supply an SVG ImageResource (with "sizes: any") while your traditional images are PNGs (with multiple resources of different, fixed size?)

It's weird for there to be a "type" member for ImageResource, when an ImageResource can now contain three different images, potentially with different types.

Further, I can see what you're trying to accomplish with "steps for inferring background", but it seems quite arbitrary to always bundle the foreground and background in the same ImageResource, *except* in this special case, where you provide a separate scalable background, with a magic algorithm that associated a scalable background with all of the other images that have no background. Does that special IR with the scalable background require a scalable "src"? The algorithm totally ignores "type", "purpose" and "platform" of the scalable background when it copies it into the fixed-size ImageResources.

Basically, I think one ImageResource should represent one image file. The little-used "purpose" member should be used to indicate adaptive icons.

We propose that instead of adding "foreground" and "background" members, we just add new "purposes" of "foreground" and "background", using the "src" member. This means if you want an adaptive icon, you specify three ImageResources (x the number of sizes you need):

```json
{
  "icons": [
    {
      "foreground": "icon/fg.png",
      "background": "icon/bg.png",
      "src": "icon/fallback.png",
      "sizes": "432x432",
      "type": "image/png"
    }
  ]
}
```
becomes:
```json
{
  "icons": [
    {
      "src": "icon/fallback.png",
      "sizes": "432x432",
      "type": "image/png"
    },
    {
      "purpose": "foreground",
      "src": "icon/fg.png",
      "sizes": "432x432",
      "type": "image/png"
    },
    {
      "purpose": "background",
      "src": "icon/bg.png",
      "sizes": "432x432",
      "type": "image/png"
    }
  ]
}
```

You would still be required to supply a non-purposed icon as the "main" icon, and then optionally specify an alternate "adaptive" icon in the form of a foreground and background purpose. Both of these are optional, but if one is missing, it would just be equivalent to fully transparent. (So, you can make a non-parallaxing but still full-bleed-and-masked icon by supplying just a background.)

This also means you can mix and match sizes arbitrarily. You could supply just a high-res background and have it scaled by the UA, while the foreground is specified in 10 different sizes. Or you could supply an SVG foreground with many fixed-size backgrounds. So you get your "steps for inferring background" for free.

kenchris:
I am OK with using purpose instead, that is probably better. Do you feel like trying to write the spec text for this? You are of course free to re-use parts of this PR, if that makes any sense.

I understand you point with dead-zone, but I am not sure there is any usefulness in any dead-zone like 1/2 - it just doesn't make practical sense, so it is maybe a bit of an invented/theoretical problem.

You can say that 1/3 is because of Oreo, but it is also a very nicely (beautiful?) chosen number. It is not like 2/5 or 6/10 :-) It doesn't feel that arbitrary and I wouldn't be surprised if others would go with something similar like 1/3 or 1/4 - so with 1/3 we should be covered well.

kenchris:
OK, let's try masking for now then, https://github.com/w3c/manifest/pull/657

clagom:
Hello everyone, I’m a co-worker of @kenchris 
I had looked at the Android adaptive icons while ago, and the official guidelines are very useful and clear. I didn’t know about the “safe zone” area mentioned in the Medium article, but it’s a nice detail to know. I wonder though if it overcomplicates the guidelines if included in the specs with the proportion "11/36". I think providing the image itself as an example would be a simpler explanation, also for designers.
This "safe zone" is not included in the Android guidelines, so could probably be simplified mentioning a ‘1/3’ instead (bigger than the safe area, but still within the margins)?

mgiuca:
Hi Ken,

This has become somewhat urgent since Chrome 69 seems to be upgrading to the Android O SDK and suddenly all the web app icons are going inside white circles (https://crbug.com/862041). It would be good if we can let web apps specify Adaptive icons soon.

Do you think we can start with _just_ the masking (not parallax) so the proposal basically just becomes:

- Introduce a new `"purpose": "maskable"` type of icon.
- Give it semantics of: not supposed to have an alpha channel; define a "safe zone" (such as a circle of some radius around the center) of pixels that *must* be shown, and say that the user agent may mask off any other pixels outside of the safe zone.
- Optional; a non-maskable icon is still mandatory. And it's completely optional for user agents to use the maskable icons.

No foreground/background. No "dead zone". Should be fairly straightforward. I think we can commit to implementing this in Chrome.

Do you have time to work on this proposal? Otherwise, I can draft one up.

kenchris:
Great, I can work on this

kenchris:
I actually have an existing PR for this (need to check if I need to update it)

https://github.com/w3c/manifest/pull/657

kenchris:
We merged a derivative of this PR, so closing

PR #566, opened 2017-04-12
Add min_version and fingerprints to application objects

### LINKED ISSUE #564: related_applications needs fingerprints array

domenic:
Closes #564. Closes https://github.com/w3c/webpayments/issues/225.

/cc @mgiuca @zkoch @rsolomakhin

### DIFF
--- a/index.html
+++ b/index.html
@@ -2295,10 +2295,20 @@ Content-Type: application/manifest+json
                   <code>undefined</code>, move onto the next item if any are
                   left.
                   </li>
-                  <li>Otherwise, let <var>application</var> be an object with
-                  properties <code>platform</code>, <code>id</code>,
-                  <code>url</code> respectively set to <var>platform</var>,
-                  <var>id</var> and <var>url</var>.
+                  <li>Let <var>minVersion</var> be the result of running the
+                  <a>steps for processing the <code>min_version</code> member
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>fingerprints</var> be the result of running the
+                  <a>steps for processing the <code>fingerprints</code> member
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>application</var> be an object with properties
+                  <code>platform</code>, <code>id</code>, <code>url</code>,
+                  <code>min_version</code>, and <code>fingerprints</code>
+                  respectively set to <var>platform</var>, <var>id</var>, <var>
+                    url</var>, <var>minVersion</var>, and
+                    <var>fingerprints</var>.
                   </li>
                   <li>Append <var>application</var> to <var>applications</var>.
                   </li>
@@ -3141,11 +3151,44 @@ Content-Security-Policy: img-src icons.example.com
       </h2>
       <p>
         Each <dfn>application object</dfn> represents an application related to
-        the web application. An application object has three properties: a
-        <code>platform</code> which represents the platform it is associated
-        to, a <code>url</code> which represents the URL where the application
-        can be found and an <code>id</code> which can be used as an information
-        additional to the URL or instead of the URL, depending on the platform.
+        the web application. An application object has five properties:
+      </p>
+      <dl>
+        <dt>
+          <code>platform</code>
+        </dt>
+        <dd>
+          the platform it is associated to
+        </dd>
+        <dt>
+          <code>url</code>
+        </dt>
+        <dd>
+          the URL where the application can be found
+        </dd>
+        <dt>
+          <code>id</code>
+        </dt>
+        <dd>
+          information additional to the URL or instead of the URL, depending on
+          the platform
+        </dd>
+        <dt>
+          <code>min_version</code>
+        </dt>
+        <dd>
+          information about the minimum version of an application required for
+          certain scenarios
+        </dd>
+        <dt>
+          <code>fingerprints</code>
+        </dt>
+        <dd>
+          an array of <a data-lt="fingerprint object">fingerprint objects</a>
+          used for verifying the application
+        </dd>
+      </dl>
+      <p>
         A valid <a>application object</a> MUST have <code>platform</code> and
         either an <code>url</code> or an <code>id</code> (or both).
       </p>
@@ -3153,7 +3196,9 @@ Content-Security-Policy: img-src icons.example.com
         <p>
           In the following example, the web application is listing two
           different related applications, one on Google Play Store and the
-          other one on the iTunes Store:
+          other one on the iTunes Store. The one on the Google Play Store has a
+          minimum version requirement and cryptographic fingerprints used for
+          verification, in a Play-Store-specific manner.
         </p>
         <pre class="">
 {
@@ -3161,7 +3206,12 @@ Content-Security-Policy: img-src icons.example.com
       {
         "platform": "play",
         "url": "https://play.google.com/store/apps/details?id=com.example.app1",
-        "id": "com.example.app1"
+        "id": "com.example.app1",
+        "min_version": "2",
+        "fingerprints": [{
+          "type": "sha256_cert",
+          "value": "92:5A:39:05:C5:B9:EA:BC:71:48:5F:F2"
+        }]
       }, {
         "platform": "itunes",
         "url": "https://itunes.apple.com/app/example-app1/id123456789",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3141,27 +3151,67 @@ <h3 id="applying">
       </h2>
       <p>
         Each <dfn>application object</dfn> represents an application related to
-        the web application. An application object has three properties: a
-        <code>platform</code> which represents the platform it is associated
-        to, a <code>url</code> which represents the URL where the application
-        can be found and an <code>id</code> which can be used as an information
-        additional to the URL or instead of the URL, depending on the platform.
+        the web application. An application object has five properties:
```
mgiuca: Can we change this to "has the following properties" so we don't have to remember to keep this number up to date as we add more?

index.html
```diff
@@ -3141,27 +3151,67 @@ <h3 id="applying">
...
+      <p>
         A valid <a>application object</a> MUST have <code>platform</code> and
         either an <code>url</code> or an <code>id</code> (or both).
       </p>
       <div class="example">
         <p>
           In the following example, the web application is listing two
           different related applications, one on Google Play Store and the
-          other one on the iTunes Store:
+          other one on the iTunes Store. The one on the Google Play Store has a
```
mgiuca: "... has an Android package name, a minimum version ..."

index.html
```diff
@@ -3244,6 +3294,108 @@ <h3 id="applying">
           </li>
         </ol>
       </section>
+      <section>
+        <h3>
+          <code>min_version</code> member
+        </h3>
+        <p>
+          The <dfn data-lt="application-min-version">min_version</dfn> member
+          of an application object represents the minimum version required for
```
mgiuca: _"minimum version required"_

"required" is a bit weird here since referring to an application in the manifest does not "require" anything. Perhaps, "the minimum version of the application that is considered related to this web app."

index.html
```diff
@@ -3244,6 +3294,108 @@ <h3 id="applying">
...
+        <p>
+          The <dfn data-lt="application-fingerprints">fingerprints</dfn> member
+          of an application object represents a set of cryptographic
+          fingerprints used for verifying the application. It is given as an
+          array of <a data-lt="fingerprint object">fingerprint objects</a>.
+        </p>
+        <p>
+          A <dfn>fingerprint object</dfn> has two properties: <code>type</code>
+          and <code>value</code>. Each of these are strings, but their
+          semantics are platform-defined.
```
mgiuca: syntax and semantics

index.html
```diff
@@ -3244,6 +3294,108 @@ <h3 id="applying">
...
+          </li>
+          <li>If <var>unprocessed fingerprints</var> is an array, then for each
+          <var>potential fingerprint</var> in <var>unprocessed
+          fingerprints</var>:
+            <ol>
+              <li>Let <var>type</var> be the result of calling the
+              <a>[[\GetOwnProperty]]</a> internal method of <var>potential
+              fingerprint</var> passing "<code>type</code>" as an argument.
+              </li>
+              <li>If Type(<var>type</var>) is not "string", then issue a
```
mgiuca: Oh no :(

The type/type name clash has already come to bite us.

index.html
```diff
@@ -3244,6 +3294,108 @@ <h3 id="applying">
...
+              developer warning that the type is unsupported, and move on to
+              the next item in <var>unprocessed fingerprints</var> (if any are
+              left).
+              </li>
+              <li>Let <var>value</var> be the result of calling the
+              <a>[[\GetOwnProperty]]</a> internal method of <var>potential
+              fingerprint</var> passing "<code>value</code>" as an argument.
+              </li>
+              <li>If Type(<var>value</var>) is not "string", then issue a
+              developer warning that the type is unsupported, and move on to
```
mgiuca: Same as above.

index.html
```diff
@@ -3244,6 +3294,108 @@ <h3 id="applying">
...
+          <li>If <var>unprocessed fingerprints</var> is an array, then for each
+          <var>potential fingerprint</var> in <var>unprocessed
+          fingerprints</var>:
+            <ol>
+              <li>Let <var>type</var> be the result of calling the
+              <a>[[\GetOwnProperty]]</a> internal method of <var>potential
+              fingerprint</var> passing "<code>type</code>" as an argument.
+              </li>
+              <li>If Type(<var>type</var>) is not "string", then issue a
+              developer warning that the type is unsupported, and move on to
```
mgiuca: "issue a developer warning" should be linked.

I noticed that the other uses of this include the word "optionally" (see id and min_version). However, the definition of "issue a developer warning" clearly says that the UA can decide to do nothing, so the word "optionally" is redundant here. Perhaps remove it from the other usages?

index.html
```diff
@@ -2295,10 +2295,20 @@ <h3 id="applying">
...
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>fingerprints</var> be the result of running the
+                  <a>steps for processing the <code>fingerprints</code> member
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>application</var> be an object with properties
+                  <code>platform</code>, <code>id</code>, <code>url</code>,
+                  <code>min_version</code>, and <code>fingerprints</code>
+                  respectively set to <var>platform</var>, <var>id</var>, <var>
```
mgiuca: nit: Wrap so that <var>url</var> is together on one line.

Don't indent these two lines by extra spaces.

domenic: The formatting is generated by htmltidy, so I don't have any control over that.

domenic:
@mgiuca updated, PTAL!

index.html
```diff
@@ -1563,9 +1562,8 @@ <h3 id="applying">
           </li>
           <li>If <a>Type</a>(<var>value</var>) is not "string":
             <ol>
-              <li>If <a>Type</a>(<var>value</var>) is not "undefined",
-              optionally <a>issue a developer warning</a> that the type is not
-              supported.
+              <li>If <a>Type</a>(<var>value</var>) is not "undefined", <a>issue
+              a developer warning</a> that the type is not supported.
```
kenchris: Maybe we should use unsupported instead of not supported

index.html
```diff
@@ -2295,10 +2289,20 @@ <h3 id="applying">
...
+                  <a>steps for processing the <code>min_version</code> member
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>fingerprints</var> be the result of running the
+                  <a>steps for processing the <code>fingerprints</code> member
+                  of an application</a> with <var>potential application</var>.
+                  </li>
+                  <li>Let <var>application</var> be an object with properties
+                  <code>platform</code>, <code>id</code>, <code>url</code>,
+                  <code>min_version</code>, and <code>fingerprints</code>
```
kenchris: , before respectively?

index.html
```diff
@@ -3141,27 +3142,69 @@ <h3 id="applying">
...
-        additional to the URL or instead of the URL, depending on the platform.
+        the web application. An application object has the following
+        properties:
+      </p>
+      <dl>
+        <dt>
+          <code>platform</code>
+        </dt>
+        <dd>
+          the platform it is associated to
```
kenchris: to, not with?

index.html
```diff
@@ -3244,6 +3285,107 @@ <h3 id="applying">
...
+              developer warning</a> that the type is unsupported, and move on
+              to the next item in <var>unprocessed fingerprints</var> (if any
+              are left).
+              </li>
+              <li>Let <var>value</var> be the result of calling the
+              <a>[[\GetOwnProperty]]</a> internal method of <var>potential
+              fingerprint</var> passing "<code>value</code>" as an argument.
+              </li>
+              <li>If Type(<var>value</var>) is not "string", then <a>issue a
+              developer warning</a> that the type is unsupported, and move on
```
kenchris: Here we use unsupported instead of 'not supported'

kenchris:
LGTM

index.html
domenic: The current document appears to use "not supported" 14 times and "unsupported" 15 times. I'll switch the ones I introduce to "unsupported" but leave the cleanup for others.

domenic: Actually this is not even in my code. This is just stuff I cleaned up as requested in an earlier code review to remove "optionally". So I'd prefer to not mess with this at all.

index.html
domenic: That seems less correct to me.

index.html
domenic: This was the existing text; I just moved it around.

kenchris:
OK I am fine with merging this, if you don't have any other changes planned

marcoscaceres:
I'm on vacation for the next week - so will have to review then. But as @kenchris approves, please merge.

index.html
mgiuca: FYI I think the normal way of writing it is "x, y and z, respectively, set to..." (comma before and after respectively). But it doesn't matter.

PR #567, opened 2017-04-18
feat: add iarc_rating_id member (resolves #523)

### LINKED ISSUE #523: Proposal: Add Optional Member for IARC Rating
It may be useful for web apps to be able to optionally indicate their [IARC](https://www.globalratings.com/) rating.  I'm thinking this may be as simple as: 

`"iarc_rating": "opaque_string",`

App stores participating in IARC include the Google Play Store, Windows Store, and Nintendo Store.

Locales covered by IARC include: Australia, Brazil, Germany, Europe (PEGI), and North America (ESRB)

For more on IARC:
* https://www.globalratings.com/intro-to-iarc-en.aspx
* https://www.globalratings.com/for-developers.aspx

RobDolinMS:
See discussion in Issue #523

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/manifest/RobDolinMS-iarc-rating_id.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/a34e938...4ed3f16.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1339,6 +1339,11 @@ Content-Type: application/manifest+json
           <code>background_color</code> member</a> with <var>manifest</var> as
           argument.
           </li>
+          <li>Let <var>iarc_rating_id</var> of <var>parsed manifest</var> be
+          the result of running the <a>steps for processing the
+          <code>iarc_rating_id</code> member</a> with <var>manifest</var> as
+          argument.
+          </li>
           <li>Let <var>screenshots</var> of <var>parsed manifest</var> be the
           result of running the <a>steps for processing an array of images</a>
           with <var>manifest</var>, <var>manifest URL</var>, and "screenshots"
--- a/index.html
+++ b/index.html
@@ -2929,6 +2929,13 @@ Content-Security-Policy: img-src icons.example.com
           <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
           </li>
         </ol>
+        <p>
+          More information on the <a href="https://www.globalratings.com/">
+          International Age Rating Coalition (IARC)</a> can be found at: 
+          <a href="https://www.globalratings.com/how-iarc-works.aspx">
+          https://www.globalratings.com/how-iarc-works.aspx<a/> and 
+          <a href="https://www.globalratings.com/for-developers.aspx">
+          https://www.globalratings.com/for-developers.aspx</a>.
       </section>
     <section>
       <h2>
--- a/index.html
+++ b/index.html
@@ -2891,6 +2891,40 @@ Content-Security-Policy: img-src icons.example.com
         </ol>
       </section>
     </section>
+      <section>
+        <h3>
+          <code title="">iarc_rating_id</code> member
+        </h3>
+        <p>
+          The <dfn id="member-iarc_rating_id"><code>iarc_rating_id</code> member</dfn>
+          is a <a>string</a> that represents an ID value of the IARC rating of the 
+          web application. It is intended to be used to determine which ages 
+          the web application is appropriate for.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>iarc_rating_id</code>
+          member</dfn> is given by the following algorithm. The algorithm takes
+          a <var>manifest</var> as an argument. This algorithm returns a string
+          or <code>undefined</code>.
+        </p>
+        <ol>
+          <li>Let <var>value</var> be the result of calling the
+          <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
+          with argument "<code>iarc_rating_id</code>".
+          </li>
+          <li>If <a>Type</a>(<var>value</var>) is not "string":
+            <ol>
+              <li>If <a>Type</a>(<var>value</var>) is not "undefined", <a>issue
+              a developer warning</a> that the type is not supported.
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
+          </li>
+        </ol>
+      </section>
     <section>
       <h2>
         Multi-purpose members
--- a/index.html
+++ b/index.html
@@ -2929,7 +2929,7 @@ Content-Security-Policy: img-src icons.example.com
           <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
           </li>
         </ol>
-        <section class="informative">
+        <div class="informative">
         <p>
           The following shows a very simple <a>manifest</a> with the 
           <code>iarc_rating_id</code> member.
@@ -2945,7 +2945,7 @@ Content-Security-Policy: img-src icons.example.com
           }]
         }
         </pre>
-        </section>
+        </div>
 
         <p>
           More information on the <a href="https://www.globalratings.com/">
--- a/index.html
+++ b/index.html
@@ -2929,6 +2929,24 @@ Content-Security-Policy: img-src icons.example.com
           <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
           </li>
         </ol>
+        <section class="informative">
+        <p>
+          The following shows a very simple <a>manifest</a> with the 
+          <code>iarc_rating_id</code> member.
+        </p>
+        <pre class="example json" title="very simple manifest">
+        {
+          "name": "Donate App",
+          "description": "This app helps you donate to worthy causes.",
+          "iarc_rating_id": "e84b072d-71b3-4d3e-86ae-31a8ce4e53b7",
+          "icons": [{
+            "src": "images/icon.png",
+            "sizes": "192x192"
+          }]
+        }
+        </pre>
+        </section>
+
         <p>
           More information on the <a href="https://www.globalratings.com/">
           International Age Rating Coalition (IARC)</a> can be found at: 
--- a/index.html
+++ b/index.html
@@ -2936,6 +2936,7 @@ Content-Security-Policy: img-src icons.example.com
           https://www.globalratings.com/how-iarc-works.aspx<a/> and 
           <a href="https://www.globalratings.com/for-developers.aspx">
           https://www.globalratings.com/for-developers.aspx</a>.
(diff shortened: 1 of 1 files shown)


marcoscaceres:
(will review mid next week, on vacation till then)

RobDolinMS:
I think this may need a line in the [Processing](http://www.w3.org/TR/appmanifest/#processing) section; and it might be good to add this to the [Examples](http://www.w3.org/TR/appmanifest/#example-manifests) section as well.  

@marcoscaceres @kenchris @mounirlamouri @anssiko Are there other places where I should add this?

marcoscaceres:
> I think this may need a line in the Processing section;

Correct. 

> and it might be good to add this to the Examples section as well.

I'm not inclined to add it there as it's not something that has received backing from other implementers - thus I don't think we want developers to worry too much about it until it becomes more widely supported by search engines in general. 

Instead, I'd suggest just adding an example in the section you have defined for now.

index.html
```diff
@@ -2891,6 +2891,40 @@ <h3 id="applying">
         </ol>
       </section>
     </section>
+      <section>
+        <h3>
+          <code title="">iarc_rating_id</code> member
+        </h3>
+        <p>
+          The <dfn id="member-iarc_rating_id"><code>iarc_rating_id</code> member</dfn>
+          is a <a>string</a> that represents an ID value of the IARC rating of the
```
kenchris: That would be great, but when I looked I am not sure these things are actually public, so maybe we could create some examples in the spec instead or some guidance? ie a Note?

index.html
```diff
@@ -2891,6 +2891,40 @@ <h3 id="applying">
...
     </section>
+      <section>
+        <h3>
+          <code title="">iarc_rating_id</code> member
+        </h3>
+        <p>
+          The <dfn id="member-iarc_rating_id"><code>iarc_rating_id</code> member</dfn>
+          is a <a>string</a> that represents an ID value of the IARC rating of the 
+          web application. It is intended to be used to determine which ages 
+          the web application is appropriate for.
```
kenchris: should we add that this is global thing, ie not a country specific thing... it maybe a bit of info in a note about IARC

kenchris: Maybe the note, could even have a small pictures showing how it could look

index.html
marcoscaceres: If it's not public, then I have ... issues :(

kenchris: It is not like it is hidden, but their sites it just not very precise at explaining things :-) and you will have to go through a ratings authority anyway, so they will probably be the ones ending up giving you the string - but I cannot find docs explaining how that string looks

https://www.globalratings.com/for-developers.aspx

kenchris: https://www.globalratings.com/how-iarc-works.aspx

marcoscaceres:
❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️
❤️🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥❤️
❤️SPOTTED THE `iarc_rating_id` MEMBER AT MICROSOFT BUILD❤️
❤️🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥🔥❤️
❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️❤️

(well, maybe... in the screenshot I saw of a PWA in the Microsoft store! ... super exciting!)

index.html
RobDolinMS: Good suggestions @kenchris. I've updated the PR with an additional paragraph linking to these.

index.html
RobDolinMS: I've spelled-out "International Age Rating Coalition (IARC)" in the last paragraph so hopefully "International" makes it clear that this is global.

index.html
```diff
@@ -2891,6 +2896,66 @@ <h3 id="applying">
         </ol>
       </section>
     </section>
+      <section>
```
marcoscaceres: This section should be part of "8. Manifest and its members". Not it's own section.

marcoscaceres: (sorry I mean, a sub-section of "8. Manifest and its members")

index.html
```diff
@@ -2891,6 +2896,66 @@ <h3 id="applying">
...
+              a developer warning</a> that the type is not supported.
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
+          </li>
+        </ol>
+        <section class="informative">
```
marcoscaceres: Probably just change this to a `div` and drop the `class=`.

index.html
```diff
@@ -2891,6 +2896,66 @@ <h3 id="applying">
...
+          "iarc_rating_id": "e84b072d-71b3-4d3e-86ae-31a8ce4e53b7",
+          "icons": [{
+            "src": "images/icon.png",
+            "sizes": "192x192"
+          }]
+        }
+        </pre>
+        </section>
+
+        <p>
```
marcoscaceres: this should have a `class="note"`

index.html
```diff
@@ -2891,6 +2896,66 @@ <h3 id="applying">
         </ol>
       </section>
     </section>
+      <section>
+        <h3>
+          <code title="">iarc_rating_id</code> member
```
marcoscaceres: Looking at the cert you sent us, I wonder if we should actually rename this "iarc_certificate_id"? It's not a rating id, it's actually the certificate UUID that is unique for this particular application, no? Or did I misunderstand.

marcoscaceres: Also, Also, what happens if I list this in another store? The certificate only seems to apply to one store. 

Maybe it should be:
```JSON
"iarc_certs": [
 {"id": "E84b072d-71b3-4d3e-86ae-31a8ce4e53b7", "store": "Microsoft"},
 {"id": "12dasfdf-71b3-4d3e-86ae-adsfafasfasd", "store": "Mozilla"},
]
 ```

Comment (marcoscaceres) [inline index.html]: Alternatively, can an "id" be used on multiple stores?

**Comment by marcoscaceres** [comment]
@RobDolinMS, we should have a call with IARC and get their blessing that we've captured the right information in the optimal way.

File: index.html (hunk 6 above)
Comment (kenchris) [inline index.html]: I was wondering about that "store" part as well. I though it was supposed to work across stores.

We already call stores platform elsewhere like "play" - so we should reuse the same names

Comment (marcoscaceres) [inline index.html]: >  I though it was supposed to work across stores.

If it does, then we are golden.

Comment (marcoscaceres) [inline index.html]: > Also if MS plans to show these in Edge at some point (say with a prompt) we need to find out what to do in that case. Talk to IARC?

We might need to define how to actually do that. This means pinging the IARC servers for validation, no? Is it just a GET request? Do we need any special HTTP headers? What happens if you get a 404? etc.

Comment (RobDolinMS) [inline index.html]: >I @kenchris was wondering about that "store" part as well. I though it was supposed to work across stores.

AFAIK, the IARC ID is per-app and can be used across stores.  

During the process of submitting an app to a store, an app developer has a choice of entering their IARC ID or getting a new one.

>This means pinging the IARC servers for validation, no? 
> Is it just a GET request? Do we need any special HTTP headers?

In the confirmation email there is a permalink: 
>A permanent record of this certificate may be accessed here.
"here" is hyperlinked.

**Comment by RobDolinMS** [comment]
@marcoscaceres @kenchris I believe my [comments](https://github.com/w3c/manifest/pull/567#discussion_r120516925) and [edits](https://github.com/w3c/manifest/pull/567/commits/4ed3f166ee79aa1ffcf80c8a764c6a9e28ab59a6) resolve remaining questions.  

Please let me know if there's anything missing

File: index.html (hunk 3 above)
Comment (marcoscaceres) [inline index.html]: @RobDolinMS, this is still unaddressed. Should move up to Section 8.

File: index.html (hunk 6 above)
Comment (marcoscaceres) [inline index.html]: Thanks for this info. 👍

**Comment by marcoscaceres** [comment]
@kenchris, for the record, can you please check that all your comments are addressed and approve the PR if so.

**Comment by marcoscaceres** [comment]
(don't merge tho 😇)

**Comment by kenchris** [comment]
Ok merging! :smiling_imp: ... :-) NOT! :)

File: index.html [hunk 7]
```diff
@@ -2891,6 +2896,66 @@ <h3 id="applying">
...
+          <li>If <a>Type</a>(<var>value</var>) is not "string":
+            <ol>
+              <li>If <a>Type</a>(<var>value</var>) is not "undefined", <a>issue
+              a developer warning</a> that the type is not supported.
+              </li>
+              <li>Return <code>undefined</code>.
+              </li>
+            </ol>
+          </li>
+          <li>Otherwise, <a>Trim</a>(<var>value</var>) and return the result.
```
kenchris: The ratings seems to always be lowercase... should we allow both and just apply lowercase here?

marcoscaceres: No, let's allow both. Prefer not to fiddle with the value as it could change the meaning.

RobDolinMS:
[Previously](https://github.com/w3c/manifest/pull/567#issuecomment-306385574), @marcoscaceres wrote:
>@RobDolinMS, we should have a call with IARC and get their blessing that we've captured the right information in the optimal way.

One of my colleagues who is involved in IARC has reached-out and gotten a thumbs-up:

>From: [Vance, Patricia](https://www.esrb.org/about/vance_bio.aspx)
>Sent: Thursday, August 3, 2017 3:30 PM
>
>Sounds awesome!  Thanks, …please proceed!
>
>P

index.html
RobDolinMS: @marcoscaceres or @kenchris  If everything else looks good beside the section nesting, would you want to fix the section nesting when merging?  THANK YOU for helping to land this.

marcoscaceres:
@RobDolinMS, thanks for doing the due diligence with IARC. I'll try to make the final markup changes and merge this in.

marcoscaceres:
Merging this in and will do the cleanup locally.

PR #570, opened 2017-05-02
Use UTF-8 decode when obtaining a manifest

### LINKED ISSUE #565: It is not stated how to decode the bytes obtained
https://www.w3.org/TR/appmanifest/#obtaining assumes the body is a string, but it is not. It is a stream of bytes. You need to decode it, preferably using UTF-8 decode.

This is important as people might be assuming that you e.g. respect the Content-Type header or do some sort of sniffing.

For an example of how to do this see https://domenic.github.io/payment-method-manifest/#processing-model

anssiko:
Fixes #565.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1130,11 +1130,14 @@ if (standalone.matches) {
           <li>If <var>response</var> is a <a>network error</a>, terminate this
           algorithm.
           </li>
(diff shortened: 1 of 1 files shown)


anssiko:
Thanks for the swift review @domenic and @kenchris.

Issue #573, opened 2017-05-03
Use data-cite for all references


anssiko:
Use https://github.com/w3c/respec/wiki/data--cite for references to allow them to be navigated to directly.

(This is an editorial thing.)

kenchris:
Basically done, expect for Service Workers because there is no ref for the editors draft @marcoscaceres

marcoscaceres:
Great! thanks again!

PR #574, opened 2017-05-03
Use data-cite

anssiko:
Fixes #570.

@marcoscaceres, I do use data-cite for other specs extensively. I followed the convention of this spec of not using it initially :-)

That said, we should move all the references to use data-cite, so I opened #573.


anssiko:
Editorial: merging without review.

PR #584, opened 2017-06-07
Changed to allow site-triggered prompt without defaulting to automatic prompt

### LINKED ISSUE #576: Install prompts: Allow manual prompting but not automatic prompting
Implementations may wish to show a prompt if, and only if, the site explicitly requests it via `BeforeInstallPromptEvent.prompt()`, but not automatically without the site's approval. Is this something we want to leave to the discretion of the user agent? (This would require changing the language in §2.5 "Install prompts": "steps to notify before an automated install prompt" would become "steps to notify that an install prompt is available".)

mgiuca:
Closes #576

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/manifest/allow-only-manual-prompts.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/2f5edbf...mgiuca:d1eeae7.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -658,10 +658,10 @@
           </h4>
           <p>
             This example shows how one might prevent an automated install
-            prompt from showing until the user clicks a button to install the
-            app. In this way, the site can leave installation at the user's
-            discretion (rather than prompting at an arbitrary time), whilst
-            still providing a prominent UI to do so.
+            prompt from showing until the user clicks a button to show a
+            site-triggered install prompt. In this way, the site can leave
+            installation at the user's discretion (rather than prompting at an
+            arbitrary time), whilst still providing a prominent UI to do so.
           </p>
           <pre class="example" title=
           "Using beforeinstallprompt to present an install button">
--- a/index.html
+++ b/index.html
@@ -3637,10 +3637,6 @@ Content-Security-Policy: img-src icons.example.com
       </p>
       <ul>
         <li>
-          <a href="https://dom.spec.whatwg.org/#canceled-flag"><dfn>canceled
-          flag</dfn></a>
-        </li>
-        <li>
           <a href="https://dom.spec.whatwg.org/#event"><dfn><code>Event</code>
           interface</dfn></a>
         </li>
--- a/index.html
+++ b/index.html
@@ -396,7 +396,8 @@
           Install prompts
         </h2>
         <p>
-          There are three ways that the installation process can be triggered:
+          There are multiple ways that the installation process can be
+          triggered:
         </p>
         <ul>
           <li>An end-user can <dfn data-lt="manual installation">manually</dfn>
@@ -412,8 +413,8 @@
           <li>The <a>installation process</a> can occur through a
           <dfn>site-triggered install prompt</dfn>: the site can
           programmatically request that the user agent present an install
-          prompt to the user. The availability of this feature may be
-          restricted to when, for instance, there are sufficient
+          prompt to the user. The user agent MAY restrict the availability of
+          this feature to when, for instance, there are sufficient
           <a>installability signals</a> to warrant <a>installation</a> of the
           web application.
           </li>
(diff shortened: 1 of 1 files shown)


mgiuca:
Preview: https://rawgit.com/mgiuca/manifest/allow-only-manual-prompts/index.html

index.html
```diff
@@ -396,21 +396,36 @@
           Install prompts
         </h2>
         <p>
-          An end-user can <dfn data-lt="manual installation">manually</dfn>
+          There are three ways that the installation process can be triggered:
```
marcoscaceres: maybe change this to "multiple ways" instead of "three"... always forget to update those numbers when new things get added.

index.html
```diff
@@ -396,21 +396,36 @@
...
           <dfn>automated install prompt</dfn>: that is, a <abbr>UI</abbr> that
           the user agent presents to the user when, for instance, there are
           sufficient <a>installability signals</a> to warrant
           <a>installation</a> of the web application.
-        </p>
+          </li>
+          <li>The <a>installation process</a> can occur through a
+          <dfn>site-triggered install prompt</dfn>: the site can
+          programmatically request that the user agent present an install
+          prompt to the user. The availability of this feature may be
```
marcoscaceres: s/The availability of this feature may be/The user agent MAY restrict the availability of this feature/

index.html
```diff
@@ -396,21 +396,36 @@
...
+          <a>installability signals</a> to warrant <a>installation</a> of the
+          web application.
+          </li>
+        </ul>
         <p>
           Prior to presenting an <a>automated install prompt</a>, a user agent
-          MUST run the <a>steps to notify before an automated install
-          prompt</a>.
+          MUST run the <a>steps to notify that an install prompt is
+          available</a>, to give the site the opportunity to suppress the
```
marcoscaceres: I wonder if we should say, instead of "to give the site the opportunity to suppress the install prompt" something like "prevent the default action (which is to install the application)". That would maybe align better with `. preventDefault()`

index.html
```diff
@@ -396,21 +396,36 @@
...
+          install prompt. Alternatively, the user agent may run the <a>steps to
```
marcoscaceres: s/may/MAY

index.html
```diff
@@ -531,6 +545,13 @@ <h3 id="installation-sec">
         <h3>
           <code>BeforeInstallPromptEvent</code> Interface
         </h3>
+        <div class="note">
+          The <code>beforeinstallprompt</code> event is somewhat misnamed, as
+          it does not necessarily signal that an <a>automated install
+          prompt</a> will follow (depending on the user agent, it may just be
```
marcoscaceres: s/it may just/it can just
or "it might just"
(avoid RFC2119 keywords in notes... even in lowercase)

index.html
```diff
@@ -634,10 +658,10 @@ <h3 id="installation-sec">
           </h4>
           <p>
             This example shows how one might prevent an automated install
-            prompt from showing until the user clicks a button to install the
-            app. In this way, the site can leave installation at the user's
-            discretion (rather than prompting at an arbitrary time), whilst
-            still providing a prominent UI to do so.
+            prompt from showing until the user clicks a button to show a
+            site-triggered install prompt. In this way, the site can leave
```
marcoscaceres: nit:  link back to def `<a>site-triggered install prompt</a>`

marcoscaceres:
@mgiuca, just FWI, we added a magic service that generates previews/diff for us. It adds the links at the top of the Pull Request. 

screenshot 2017-06-07 18 19 05

(PS: if you want it for Web Share, just copy this file to that repo: https://github.com/w3c/manifest/blob/gh-pages/.pr-preview.json  and make any changes needed. Should be straight forward)

marcoscaceres:
*FYI

mgiuca:
Done all of those. Thanks for the review.

Whoa, the pr-preview is magical. Trying it on Web Share.

PR #590, opened 2017-07-26
Scope matching algorithm: fix and explain

### LINKED ISSUE #554: Should the scope terminate by "/"?
The step 3 of "within scope algorithm" says:

> If target is same origin as scope and target's pathname starts with scope's pathname, return true.

Suppose the `scope` is "https://example.com/foo", and `target` is "https://example.com/foobar.html", it is within scope. Is this WAI?

mgiuca:
Fixes the definition of path matching (does not change behaviour) and adds an explanation.

Closes #554.

***
[Preview](https://s3.amazonaws.com/pr-preview/mgiuca/manifest/scope-matching-fix.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/9b87698...mgiuca:353336c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -790,6 +790,15 @@ window.addEventListener("appinstalled", handleInstalled);
         <li>Otherwise, return <code>false</code>.
         </li>
       </ol>
+      <div class="note" title="Scope is a simple string match">
+        The URL string matching in this algorithm is prefix-based rather than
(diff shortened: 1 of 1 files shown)


marcoscaceres:
LGTM! thanks for putting it together.

PR #592, opened 2017-08-03
Delete some errant apostrophes.

mcgreevy:
Note: I'm not sure whether it should be:
> installation process UI

or

> installation process's UI

but 

> installation process' UI

is not right.

***
[Preview](https://s3.amazonaws.com/pr-preview/mcgreevy/manifest/gh-pages.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/effbdc3...mcgreevy:79255f6.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -359,8 +359,8 @@
               </li>
               <li>If <a>obtaining the manifest</a> results in an error, a user
               agent can, at this point, fall back to using the <a>top-level
-              browsing context</a>' <code>Document</code>'s metadata to
-              populate an <a>installation process</a>' UI.
+              browsing context</a> <code>Document</code>'s metadata to
+              populate an <a>installation process</a> UI.
               </li>
               <li>If <a>obtaining the manifest</a> succeeds, and the result of
               running the <a>steps for processing the serviceworker member</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -359,16 +359,16 @@
               </li>
               <li>If <a>obtaining the manifest</a> results in an error, a user
               agent can, at this point, fall back to using the <a>top-level
-              browsing context</a>' <code>Document</code>'s metadata to
-              populate an <a>installation process</a>' UI.
+              browsing context</a> <code>Document</code>'s metadata to populate
+              an <a>installation process</a> UI.
```
mgiuca: I think this should read "populate the installation process's UI" (i.e., s/an/the) because the installation process was already instantiated by Step 2.1.

I think the apostrophe here was the author following Rule 1c [here](http://www.grammarbook.com/punctuation/apostro.asp): "Some writers and editors add only an apostrophe to all nouns ending in s." So it is correct by some standards, but I think it's pretty unusual and better with an 's.

mgiuca:
@marcoscaceres Do you know whether @mcgreevy needs to deal with the lpr checks before submitting this trivial change?

marcoscaceres:
No need to deal with IPR. I'll override it.

index.html
marcoscaceres: I was indeed following Rule 1c.

marcoscaceres: I also agree with @mgiuca - @mcgreevy, can you please make the change as @mgiuca suggests above?

mcgreevy: Done.

marcoscaceres:
Thanks @mcgreevy. Sorry for the churn on such a small change. Appreciate your time.

mcgreevy:
That's fine; I expected this change to have a big overhead:usefulness ratio; I mostly wanted to dip my toe into the water to make sure that I was able contribute to this repo as needed.

marcoscaceres:
@mcgreevy, looking forward to more contributions from you. Any questions, comments, etc. just give us a shout.

PR #598, opened 2017-08-09
feat: add categories member (resolves #569)

### LINKED ISSUE #569: Add (optional) "category" member
Many app stores including [Apple](https://developer.apple.com/app-store/categories/), [Google](https://support.google.com/googleplay/android-developer/answer/113475?hl=en), and [Microsoft](https://docs.microsoft.com/en-us/windows/uwp/publish/category-and-subcategory-table) ask developers to choose a "category" for their app.  This would be a good (optional) field to enable web app developers to suggest which category their app should be listed in.

```json
  "category": "sports",
```

**FAQ:**
**Q: These stores (and others) don't have the same categories.  How would you align?**
A: Actually, they have many (9) of the exact same categories: Business, Education, Entertainment, Health & Fitness, Lifestyle, Medical, Productivity, Shopping, and Sports
A: And many (14) categories that are very similar:
* Book-like categories: Books, Books & Reference, Books + reference
* Food-like categories: Food & Drink, Food + dining
* Magazines and Newspapers categories: Magazines & Newspapers, News & Magazines, 
* Music-like categories: Music, Music & Audio
* Navigation-like categories: Navigation, Maps & Navigation, Navigation + maps 
A: The Manifest spec COULD define a canonical set of categories (or include this as informational in a related wiki page)

**Q: What about sub-categories?**
A: Similar to the old "font" tag, web app developers could list desired categories in descending specificity.
For example: `"category": "Hotels, Travel, Lifestyle"`

**Q: What about non-English strings?**
A: Just as other enumerations in the Manifest use English strings, these categories would use English strings. 
However, there would be nothing preventing a developer from listing non-English strings. 
For example: `"category": "Libros, Books, Reference"`

RobDolinMS:
See discussion in #569 

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/manifest/RobDolinMS-category.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/9846cef...ca6f0e1.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2674,8 +2674,10 @@ Content-Type: application/manifest+json
           The categories string array is case insensitive and converted 
           to lower-case by following the processing algorithm. Thus, 
           <code>sports</code>, <code>Sports</code>, <code>SPORTS</code>, 
-          and <code>SpOrTs</code> are all equivalent.<br />
-          NOTE: Manifest authors are encouraged to use lower-case.
+          and <code>SpOrTs</code> are all equivalent.
+        </p>
+        <p class="note">
+          Manifest authors are encouraged to use lower-case.
         </p>
         <p class="note">
           This specification does not define the particular values for a the
--- a/index.html
+++ b/index.html
@@ -1522,9 +1522,9 @@ Content-Type: application/manifest+json
           <code>background_color</code> member</a> with <var>manifest</var> as
           argument.
           </li>
-          <li>Let <var>category</var> of <var>parsed manifest</var> be
+          <li>Let <var>categories</var> of <var>parsed manifest</var> be
           the result of running the <a>steps for processing the
-          <code>category</code> member</a> with <var>manifest</var> as
+          <code>categories</code> member</a> with <var>manifest</var> as
           argument.
           </li>
           <li>Let <var>iarc_rating_id</var> of <var>parsed manifest</var> be
@@ -2617,61 +2617,75 @@ Content-Type: application/manifest+json
       </section>
       <section>
         <h3>
-          <code title="">category</code> member
+          <code title="">categories</code> member
         </h3>
         <p>
-          The <dfn id="member-category"><code>category</code>
-          member</dfn> describes the expected application category (or 
-          categories) to which the web application belongs. This is a hint 
-          to catalogs or stores of web applications under which category 
-          (or categories) to list this web application.
+          The <dfn id="member-categories"><code>categories</code>
+          member</dfn> describes the expected application categories (or 
+          single category) to which the web application belongs. This is a hint 
+          to catalogs or stores of web applications under which categories 
+          (or category) to list this web application.
         </p>
         <p>
-          The <code>category</code> member is only meant as a hint to catalogs 
+          The <code>categories</code> member is only meant as a hint to catalogs 
           or stores listing web applications and it is expected that these 
-          will make a best effort to find an appropriate category under which 
-          to list the web application.
+          will make a best effort to find appropriate categories (or category)
+          under which to list the web application.
         </p>
         <p>
-          The <dfn>steps for processing the <code>category</code>
+          The <dfn>steps for processing the <code>categories</code>
           member</dfn> are given by the following algorithm. The algorithm
-          takes a <var>manifest</var> as an argument. This algorithm returns a
-          string or <code>undefined</code>. The string MAY include a comma 
-          delimited list of categories.
+          takes a <var>manifest</var> as an argument. This algorithm returns an
+          array of strings or <code>undefined</code>. The array of strings MAY 
+          include only one value.
         </p>
         <ol>
-          <li>Let <var>value</var> be the result of calling the
-          <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
-          with argument "<code>category</code>".
-          </li>
-          <li>If <a>Type</a>(<var>value</var>) is not "string":
+          <li>Let <var>categories</var> be an empty list.</li>
+          <li>Let <var>unprocessed categories</var> be the result of calling
+          the <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
+          with argument "<code>categories</code>".</li>
+          <li>If <var>unprocessed categories</var>is an array, then:</li>
             <ol>
-              <li>If <a>Type</a>(<var>value</var>) is not "undefined", <a>issue
-              a developer warning</a> that the type is not supported.
-              </li>
-              <li>Return <code>undefined</code>.
+              <li>For each <var>potential category</var> in the array:</li>
+                <ol>
+                  <li>
+                    <a>Trim</a>(<var>value</var>) and convert 
+                    <a>to ASCII lowercase</a>
+                  </li>
+                  <li>
+                    Append <var>potential category</var> to 
+                    <var>categories</var>
+                  </li>
+                </ol>
+            </ol>
+          <li>
+            Otherwise, if <var>unprocessed categories</var> is not 
+            <code>undefined</code>
+          </li>
+            <ol>
+              <li>
+                <a>Issue a developer warning</a> that the type is not 
+                supported.
               </li>
             </ol>
-          </li>
-          <li><a>Trim</a>(<var>value</var>) and convert <a>to ASCII lowercase</a></li>
-          <li>Return <var>category</var>.
-          </li>
+          <li>Return <var>categories</var></li>
         </ol>
         <p>
-          The category string is case insensitive and converted to lower-case 
-          by following the processing algorithm. Thus, <code>sports</code>, 
-          <code>Sports</code>, <code>SPORTS</code>, and <code>SpOrTs</code> 
-          are all equivalent. Manifest authors SHOULD use lower-case.
+          The categories string array is case insensitive and converted 
+          to lower-case by following the processing algorithm. Thus, 
+          <code>sports</code>, <code>Sports</code>, <code>SPORTS</code>, 
+          and <code>SpOrTs</code> are all equivalent. Manifest authors 
+          SHOULD use lower-case.
         </p>
         <p>
-          The category string is a hint to catalogs and stores were the 
-          Manifest author suggests their web app should be listed. 
-          This is hint or suggestion from the Manifest author. Like 
-          search engine keywords, catalogs and stores are not required 
-          to honor this hint.</p>
+          The categories string array is a hint to catalogs and stores 
+          of where the Manifest author suggests their web app should be 
+          listed. This is hint or suggestion from the Manifest author. 
+          Like search engine keywords, catalogs and stores are not 
+          required to honor this hint.</p>
         <p>
-          If a category string has multiple comma delimited values, a 
-          consuming system MAY use one or more of these values.  
+          If a categories string array has multiple values, a consuming 
+          system MAY use one or more of these values.  
         </p>
         <p>
           Manifest authors SHOULD use the below common categories:
@@ -2682,10 +2696,11 @@ Content-Type: application/manifest+json
           <li>education</li>
           <li>entertainment</li>
           <li>finance</li>
+          <li>fitness</li>
           <li>food</li>
           <li>games</li>
-          <li>government & politics</li>
-          <li>health & fitness</li>
+          <li>government</li>
+          <li>health</li>
           <li>kids</li>
           <li>lifestyle</li>
           <li>magazines</li>
@@ -2695,6 +2710,7 @@ Content-Type: application/manifest+json
           <li>news</li>
           <li>personalization</li>
           <li>photo</li>
+          <li>politics</li
           <li>productivity</li>
           <li>security</li>
           <li>shopping</li>
--- a/index.html
+++ b/index.html
@@ -2641,9 +2641,9 @@ Content-Type: application/manifest+json
           <li>Let <var>unprocessed categories</var> be the result of calling
           the <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
           with argument "<code>categories</code>".</li>
-          <li>If <var>unprocessed categories</var>is an array, then:</li>
+          <li>If <var>unprocessed categories</var>is an array, then:
             <ol>
-              <li>For each <var>potential category</var> in the array:</li>
+              <li>For each <var>potential category</var> in the array:
                 <ol>
                   <li>
                     <a>Trim</a>(<var>value</var>) and convert 
@@ -2654,17 +2654,19 @@ Content-Type: application/manifest+json
                     <var>categories</var>
                   </li>
                 </ol>
+              </li>
             </ol>
+          </li>
           <li>
             Otherwise, if <var>unprocessed categories</var> is not 
             <code>undefined</code>
-          </li>
             <ol>
               <li>
                 <a>Issue a developer warning</a> that the type is not 
                 supported.
               </li>
             </ol>
+          </li>
           <li>Return <var>categories</var></li>
         </ol>
         <p>
--- a/index.html
+++ b/index.html
@@ -2662,12 +2662,6 @@ Content-Type: application/manifest+json
           <code>SpOrTs</code> are all equivalent.
         </p>
         <p>
-          Spaces, underscores, and dashes should be treated as equivalent. 
-          Thus, <code>health & fitness</code>,  
-          <code>health_&_fitness</code>, and <code>health-&-fitness</code> 
-          are all equivalent.
-        </p>
-        <p>
           If a category string has multiple comma delimited values, a 
           consuming system SHOULD use the most specific value that is is 
           familiar with, and ignore values that it is not familiar with.  
--- a/index.html
+++ b/index.html
@@ -2675,35 +2675,35 @@ Content-Type: application/manifest+json
         </p>
         <p>
           Manifest authors SHOULD use the below common categories:
-          <ul>
-            <li>books</li>
-            <li>business</li>
-            <li>education</li>
-            <li>entertainment</li>
-            <li>finance</li>
-            <li>food</li>
-            <li>games</li>
-            <li>government & politics</li>
-            <li>health & fitness</li>
-            <li>kids</li>
-            <li>lifestyle</li>
-            <li>magazines</li>
-            <li>medical</li>
-            <li>music</li>
-            <li>navigation</li>
-            <li>news</li>
-            <li>personalization</li>
-            <li>photo</li>
-            <li>productivity</li>
-            <li>security</li>
-            <li>shopping</li>
-            <li>social</li>
-            <li>sports</li>
-            <li>travel</li>
-            <li>utilities</li>
-            <li>weather</li>
-          </ul>
         </p>
+        <ul>
+          <li>books</li>
+          <li>business</li>
+          <li>education</li>
+          <li>entertainment</li>
+          <li>finance</li>
+          <li>food</li>
+          <li>games</li>
+          <li>government & politics</li>
+          <li>health & fitness</li>
+          <li>kids</li>
+          <li>lifestyle</li>
+          <li>magazines</li>
+          <li>medical</li>
+          <li>music</li>
+          <li>navigation</li>
+          <li>news</li>
+          <li>personalization</li>
+          <li>photo</li>
+          <li>productivity</li>
+          <li>security</li>
+          <li>shopping</li>
+          <li>social</li>
+          <li>sports</li>
+          <li>travel</li>
+          <li>utilities</li>
+          <li>weather</li>
+        </ul>
       </section>
       <section>
         <h3>
--- a/index.html
+++ b/index.html
@@ -2628,7 +2628,8 @@ Content-Type: application/manifest+json
           The <code>categories</code> member is only meant as a hint to catalogs 
           or stores listing web applications and it is expected that these 
           will make a best effort to find appropriate categories (or category)
-          under which to list the web application.
+          under which to list the web application. Like search engines and 
+          meta keywords, catalogs and stores are not required to honor this hint.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+          <code>health_&_fitness</code>, and <code>health-&-fitness</code> 
+          are all equivalent.
+        </p>
+        <p>
+          If a category string has multiple comma delimited values, a 
+          consuming system SHOULD use the most specific value that is is 
+          familiar with, and ignore values that it is not familiar with.  
+          For example: <code>Books, Fiction, Science Fiction & Fantasty</code>.
+        </p>
+        <p>
```
marcoscaceres: Nit: `ul` element is not allowed inside `p`. The  validator will yell at us :)

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
       </section>
       <section>
         <h3>
+          <code title="">category</code> member
```
kenchris: We don't have "The" in front of these? Could you check?

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+        </h3>
+        <p>
+          The <dfn id="member-category"><code>category</code>
+          member</dfn> describes the expected application category to which the 
+          web application belongs. This is a hint to catalogs or stores of web
```
kenchris: Can it be more than one? should we state that? Then would order matter?

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+              </li>
+            </ol>
+          </li>
+          <li>Return <var>category</var>.
+          </li>
+        </ol>
+        <p>
+          The category string should be treated as case insensitive. Thus, 
+          <code>sports</code>, <code>Sports</code>, <code>SPORTS</code>, and 
+          <code>SpOrTs</code> are all equivalent.
```
kenchris: Why not let the algorithm them lowercase the value?

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+        </p>
+        <p>
+          Spaces, underscores, and dashes should be treated as equivalent.
```
marcoscaceres: I think the should pick one. Being strict is fine.

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+        </ol>
+        <p>
+          The category string should be treated as case insensitive. Thus, 
+          <code>sports</code>, <code>Sports</code>, <code>SPORTS</code>, and 
+          <code>SpOrTs</code> are all equivalent.
+        </p>
+        <p>
+          Spaces, underscores, and dashes should be treated as equivalent. 
+          Thus, <code>health & fitness</code>,  
+          <code>health_&_fitness</code>, and <code>health-&-fitness</code>
```
kenchris: Maybe same here, @marcoscaceres what do you think?

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+          Spaces, underscores, and dashes should be treated as equivalent. 
+          Thus, <code>health & fitness</code>,  
+          <code>health_&_fitness</code>, and <code>health-&-fitness</code> 
+          are all equivalent.
+        </p>
+        <p>
+          If a category string has multiple comma delimited values, a 
+          consuming system SHOULD use the most specific value that is is 
+          familiar with, and ignore values that it is not familiar with.  
+          For example: <code>Books, Fiction, Science Fiction & Fantasty</code>.
```
kenchris: What about order if it is just as familiar with Books as it is with Fiction?

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+          Manifest authors SHOULD use the below common categories:
+          <ul>
+            <li>books</li>
+            <li>business</li>
+            <li>education</li>
+            <li>entertainment</li>
+            <li>finance</li>
+            <li>food</li>
+            <li>games</li>
+            <li>government & politics</li>
```
kenchris: Should we escape & here?

index.html
marcoscaceres: I need to check what we currently normalize... I know we don't normalize some other things, and that got baked into Chrome.

index.html
marcoscaceres: I don't think this list should be in the spec. We should move it somewhere else where it can be updated and maintained independently of the spec.

kenchris: If we don't expect these to change a lot, I think it is fine to have them in the spec.

index.html
RobDolinMS: Thanks.  I'll remove the <p> tags enclosing the <ul>

index.html
RobDolinMS: From what I can tell, the heading/h3 text is just "foo" member.  

You are correct that we use "The" to start the paragraph/p text as in: "The foo member" 

I believe this PR matches that style.

index.html
RobDolinMS: Good suggestion @kenchris.  I'll clarify that it can be a single category or multiple categories.

index.html
kenchris: Thanks for checking

index.html
kenchris: I wonder if we can avoid this & altogether. We already allow to add multiple categories, so one store could put "health" in the "Health and Fitness" group, and you could event set the category to "health, fitness" and the store could use both values to decide whether to add to its "Health and Medicin" or the "Health and Fitness" category.

index.html
RobDolinMS: Please let me know what you recommend here @marcoscaceres

index.html
RobDolinMS: Sounds good @marcoscaceres.  I'll remove this portion just focusing on spaces.

index.html
RobDolinMS: I've added language to the algorithm to `trim` and `convert to lower-case` matching what we do for the `dir` member.

index.html
RobDolinMS: I've now closed the paragraph tag with `</p>` before the `<ul>` tag.

index.html
RobDolinMS: These have been pretty static and represent a consistent list from looking at stores from Apple, Google, and Microsoft.

index.html
RobDolinMS: I've clarified that these are hints/suggestions and softened the language about which category a consuming catalog or store should use.

index.html
```diff
@@ -2653,6 +2653,7 @@ <h3 id="applying">
               </li>
             </ol>
           </li>
+          <li><a>Trim</a>(<var>value</var>) and convert to lower-case</li>
```
kenchris: like Trim is defined, we should define ToLowerCase

ie https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/ToLowerCase

kenchris: Also I notices that the links at the end of Trim are not dead links @marcoscaceres

index.html
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+        </p>
+        <ul>
+          <li>books</li>
+          <li>business</li>
+          <li>education</li>
+          <li>entertainment</li>
+          <li>finance</li>
+          <li>food</li>
+          <li>games</li>
+          <li>government & politics</li>
```
kenchris: Should we update these now? add government and politics as two entries? Maybe add an example?

kenchris:
link ToLowerCase to https://www.ecma-international.org/ecma-262/8.0/index.html#sec-string.prototype.tolowercase

If you can fix Trim to link to https://www.ecma-international.org/ecma-262/8.0/index.html#sec-string.prototype.trim

that would be great as well :)

index.html
```diff
@@ -2653,7 +2653,7 @@ <h3 id="applying">
               </li>
             </ol>
           </li>
-          <li><a>Trim</a>(<var>value</var>) and convert to lower-case</li>
+          <li><a>Trim</a>(<var>value</var>) and convert <a>to ASCII lowercase</a></li>
```
kenchris: Why not

```
<li><a>Trim</a>(<var>value</var>) and <a>ToLowerCase</a>(<var>value</var>)

Comment (kenchris) [inline index.html]: And elsewhere

```

When instructed to ToLowerCase(x), a user agent must behave as if [ECMASCRIPT]'s String.prototype.toLowerCase() function had been called on the string x.
```

Comment (kenchris) [inline index.html]: OK I see that we use to ASCII lowercase elsewhere and that is from https://html.spec.whatwg.org/#converted-to-ascii-lowercase

so that might be better... I wonder if there is something for trim as well

Comment (kenchris) [inline index.html]: Seems to: https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace

@marcoscaceres what would be the best to use here? 

There is a bug in "dir" as well, where we write

```Otherwise, Trim(value), convert to lower-case, and let direction be the result.```

Comment (kenchris) [inline index.html]: I think this is not a fault of your patch, so I filed bugs instead.

Could you look at the & issue and then maybe we can go ahead and merge?

Comment (RobDolinMS) [inline index.html]: What you propose is what I was thinking, but I searched the spec for "lower" and found:
* Three instances of: `to lower-case` (no `<a>` tag)
* Two instances of:  `to ASCII lowercase` (with `<a>` tag)

Do you know if there's any reason that all of five of these should not just follow your suggestion?

Comment (kenchris) [inline index.html]: I am not sure, but I filed a bug for this. Feel free to add more info to the bug.

I think for this patch you have done what is required and that part is fine. We can fix all the references in another PR when we decide what to do

**Comment by kenchris** [comment]
As we discussed having ```"category": "fitness, health"``` to able to put that app into a "Health & Fitness" category in a store, it makes sense to split these up in the category name list, ie "fitness" and "health" as separate values. Potentially with an example of what a store might do in such a case

File: index.html (hunk 11 above)
Comment (RobDolinMS) [inline index.html]: I just filed Issue #602 for "lower"

**Comment by kenchris** [comment]
Contributors: RobDolinMS did not make IPR commitments for this group

Is this true @RobDolinMS - how did @marcoscaceres handle it for the other PR? @anssiko ?

**Comment by RobDolinMS** [comment]
Re: [IPR](https://github.com/w3c/manifest/pull/598#issuecomment-321863979) - I believe Microsoft is a member of the Web Platform WG under which the Manifest work resides.  I'm glad to complete an CLA (or similar) as needed to help unblock.  /cc @marcoscaceres @kenchris @anssiko

**Comment by chaals** [comment]
Yes, Microsoft is inside the 45-day period usually granted as grace to re-subscribe officially to a Working Group when the charter is renewed, which the tooling doesn't understand. So @RobDolinMS is OK for IPR.

**Comment by kenchris** [comment]
OK so you would suggest that we just merge then, ignoring the IPR check? I don't want to make it as non substantial

**Comment by kenchris** [comment]
@RobDolinMS can you fix my last comment about &?

**Comment by marcoscaceres** [comment]
Thanks @chaals, I'll override the IPR issue once this is ready to go. 

I'd still like to do a "deep dive" review of the final text once @kenchris' initial review is complete.

File: index.html (hunk 10 above)
Comment (RobDolinMS) [inline index.html]: Good question @kenchris. based on an analysis of what categories are used by the Apple, Google, and Microsoft stores (below), I believe we should treat health and fitness as a single category since all three of these stores do so: 

| Apple | Google | Microsoft |
 Health & Fitness | Health & Fitness | Health + fitness |

Comment (kenchris) [inline index.html]: Yes, and my fear is that you will see developers write 

"Fitness & Health" (flipped)
"Health And Fitness" (And instead of &)
...

and that is pretty confusing.

On the other hand, if the dev were to write "health" or "fitness" or "health, fitness" or "fitness, health" that would be parsed as one of two entries which could easily be mapped to a "Health & Fitness" section within existing stores.

If in the future, say Apple wants to create  a new category "Health" separate from "Health & Fitness"  it might be able to re-categorize.

We can handle all of these automatically and avoid developers making mistakes. I already got feedback from the author of outweb.io that he really disliked the "&"

Comment (kenchris) [inline index.html]: @marcoscaceres maybe you can make a decision :)

Comment (marcoscaceres) [inline index.html]: Let's just do individual. These are keys, after all - they are not the things displayed by stores. So:

```
categories: ["government", "health", "bar", "politics"]
```

Will show up in a store as: "<li>government & politics</li>" or "<li>governo y politico</li>"  in the same store in Spanish.

Comment (marcoscaceres) [inline index.html]: Also, we don't need "health & fitness", because "fitness" === "health & fitness" for display purposes.

Comment (marcoscaceres) [inline index.html]: So: just "fitness" and just "politics" for the two "&" ones.

File: index.html [hunk 12]
```diff
@@ -2612,6 +2617,96 @@ <h3 id="applying">
...
+          The <code>category</code> member is only meant as a hint to catalogs 
+          or stores listing web applications and it is expected that these 
+          will make a best effort to find an appropriate category under which 
+          to list the web application.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>category</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <var>manifest</var> as an argument. This algorithm returns a
+          string or <code>undefined</code>. The string MAY include a comma
```
kenchris: Shouldn't we use an array instead of parsing values seperated with commas? @marcoscaceres

index.html
marcoscaceres: This also allows us to put ascii lowercase normalization on all these for the purpose of processing. Then stores can just reinterpret them however they want.

index.html
marcoscaceres: Yeah, I think we should use an array. That's what they are there for :) Hopefully we didn't screw that up elsewhere. I can't remember :(

kenchris: I don't think so, but I remember that we almost did :)

RccnMarius:
I really like the idea of having an array with single word categories. It makes it very easy for app stores to find the right category. Great work!

RobDolinMS:
Thanks @marcoscaceres, @kenchris, and @RccnMarius for your feedback.  I've updated from `category` to `categories` and also separated the two-word categories like `health & fitness`

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+          of where the Manifest author suggests their web app should be 
+          listed. This is hint or suggestion from the Manifest author. 
+          Like search engine keywords, catalogs and stores are not 
+          required to honor this hint.</p>
+        <p>
+          If a categories string array has multiple values, a consuming 
+          system MAY use one or more of these values.  
+        </p>
+        <p>
+          Manifest authors SHOULD use the below common categories:
```
marcoscaceres: We probably need to make this into a registry. The use of plural and singular form seems a bit arbitrary. 

@kenchris, can you check?

marcoscaceres: About the registry: we need to do something similar to https://www.w3.org/Payments/card-network-ids

In a separate document.

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
       </section>
       <section>
         <h3>
+          <code title="">categories</code> member
+        </h3>
+        <p>
+          The <dfn id="member-categories"><code>categories</code>
+          member</dfn> describes the expected application categories (or
```
RobDolinMS: Removed

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
       </section>
       <section>
         <h3>
+          <code title="">categories</code> member
+        </h3>
+        <p>
+          The <dfn id="member-categories"><code>categories</code>
+          member</dfn> describes the expected application categories (or 
+          single category) to which the web application belongs. This is a hint
```
RobDolinMS: Removed.

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+          The <code>categories</code> member is only meant as a hint to catalogs 
+          or stores listing web applications and it is expected that these 
+          will make a best effort to find appropriate categories (or category)
+          under which to list the web application.
+        </p>
+        <p>
+          The <dfn>steps for processing the <code>categories</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <var>manifest</var> as an argument. This algorithm returns an
+          array of strings or <code>undefined</code>. The array of strings MAY
```
RobDolinMS: Removed

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+          takes a <var>manifest</var> as an argument. This algorithm returns an
+          array of strings or <code>undefined</code>. The array of strings MAY 
+          include only one value.
+        </p>
+        <ol>
+          <li>Let <var>categories</var> be an empty list.</li>
+          <li>Let <var>unprocessed categories</var> be the result of calling
+          the <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
+          with argument "<code>categories</code>".</li>
+          <li>If <var>unprocessed categories</var>is an array, then:</li>
```
RobDolinMS: Fixed

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+                  <li>
+                    Append <var>potential category</var> to 
+                    <var>categories</var>
+                  </li>
+                </ol>
+            </ol>
+          <li>
+            Otherwise, if <var>unprocessed categories</var> is not 
+            <code>undefined</code>
+          </li>
```
RobDolinMS: Fixed

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+              </li>
+            </ol>
+          <li>Return <var>categories</var></li>
+        </ol>
+        <p>
+          The categories string array is case insensitive and converted 
+          to lower-case by following the processing algorithm. Thus, 
+          <code>sports</code>, <code>Sports</code>, <code>SPORTS</code>, 
+          and <code>SpOrTs</code> are all equivalent. Manifest authors 
+          SHOULD use lower-case.
```
RobDolinMS: Removed and added note

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+        </p>
+        <p>
+          The categories string array is a hint to catalogs and stores
```
RobDolinMS: Removed

index.html
```diff
@@ -2612,6 +2617,112 @@ <h3 id="applying">
...
+          SHOULD use lower-case.
+        </p>
+        <p>
+          The categories string array is a hint to catalogs and stores 
+          of where the Manifest author suggests their web app should be 
+          listed. This is hint or suggestion from the Manifest author. 
+          Like search engine keywords, catalogs and stores are not 
+          required to honor this hint.</p>
+        <p>
+          If a categories string array has multiple values, a consuming
```
RobDolinMS: Removed

index.html
```diff
@@ -2612,6 +2617,102 @@ <h3 id="applying">
...
+          <li>kids</li>
+          <li>lifestyle</li>
+          <li>magazines</li>
+          <li>medical</li>
+          <li>music</li>
+          <li>navigation</li>
+          <li>news</li>
+          <li>personalization</li>
+          <li>photo</li>
+          <li>politics</li
```
kenchris: missing > after </li

RccnMarius:
Please add category for software development related apps.

marcoscaceres:
So @RccnMarius implicitly demonstrates why the registry needs to be in an external document. The list of categories will continue to grow and change long after this spec is done.

kenchris:
Can't that list not be auto added to the spec then? Or we add a list now and links to a place with updated categories and can add those in a level 2 etc

index.html
```diff
@@ -1522,6 +1522,11 @@ <h3 id="navigation-scope-security-considerations">
           <code>background_color</code> member</a> with <var>manifest</var> as
```
kenchris: Guess you could add yourself as editor to this patch? @marcoscaceres ?

index.html
```diff
@@ -2612,6 +2617,102 @@ <h3 id="applying">
...
+          <li>photo</li>
+          <li>politics</li
+          <li>productivity</li>
+          <li>security</li>
+          <li>shopping</li>
+          <li>social</li>
+          <li>sports</li>
+          <li>travel</li>
+          <li>utilities</li>
+          <li>weather</li>
```
kenchris: I am fine with this list in the spec, but I think we should create a wiki like https://github.com/w3c/manifest/wiki/Extensions-Registry maybe https://github.com/w3c/manifest/wiki/Categories that replicates the list and where we can add new ones after a published stable spec, as they become available.

We would of course have to link to that wiki entry in the spec with appropriate text.

kenchris: OK, I created this page, feel free to edit

index.html
RobDolinMS: Doh; missing `>`  Fixing now...

index.html
RobDolinMS: @kenchris I'd be honored to do so; but do I need any sort of formal approval from the editors and/or WG Chairs?

index.html
kenchris: @RobDolinMS could you add some text and link to this?

index.html
kenchris: As long as you are approved to actually contribute to the spec, ie WG member etc, I don't see why Marcos and I, couldn't just add/nominate you. Maybe @marcoscaceres would know,

marcoscaceres: @RobDolinMS you don't need any formal approval.

index.html
RobDolinMS: Thanks @kenchris.  I've just added text (lines 2680 - 2686) mirroring the language used for the platform member.

Do you think I should also delete the list of common categories values from the spec (since it's in the wiki page?)

index.html
marcoscaceres: @kenchris let's do this separately. @RobDolinMS will need to find his W3C ID, etc.

index.html
marcoscaceres: I think they should be removed, to be honest.

marcoscaceres:
Merged. I've tidied on gh-pages branch.

PR #601, opened 2017-08-11
Update link destination for Trim( ) definition

RobDolinMS:
This moves the reference from a paragraph to be more consistent with other references 

This was craeted per discussion with @KenChris on PR #598, but I'm not sure if this is what was intended.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/manifest/RobDolinMS-trim-definition.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/797c5b1...b70522a.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -3652,12 +3652,6 @@
         notation are defined in [[!ECMASCRIPT]].
       </p>
       <p>
-        When instructed to <dfn>Trim</dfn>(<var>x</var>), a user agent MUST
-        behave as if [[!ECMASCRIPT]]'s <a data-cite=
-        "!ECMASCRIPT#sec-string.prototype.trim">String.prototype.trim()</a>
-        function had been called on the string <var>x</var>.
-      </p>
-      <p>
         As the manifest uses the JSON format, this specification relies on the
         types defined in [[!ECMA-404]] specification: namely <dfn>object</dfn>,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3652,12 +3652,6 @@ <h3 id="applying">
         notation are defined in [[!ECMASCRIPT]].
       </p>
       <p>
-        When instructed to <dfn>Trim</dfn>(<var>x</var>), a user agent MUST
-        behave as if [[!ECMASCRIPT]]'s <a data-cite=
-        "!ECMASCRIPT#sec-string.prototype.trim">String.prototype.trim()</a>
```
kenchris: I wonder if @marcoscaceres needs to update the ECMASCRIPT reference is one of the respec tools instead.

kenchris: its seems that respec gets the list from http://www.specref.org/?q=ECMASCRIPT which shows https://tc39.github.io/ecma402/ (current editors draft) which is not what shows up in our spec

kenchris: @marcoscaceres do you have any idea about how to solve this the proper way?

marcoscaceres: https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace

marcoscaceres: Let's use infra where we can.

marcoscaceres:
I'll use infra and update the references when we do another round of cleanup.

PR #612, opened 2017-09-19
Add IDL section and define IDL for all of the members

kenchris:
Created an IDL section in appendix and defined some of the members as IDL.

I can continue this, but would like some initial feedback :-)

***
[Preview](https://s3.amazonaws.com/pr-preview/kenchris/manifest/gh-pages.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/8c837b6...kenchris:fa084d2.html)


mgiuca:
Hi Ken,

This is a good start. Thanks for taking this; I wasn't expecting you to do it but if you're happy to continue, go ahead.

One point that keeps coming up is DOMString vs USVString. The difference is DOMString allows illegal surrogate pairs while USVString does not. The WebIDL spec tells you not to use USVString unless you have a good reason (the main accepted use of USVString is URLs). My opinion is the opposite; that USVString should generally be preferred because illegal surrogate pairs are meaningless, and cannot be represented outside of UTF-16. I think the rule of thumb should be that if the string characters are displayed to the user or leave the browser, then it should be a USVString, but if it's an internal string only (e.g., an ID string) then it can be a DOMString.

I successfully argued this logic in Web Share which is why the [ShareData fields are USVStrings](https://wicg.github.io/web-share/#sharedata-dictionary).

At the very least, all URL fields should be USVStrings. For Manifest, many of the strings will be exported out to the OS (e.g., on shortcuts) and illegal surrogate pairs may not be representable. I would advise the following strings be USVString:

- name
- short_name
- description
- scope
- start_url
- related_applications: all fields
- categories
- Image.src (but not size)

Agree/disagree?

marcoscaceres:
Apart from agreeing with @mgiuca about URLs, I don't have a strong opinion about the others.

marcoscaceres:
... but as @mgiuca points out, these strings leave the browser... so USVString might be safer.

kenchris:
Great, that was actually one of those questions that I had. I will move it all to USVString

kenchris:
Ok, new version up, though I don't know how to fix the platform member because it is reused across two members and I cannot have two data-dfn-for. Help me Obi Wan Caceres @marcoscaceres, you are my only hope :-)

mgiuca:
How come this got reverted? (e0e053f)

marcoscaceres:
@mgiuca, @kenchris accidentally pushed on main branch, so he reverted it.

marcoscaceres:
The changes from e0e053f are in https://github.com/w3c/manifest/pull/613

PR #613, opened 2017-09-20
Add IDL section and define IDL for all the members

kenchris:
***
[Preview](https://s3.amazonaws.com/pr-preview/kenchris/manifest/idl.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/e553652...kenchris:54af834.html)


kenchris:
@mgiuca I did a little screw up, so new PR now :-) Comments welcome

index.html
```diff
@@ -1592,17 +1590,40 @@ <h3 id="applying">
...
-        as well as how their values are processed.
+        launched. A manifest consists of a top-level <dfn>WebAppManifest</dfn>
+        <a>object</a> that contains zero or more members.
+        Each of the members are defined below, as well as how their values are processed.
       </p>
+      <pre class="idl">
+          [NoInterfaceObject]
+          dictionary WebAppManifest {
+             TextDirectionType? dir;
+             USVString? lang;
```
mgiuca: Conversely to what I said yesterday, I think "lang", "iarc_rating_id", and any other "id" or other ASCII-only fields could be DOMString instead of USVString. Practically, it probably doesn't really matter, but the WebIDL spec says "if in doubt, use DOMString".

index.html
```diff
@@ -2769,47 +2803,58 @@ <h3 id="applying">
...
     <section>
       <h2>
-        Image object and its members
+        ImageResource and its members
       </h2>
       <div class="issue" data-number="361"></div>
+      <pre class="idl">
+        [NoInterfaceObject]
+        dictionary ImageResource {
+          USVString src;
```
mgiuca: required?

index.html
```diff
@@ -2769,47 +2803,58 @@ <h3 id="applying">
...
+          DOMString sizes;
```
mgiuca: required?

index.html
```diff
@@ -3292,25 +3335,33 @@ <h3 id="applying">
...
       <p>
-        A <dfn>serviceworker object</dfn> represents a service worker
+        A <dfn>ServiceWorkerRegistration</dfn> object represents a service worker
         registration for the web application.
       </p>
-      <section>
+      <pre class="idl">
+        [NoInterfaceObject]
+        dictionary ServiceWorkerRegistration {
+          USVString src;
```
mgiuca: required?

index.html
```diff
@@ -3472,11 +3522,21 @@ <h3 id="applying">
     </section>
     <section>
       <h2>
-        Application object and its members
+        ExternalApplicationResource and its members
       </h2>
+      <pre class="idl">
+        [NoInterfaceObject]
+        dictionary ExternalApplicationResource {
+          USVString platform;
```
mgiuca: required

index.html
marcoscaceres: Agree. Let's change this to DOMString... it will make this less controversial if this later gets reviewed by browser folks.

index.html
```diff
@@ -1592,17 +1590,40 @@ <h3 id="applying">
...
         parameters and application defaults for when a web application is
-        launched. A manifest consists of a top-level <a>object</a> that
-        contains zero or more members. Each of the members are defined below,
-        as well as how their values are processed.
+        launched. A manifest consists of a top-level <dfn>WebAppManifest</dfn>
+        <a>object</a> that contains zero or more members.
+        Each of the members are defined below, as well as how their values are processed.
       </p>
+      <pre class="idl">
+          [NoInterfaceObject]
```
marcoscaceres: Remove [NoInterfaceObject]  ... it's not an interface, it's a dictionary 😉. Also, you should never ever need to use NoInterfaceObject in a spec unless you are creating some kind of mix-in.

index.html
```diff
@@ -1592,17 +1590,40 @@ <h3 id="applying">
         <div class="issue" data-number="446"></div>
       </section>
     </section>
-    <section>
+    <section data-dfn-for="webappmanifest" data-link-for="webappmanifest">
```
marcoscaceres: The `data-dfn-for="webappmanifest" data-link-for="webappmanifest"` need to match actual IDL things, so, should be "WebAppManifest".

index.html
```diff
@@ -1592,17 +1590,40 @@ <h3 id="applying">
...
         Manifest and its members
       </h2>
       <p>
         A <dfn>manifest</dfn> is a JSON document that contains startup
         parameters and application defaults for when a web application is
-        launched. A manifest consists of a top-level <a>object</a> that
-        contains zero or more members. Each of the members are defined below,
-        as well as how their values are processed.
+        launched. A manifest consists of a top-level <dfn>WebAppManifest</dfn>
+        <a>object</a> that contains zero or more members.
```
marcoscaceres: It's a dictionary, not an object.

index.html
```diff
@@ -1592,17 +1590,40 @@ <h3 id="applying">
...
+        Each of the members are defined below, as well as how their values are processed.
```
marcoscaceres: Nit: make sure you run tidy over the doc.

index.html
```diff
@@ -1612,45 +1633,49 @@ <h3 id="applying">
           <code>dir</code> member
```
marcoscaceres: Can you replace this with: `<dfn>dir</dfn>member`, and the same for the rest?

marcoscaceres: you won't need `<code>`, ReSpec will automatically code them.

index.html
```diff
@@ -1612,45 +1633,49 @@ <h3 id="applying">
...
+            [NoInterfaceObject]
+            enum TextDirectionType { "ltr", "rtl", "auto" };
+        </pre>
         <p>
           The <dfn>directionality-capable members</dfn> are:
         </p>
         <ul>
           <li>
-            <a><code>description</code> member</a>.
+            <a><code>description</code></a> member.
```
marcoscaceres: s/`<a><code>description</code></a> member.`/`<a>description</a> member.`

Same for the rest. You should never need `<code>` anywhere.

index.html
```diff
@@ -1612,45 +1633,49 @@ <h3 id="applying">
...
           <li>
-            <a><code>short_name</code> member</a>.
+            <a><code>short_name</code></a> member.
           </li>
         </ul>
         <p>
-          The <dfn data-lt="text-direction value">text-direction values</dfn>
-          are the following, implying that the value of the
+          The <dfn data-lt="text-direction value">text-direction values</dfn> defined by
+          <dfn>TextDirectionType</dfn>, are the following, implying that the value of the
```
marcoscaceres: The `TextDirectionType` should have it's own top-level section.

index.html
```diff
@@ -2090,8 +2124,8 @@ <h3 id="applying">
           <code>orientation</code> member
         </h3>
         <p>
-          The <dfn id="member-orientation"><code>orientation</code>
-          member</dfn> is a <a>string</a> that serves as the <a>default
+          The <dfn><code>orientation</code></dfn>
```
marcoscaceres: s/`<dfn><code>orientation</code></dfn>`/`<dfn>orientation</dfn> member`

As above, the `dfn` for this should occur inside a `h3` element in its own section.

marcoscaceres: The above applies for all the things below.

index.html
marcoscaceres: I don't think this is required.

index.html
marcoscaceres: (I'll just emphasize, if you ever see `NoInterfaceObject` in a spec, yell at them... or tell me, and I will yell at them 📣😝)

index.html
kenchris: Done!

index.html
kenchris: src is required at least. Would you assume it is "any" if sizes is not set? Maybe file an issue?

index.html
kenchris: done

index.html
kenchris: done

index.html
kenchris: ok, fixed

index.html
kenchris: ok

index.html
kenchris: Sure :-) done...

index.html
kenchris: Any info on this tidy and how to install/run it?

index.html
kenchris: I actually think it was done this way so that you dont have cursive in the headers, which looks stupid. The next line does the dfn

index.html
kenchris: How do you want me to refer to that then. Object is defined as part of JSON, dictionary is not:

      <p>
        As the manifest uses the JSON format, this specification relies on the
        types defined in [[!ECMA-404]] specification: namely <dfn>object</dfn>,
        <dfn>array</dfn>, <dfn>number</dfn>, <dfn>string</dfn>,
        <code>true</code>, <code>false</code>, and <code>null</code>. Strict
        type checking is not enforced by this specification. Instead, each
        member's definition specifies the steps required to process a
        particular member and what to do when a type does not match what is
        expected.
      </p>

index.html
kenchris: yes I can remove <code> from inside <dfn> and <a>, doing

index.html
```diff
@@ -1261,7 +1259,7 @@ <h3 id="navigation-scope-security-considerations">
         </p>
         <p>
           A <a>manifest</a> is obtained and applied regardless of the
-          <a><code>media</code></a> attribute of the <a><code>link</code>
+          <a>media</a> attribute of the <a><code>link</code>
```
marcoscaceres: Nit: seems you missed a closing `</a>`. Also, please remove the `<code>`.

index.html
```diff
@@ -1574,7 +1572,7 @@ <h3 id="applying">
         <div>
           <p>
             Please note that the <a>start URL</a> is not necessarily the value
-            of the <a><code>start_url</code> member</a>: the user or user agent
+            of the <a data-link-for="WebAppManifest"><code>start_url</code></a> member: the user or user agent
```
marcoscaceres: Bit cleaner, you can do: 
```HTML
 <a data-tl="WebAppManifest.start_url">start_url</a>
```

index.html
```diff
@@ -1592,17 +1590,39 @@ <h3 id="applying">
         <div class="issue" data-number="446"></div>
       </section>
     </section>
-    <section>
+    <section data-dfn-for="WebAppManifest" data-link-for="WebAppManifest">
       <h2>
         Manifest and its members
```
marcoscaceres: This should change to:

```HTML
<h2>
 <dfn>WebAppManifest</dfn> dictionary 
</h2> 
```

index.html
```diff
@@ -1592,17 +1590,39 @@ <h3 id="applying">
...
       </h2>
       <p>
         A <dfn>manifest</dfn> is a JSON document that contains startup
         parameters and application defaults for when a web application is
-        launched. A manifest consists of a top-level <a>object</a> that
-        contains zero or more members. Each of the members are defined below,
-        as well as how their values are processed.
+        launched. A manifest consists of a top-level <dfn>WebAppManifest</dfn>
```
marcoscaceres: See comment above... so here ` <a>WebAppManifest</a>`

marcoscaceres: Actually,  " A manifest consists of a top-level..." is redundant... remove this sentence. The WebIDL itself already says this.

index.html
```diff
@@ -1592,17 +1590,39 @@ <h3 id="applying">
...
+        <a>dictionary</a> that contains zero or more members.
+        Each of the members are defined below, as well as how their values are processed.
       </p>
+      <pre class="idl">
```
marcoscaceres: Nit: I prefer seeing the IDL before the description... "an IDL block speaks a thousand words" 😂

index.html
```diff
@@ -1663,7 +1686,7 @@ <h3 id="applying">
           end-user, the user agent MUST use the <a>base direction</a> to
           compute directional runs and layout or position text correctly in
           text containing mixed-direction sequences [[!BIDI]]. When the <a>base
-          direction</a> is "<a data-lt="direction auto">auto</a>" the user
+          direction</a> is "<a data-link-for="TextDirectionType">auto</a>" the user
```
marcoscaceres: move `data-link-for="TextDirectionType"` inside ` <dl data-dfn-for="TextDirectionType">`

index.html
```diff
@@ -2090,15 +2121,15 @@ <h3 id="applying">
           <code>orientation</code> member
         </h3>
         <p>
-          The <dfn id="member-orientation"><code>orientation</code>
-          member</dfn> is a <a>string</a> that serves as the <a>default
+          The <dfn>orientation</dfn>
```
marcoscaceres: Nit: the `<dfn>orientation</dfn>` and so on, should all be defined inside the `h3`s... like this:

```HTML
<h3>
  <dfn>orientation</dfn> member 
</h3>
<p>The <a>orientation</a> member bla bla..</p>
```

Applies to all of the ones below...

index.html
```diff
@@ -2995,13 +3035,13 @@ <h3 id="applying">
           </pre>
         </div>
       </section>
-      <section>
+      <section data-dfn-for="imageresource">
```
```
<section data-dfn-for="ImageResource" data-link-for="ImageResource">
```

index.html
```diff
@@ -3094,18 +3134,18 @@ <h3 id="applying">
           </li>
         </ol>
       </section>
-      <section>
+      <section data-dfn-for="imageresource">
```
marcoscaceres: You shouldn't need:
```
data-dfn-for="imageresource"
```

ReSpec will find it from the parent section.

index.html
```diff
@@ -3234,41 +3274,40 @@ <h3 id="applying">
         Each member specifies which object a multi-purpose member can be used
         with.
       </p>
-      <section>
+      <section data-dfn-for="ExternalApplicationResource" data-link-for="ExternalApplicationResource">
```
marcoscaceres: `platform` should be defined in the ExternalApplicationResource section. If this is a shared definition, then we should make a note that two things share the same definition.

index.html
```diff
@@ -3472,11 +3517,20 @@ <h3 id="applying">
     </section>
     <section>
```
marcoscaceres: Add:
`data-dfn-for= "ExternalApplicationResource"` and `data-link-for="ExternalApplicationResource"`

marcoscaceres: Please make sure all sections that define IDL things have both.

index.html
kenchris: its there :-) comes a bit later

index.html
kenchris: ok!

index.html
kenchris: done

index.html
kenchris: doesnt work, paragraph is not inside <dl>

index.html
kenchris: but that makes them all cursive :(

index.html
kenchris: Maybe it should, but it doesn't ... removing this and I get one more error

index.html
kenchris: Maybe you can help me wth that in a follow up

marcoscaceres:
So... although we can define processing in IDL, we need to be careful if we are going to expose these on the web. In particular, “foo_bar” property names would make this API inconsistent with the rest of the platform. 

It’s a bit similar (but obviously not the same) as CSS in JS, where “foo-bar” becomes “fooBar”. See casing rules https://w3ctag.github.io/design-principles/#casing-rules

index.html
```diff
@@ -1768,42 +1698,34 @@ <h3 id="applying">
...
+          algorithm returns a <code>DOMString?</code>.
         </p>
         <ol>
-          <li>Let <var>value</var> be the result of calling the
-          <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
-          with argument "<code>lang</code>".
+          <li>Let <var>V</var> be <var>manifest.lang</var>.
           </li>
-          <li>If <a>Type</a>(<var>value</var>) is not "string":
+          <li>If <a>Type</a>(<var>V</var>) is <code>String</code>:
```
marcoscaceres: how come you changed these? I can live with them, but single letter attributes are not my favorite.

marcoscaceres:
🤔 overall, something doesn't feel right about how we are going about this... What I was expecting was:

 * Bytes over the wire => UTF-8 text => JSON Object => IDL Conversion => member normalization => return idlObejct. 

I need to think a bit more.

marcoscaceres:
To be bit more clear: we need to look at each member and see what kind of normalization we need to do (if any). I was expecting to see a fairly massive reduction in spec text, because at lot of the processing algorithms should go away when the do the IDL conversion.

index.html
kenchris: I like them shorter, makes it easier for me to read, plus aligned with WebIDL spec

kenchris:
I discussed some of this processing with @tobie so he might have ideas

index.html
tobie: > plus aligned with WebIDL spec

That's something I'm hoping to fix. It's a dreadful practice.

tobie:
So ["parse JSON from bytes"](https://infra.spec.whatwg.org/#json) => ES => [IDL Conversion](https://heycam.github.io/webidl/#es-any). That's easy. What you after depends on a lot of factors. What you want to work with (JSON, WebIDL, JS?), wether there's going to be JS APIs involved in the future, etc.

index.html
marcoscaceres: I agree. It gets really confusing because you always need to go back to look up what something was. Also, the upper-case thing makes it look like a constant.

kenchris: OK guys I will change it :-) "value" it is

index.html
```diff
@@ -2042,46 +1899,34 @@ <h3 id="applying">
...
         <p>
-          The <dfn id="member-display"><code>display</code> member</dfn> is a
-          <a>string</a>, whose value is one of <a>display modes values</a>. The
-          item represents the developer's preferred <a>display mode</a> for the
-          web application. When the member is missing or erroneous, the user
-          agent MUST use the <a>fallback display mode</a>.
+          The <dfn>display</dfn> member is a <a>DisplayModeType</a>, whose
+          value is one of <a>display modes values</a>. The item represents the
+          developer's preferred <a>display mode</a> for the web application.
+          When the member is missing or erroneous, the user agent MUST use the
```
mgiuca: This sentence isn't necessary; it's covered by the algorithm below (and/or the default above if one is added).

index.html
```diff
@@ -2042,46 +1899,34 @@ <h3 id="applying">
...
-              a developer warning</a> that the type is unsupported.
-              </li>
-              <li>Return the <a>fallback display mode</a>'s value.
-              </li>
-            </ol>
-          </li>
-          <li>Otherwise, <a>Trim</a>(<var>value</var>), convert to lower-case,
-          and set <var>value</var> to be the result.
+          <li>If <a>Type</a>(<var>V</var>) is <code>null</code> return the <a>
+            fallback display mode</a>'s value.
```
mgiuca: I'm confused about this (both the old and the new text, but since we're changing it, let's consider it): I think possibly this text has always been wrong, and should've said "default display mode", not "fallback display mode". Each display mode has a defined "fallback display mode" (e.g., "fullscreen"'s FDM is "standalone"), so it doesn't make sense to say "the fallback display mode" when it's undefined.

Rather it should say "the default display mode", but since that is defined as "browser", why not just say "browser"? And if you're going to say that, why not just set a default of "browser" in the IDL, and delete this entire post-processing step?

index.html
```diff
@@ -1456,95 +1453,60 @@ <h3 id="navigation-scope-security-considerations">
...
-          <li>Let <var>screenshots</var> of <var>parsed manifest</var> be the
-          result of running the <a>steps for processing an array of images</a>
+          <li>Let <var>manifest["<a>screenshots</a>"]</var> be the result of
+          running the <a>steps for post-processing ImageResource members</a>
           with <var>manifest</var>, <var>manifest URL</var>, and "screenshots"
           as arguments.
           </li>
-          <li>Return <var>parsed manifest</var>.
+          <li>Let <var>manifest["<a>related_applications</a>"]</var> be the
+          result of running the <a>steps for processing the
```
mgiuca: Why are this, and the next, ones called "processing" rather than "post-processing"?

index.html
```diff
@@ -2456,70 +2216,48 @@ <h3 id="applying">
...
-                    steps for processing the <code>platform</code> member of an
-                    application</a> with <var>potential application</var>.
-                  </li>
-                  <li>If <var>platform</var> is <code>undefined</code>, move
-                  onto the next item if any are left.
-                  </li>
-                  <li>Let <var>id</var> be the result of running the <a>steps
-                  for processing the <code>id</code> member of an
-                  application</a> with <var>potential application</var>.
+                  <li>Let <var>V'</var> be a new object created as if by the
```
mgiuca: Can this just destructively update V?

Just delete Steps 1--4, and use *V* instead of *V'*. Should be equivalent, and simpler.

index.html
```diff
@@ -2627,56 +2334,43 @@ <h3 id="applying">
...
-                    <a>Trim</a>(<var>value</var>) and convert <a>to ASCII
-                    lowercase</a>
-                  </li>
-                  <li>Append <var>potential category</var> to
-                  <var>categories</var>
-                  </li>
-                </ol>
-              </li>
-            </ol>
+          <li>Initialize <var>i</var> to be 0.
```
mgiuca: Does this have to be an index-based loop? Can you just write "For each element of *A*, *e*: Trim(*e*) and convert to ASCII lowercase"? (Optionally, "updating the value in *A*" if you are worried about someone taking it too literally.)

index.html
```diff
@@ -2627,56 +2334,43 @@ <h3 id="applying">
...
-          <code>undefined</code>
+          <li>While <var>i</var> &lt; <var>n</var>:
             <ol>
+              <li>Let <var>V</var> be the value in <var>A</var> at index <var>
+                i</var>.
+              </li>
               <li>
-                <a>Issue a developer warning</a> that the type is not
-                supported.
+                <a>Trim</a>(<var>V</var>) and convert <a>to ASCII lowercase</a>
```
mgiuca: Curious that this is the last remaining Trim(). As I said above, I think we should keep all the Trims, but if you are certain of taking them out, take this one out too.

index.html
```diff
@@ -2689,56 +2383,33 @@ <h3 id="applying">
           Manifest authors are encouraged to use lower-case.
         </p>
         <p class="note">
-          This specification does not define the particular values for a the
-          <a><code>categories</code> member</a>. However, the working group
-          maintains a <a href=
+          This specification does not define the particular values for
+          <dfn>USVString</dfn>s for a the <a>categories</a> member. However,
```
mgiuca: Drop the "a".

index.html
```diff
@@ -1456,95 +1453,60 @@ <h3 id="navigation-scope-security-considerations">
...
           </li>
-          <li>Let <var>name</var> of <var>parsed manifest</var> be the result
-          of running the <a>steps for processing the <code>name</code>
-          member</a> with <var>manifest</var> as the argument.
-          </li>
-          <li>Let <var>description</var> of <var>parsed manifest</var> be the
-          result of running the <a>steps for processing the
-          <code>description</code> member</a> with <var>manifest</var> as the
-          argument.
+          <li>Let <var>manifest["<a>start_url</a>"]</var> be the result of
```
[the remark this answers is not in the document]
marcoscaceres: Pre-processing the JSON seems really ugly, but a good fallback. Let's see what the data from your analysis says about breakage.

index.html
```diff
@@ -1456,95 +1453,60 @@ <h3 id="navigation-scope-security-considerations">
...
+          running the <a>steps for post-processing the <code>start_url</code>
+          member</a> with <var>manifest</var>, <var>manifest URL,</var> and
+          <var>document URL</var> as arguments.
           </li>
-          <li>Let <var>language</var> of <var>parsed manifest</var> be the
-          result of running the <a>steps for processing the <code>lang</code>
+          <li>Let <var>manifest["<a>display</a>"]</var> be the result of
+          running the <a>steps for post-processing the <code>display</code>
```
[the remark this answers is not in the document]
marcoscaceres: This is a really good suggestion... now that we have explicit types, we should definitely make use of them.

index.html
marcoscaceres: Agree! We get "browser" for free now 🎉

index.html
```diff
@@ -1456,95 +1453,60 @@ <h3 id="navigation-scope-security-considerations">
...
-          <li>Let <var>display mode</var> of <var>parsed manifest</var> be the
-          result of running the <a>steps for processing the
-          <code>display</code> member</a> with <var>manifest</var> as the
-          argument.
-          </li>
-          <li>Let <var>orientation</var> of <var>parsed manifest</var> be the
-          result of running the <a>steps for processing the
-          <code>orientation</code> member</a> with <var>manifest</var> and
-          <var>display mode</var> as arguments.
+          <li>Let <a>WebAppManifest</a> <var>manifest</var> be the result of
```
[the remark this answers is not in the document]
kenchris: @tobie Could you confirm?

index.html
```diff
@@ -1592,16 +1554,36 @@ <h3 id="applying">
...
     </section>
-    <section>
+    <section data-dfn-for="WebAppManifest" data-link-for="WebAppManifest">
       <h2>
-        Manifest and its members
+        <dfn>WebAppManifest</dfn> dictionary
       </h2>
+      <pre class="idl">
+          dictionary WebAppManifest {
+             TextDirectionType? dir;
```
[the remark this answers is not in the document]
kenchris: Because you don't have to have these declared in your manifest, they are optional. I was looking for optional in the web idl spec, but I see your point that they are all optional because of them being in a dict. This means that I will have to change most of the algorithms again though :)

index.html
kenchris: Well, the idea is that if a browser doesnt support "fullscreen" but "standalone" then it should render "fullscreen" as "standalone" and not "browser" - that is how it was implemented in Chrome when it didnt support fullscreen

index.html
kenchris: We could, I guess. @marcoscaceres  ?

index.html
kenchris: Nope. Actually WebIDL uses it this way and I was looking for examples on how to write it properly without doing indices. @tobie is there a preferred wording?

index.html
kenchris: I missed that one :-)

index.html
marcoscaceres: As a general rule, don't copy WebIDL's style... it's a very old spec and exhibits a ton of really bad practices (for legacy reasons).

index.html
```diff
@@ -2456,70 +2216,48 @@ <h3 id="applying">
...
+          <li>Let <var>A</var> be <var>manifest["related_applications"]</var>.
           </li>
-          <li>Let <var>unprocessed applications</var> be the result of calling
-          the <a>[[\GetOwnProperty]]</a> internal method of <var>manifest</var>
-          with argument "<code>related_applications</code>".
+          <li>Let <var>A'</var> be a new Array object created as if by the
+          expression [].
           </li>
-          <li>If <var>unprocessed applications</var> is an <a>array</a>, then:
+          <li>For <var>V</var> in <var>A</var>:
```
marcoscaceres: Nit: please use for each, IDL sequences are just arrays, so it's nice to keep the language that maps to things in JS. Also, please don't use the bracket syntax, just use:
 
```HTML
For each `<var>relatedApp</var>` in `<var>manifest</var>. <a>related_applications</a>.
```

Then we get a link back to related_applications, and don't need to guess on the type.

marcoscaceres: fixed markup above.

marcoscaceres: That translates nicely to:

```js
for(const relatedApp of manifest.related_applications) {... do stuff ... }
```

Or, better:

```js
manifest.related_applications.forEach(relatedApp => ... do stuff... )
```

index.html
marcoscaceres: Agree... I'm not sure what "`'`" signifies there.

index.html
kenchris: @tobie said that we cannot use dot notation for these as it is WebIDL, have to use []

marcoscaceres: I don’t think that’s correct. Or, if it is, I’m inclined to willfully violate that. We do the same in Payment Request. And I’ve never seen any other spec use [] notation.

kenchris: Apparently dot notation for Web IDL is not spec'ed. I went with brackets for now, but it will be easy to change (like in follow up) when dot notation gets spec'ed

index.html
kenchris: done

index.html
kenchris: fixed

index.html
kenchris: Very common in math books :-) but let me change

kenchris: ok done

index.html
```diff
@@ -1592,16 +1554,36 @@ <h3 id="applying">
...
+             DOMString? lang;
+             USVString? name;
+             USVString? short_name;
+             USVString? description;
+             sequence&lt;ImageResource&gt;? icons;
+             sequence&lt;ImageResource&gt;? screenshots;
+             sequence&lt;USVString&gt;? categories;
+             DOMString? iarc_rating_id;
+             USVString? start_url;
+             DisplayModeType? display;
```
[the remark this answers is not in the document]
kenchris: done

index.html
kenchris: done

index.html
kenchris: ok followed advice

index.html
kenchris: fixed

index.html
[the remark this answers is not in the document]
tobie: Yes, your want to convert to that particular dictionary type. Not dictionaries in general. The json var is an ES object at this point, right?

index.html
tobie: Can you open a ticket against WebIDL for that? Thanks.

index.html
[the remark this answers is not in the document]
kenchris: yes, it should be - that is at least the intent :-)

index.html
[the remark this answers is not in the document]
kenchris: fixed

index.html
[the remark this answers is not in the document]
kenchris: I think that is a good idea too, but let's make that a follow up patch.

index.html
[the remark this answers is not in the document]
kenchris: Looks like we can avoid all this with this breaking trim change

index.html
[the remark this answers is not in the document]
kenchris: done

index.html
kenchris: Done: https://github.com/heycam/webidl/issues/453

index.html
```diff
@@ -1560,24 +1558,24 @@ <h3 id="applying">
...
-             USVString? start_url;
-             DisplayModeType? display;
-             OrientationLockType? orientation;
-             USVString? theme_color;
-             USVString? background_color;
-             USVString? scope;
-             ServiceWorkerRegistration? serviceworker;
-             sequence&lt;ExternalApplicationResource&gt;? related_applications;
-             boolean? prefer_related_applications;
+             TextDirectionType dir;
```
mgiuca: `= "auto"`

index.html
```diff
@@ -1560,24 +1558,24 @@ <h3 id="applying">
...
+             DOMString lang;
+             USVString name;
```
mgiuca: Thought bubble: This should possibly be `required` (though let's not change it right now because the existing algorithm doesn't require it, and I want to change the semantics as little as possible).

index.html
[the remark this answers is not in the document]
mgiuca: Up to you, but it seems it would actually make this patch smaller if you did it now (i.e., you would delete more lines without adding any).

index.html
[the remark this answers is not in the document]
mgiuca: We've [all made this mistake before](https://github.com/WICG/web-share/issues/19) :)

index.html
mgiuca: > Well, the idea is that if a browser doesnt support "fullscreen" but "standalone" then it should render "fullscreen" as "standalone" and not "browser"

Yeah. Looks good now. To be clear, I'm not proposing we remove the concept of FDM --- that is good that it works like you describe. Just saying that there is no "global fallback display mode" (rather, that each display mode has its own FDM). It's more of a default than a fallback.

No further action required.

index.html
mgiuca: Having an *X'* as the updated value for *X* is fine. The issue here is that you went to all this trouble to create a clone which isn't necessary if you just updated it.

Again, this looks good now and no further action required.

index.html
kenchris: Ok, let me file an issue

kenchris: https://github.com/w3c/manifest/issues/622

marcoscaceres:
I'm worried about the use of enums invalidating the whole manifest... We might need to make those DOMStrings instead of enums (particularly for backwards compat in the future).

kenchris:
Yes, I agree. Didn't you have some tests we can start with?

marcoscaceres:
The tests I have are very Gecko specific. We will need to write new ones. It will be a good opportunity to review the spec tho.

PR #614, opened 2017-09-22
Removed six occurrences of superfluous spaces after quotes.

tomayac:


kenchris:
Lgrm

anssiko:

anssiko:
Thanks!

(I'm wondering whether this was a tidy bug, since it was reoccurring.)

marcoscaceres:
@anssiko, I think it was caused by a previous version of tidy (< 5.4), and then the new version didn't auto fix it. However, the manual fixes actually hold in the newer version of tidy - so these are great.

PR #625, opened 2017-10-26
Fix validator errors and warnings

anssiko:
***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/manifest/fix-validator-errors.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/manifest/9deef11...a34223a.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2707,12 +2707,13 @@
           Array&lt;<a>ImageResource</a>&gt;
         </p>
         <ol>
-          <ol>
-            <li>Let <var>imageResources</var> be a new Array object created as
-            if by the expression [].
-            </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -84,6 +84,9 @@
       license: "w3c-software-doc",
     };
     </script>
+    <style>
+      table#linking-table, table#linking-table th, table#linking-table td { border: 1px solid #000; }
```
marcoscaceres: Is there anything in base.css we could use instead? If no, then I'm ok with merging this.

anssiko: If we are not fixated on that HTML4-era table look, we can use `complex table` class and add `<thead>` and `<tbody>`. I pushed an update with a more modern look using base.css classes.

PR #634, opened 2018-01-03
Add payment handler installation URL.

rsolomakhin:
This patch adds a `payment_handler_install_url` to the webapp manifest. A user agent may use this URL for just-in-time installation of a payment handler as described in https://github.com/w3c/payment-handler/issues/240.

***
<a href="https://pr-preview.s3.amazonaws.com/rsolomakhin/manifest/pull/634.html" title="Last updated on Jan 3, 2018, 3:27 PM GMT (54099e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/634/0707d4a...rsolomakhin:54099e4.html" title="Last updated on Jan 3, 2018, 3:27 PM GMT (54099e4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1551,6 +1551,7 @@
              sequence&lt;USVString&gt; categories;
              DOMString iarc_rating_id;
              USVString start_url;
+             USVString payment_handler_install_url;
              DisplayModeType display = "browser";
              OrientationLockType orientation;
              USVString theme_color;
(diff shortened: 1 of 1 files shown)


rsolomakhin:
@marcoscaceres & @domenic: please take a look.

zkoch:
Hm, isn't there already a `serviceworker` member inside of the manifest spec? Why do we need something else?

kenchris:
There is :)

domenic:
Yeah, fair point, sorry I didn't catch that earlier. I might be missing something and there might be some reason it'd be separate, but I think that the existing serviceworker member would make sense here.

rsolomakhin:
The `payment_handler_install_url` member is the HTML page that installs the service worker and sets up the payment instruments. The `serviceworker` member does not have an equivalent.

marcoscaceres:
could the start_url just be this? I’m concerned about adding additional complexity.

rsolomakhin:
Sounds good. CC @anthonyvd on this decision. Closing this PR.

PR #644, opened 2018-01-29
Properly define "navigation scope" and "within scope".

mgiuca:
Previously, the spec did not explicitly say that you get the navigation scope of a manifest by looking at its "scope" member. Now it does. Also:
    
- All references to "navigation scope" are given a specific manifest, and
- All references to "within scope" are given a specific scope URL.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/644.html" title="Last updated on Jan 30, 2018, 7:05 AM GMT (9318e54)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/644/86b1ce5...mgiuca:9318e54.html" title="Last updated on Jan 30, 2018, 7:05 AM GMT (9318e54)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -749,7 +749,7 @@
         scope</a> algorithm.
       </p>
       <p>
-        A string <var>targetURL</var> is said the be <dfn>within scope</dfn> of
+        A string <var>targetURL</var> is said to be <dfn>within scope</dfn> of
         <a>navigation scope</a> <var>scopeURL</var> if the following algorithm
         returns <code>true</code>:
       </p>
--- a/index.html
+++ b/index.html
@@ -741,12 +741,13 @@
       <h2>
         Navigation scope
       </h2>
-      <p>
+      <p data-link-for="WebAppManifest">
         A <dfn>navigation scope</dfn> is a [[!URL]] that represents the set of
         URLs to which an <a>application context</a> can be navigated while the
-        manifest is <a>applied</a>. To determine if a URL is within the
-        <a>navigation scope</a>, the user agent MUST run the <a>within
-        scope</a> algorithm.
+        manifest is <a>applied</a>. The <a>navigation scope</a> of a manifest
+        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"]. If this is
+        <code>undefined</code> (because of an error or it was not declared in
+        the manifest), the navigation scope is <dfn>unbounded</dfn>.
       </p>
       <p>
         A string <var>targetURL</var> is said to be <dfn>within scope</dfn> of
@@ -755,8 +756,7 @@
       </p>
       <ol>
         <li>If <var>scopeURL</var> is <code>undefined</code> (i.e., it is
-        <a>unbounded</a> because of an error or it was not declared in the
-        manifest), return <code>true</code>.
+        <a>unbounded</a>), return <code>true</code>.
         </li>
         <li>Let <var>target</var> be a new URL using <var>targetURL</var> as
         input. If <var>target</var> is failure, return <code>false</code>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -741,12 +741,13 @@ <h4>
...
-      <p>
+      <p data-link-for="WebAppManifest">
         A <dfn>navigation scope</dfn> is a [[!URL]] that represents the set of
         URLs to which an <a>application context</a> can be navigated while the
-        manifest is <a>applied</a>. To determine if a URL is within the
-        <a>navigation scope</a>, the user agent MUST run the <a>within
-        scope</a> algorithm.
+        manifest is <a>applied</a>. The <a>navigation scope</a> of a manifest
+        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"]. If this is
+        <code>undefined</code> (because of an error or it was not declared in
```
kenchris: due to an error, or because it was not declared in the manifest

index.html
```diff
@@ -795,18 +795,20 @@ <h2>
...
-        the navigation scope, then the user agent MUST behave as if the
-        application context is not <a>allowed to navigate</a> (this provides
-        the ability for the user agent to perform the navigation in a different
-        browsing context, or in a different user agent entirely). If during the
-        handle redirects step of HTML's <a>navigate algorithm</a> the redirect
-        URL is not <a>within scope</a>, abort HTML's navigation algorithm with
-        a <code>SecurityError</code>.
+        the navigation scope of the application context's manifest, then the
+        user agent MUST behave as if the application context is not <a>allowed
+        to navigate</a> (this provides the ability for the user agent to
```
kenchris: I am wondering whether we should do the () part without ()'s and with uppercase This

index.html
```diff
@@ -795,18 +795,20 @@ <h2>
...
+        perform the navigation in a different browsing context, or in a
+        different user agent entirely). If during the handle redirects step of
```
kenchris: Can we link to 'handle redirects step'  definition

index.html
```diff
@@ -1800,7 +1801,7 @@ <h3>
             <ol>
               <li>
                 <a>Issue a developer warning</a> that the start URL is not
-                <a>within scope</a> of the navigation scope.
+                <a>within scope</a> of the scope URL.
```
kenchris: scope URL should either be a variable or a link here, no?

kenchris:
I am fine with this, good to make this better defined. I added a few comments though

index.html
mgiuca: Done, but note I am about to rewrite this sentence in the next PR (changing default scope) anyway.

index.html
mgiuca: I can't find it anywhere. The navigate algorithm doesn't seem to deal with redirects at all; that is taken care of by the HTTP fetch algorithm (in the fetch spec).

Looks like things have been refactored. I think this change is out of scope from this PR, especially since I plan to rework things shortly in this area.

index.html

index.html
mgiuca: Not really, because this is the pseudo-suggested text of the developer warning, not part of the algorithm itself. Same as how "the start URL" is already referred to here without variable or link.

It's kind of just suggesting that the user agent show a warning that reads "The start URL is not within scope of the scope URL". I even thought about removing the link to "within scope" for the same reason; this isn't a reference to the "within scope" concept. It's just suggested text.

mgiuca:
Thanks. I'll wait for Kenneth to have another look, then hopefully merge tomorrow.

(It's annoying how GitHub doesn't really let you review a PR based on top of another outstanding PR, so I can't upload my next patch until this is landed.)

kenchris:
lgtm

PR #645, opened 2018-01-30
Fix up start_url and scope processing algorithm.

mgiuca:
There was a mix-up of the variable names.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/645.html" title="Last updated on Jan 30, 2018, 6:29 AM GMT (1bb81a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/645/88f42e1...mgiuca:1bb81a0.html" title="Last updated on Jan 30, 2018, 6:29 AM GMT (1bb81a0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1758,19 +1758,19 @@
         </p>
         <p>
           The steps for <dfn>processing the <code>scope</code> member</dfn> is
-          given by the following algorithm. The algorithm takes an
-          <a>USVString</a> <var>scope URL</var>, a <a>URL</a> <var>manifest
+          given by the following algorithm. The algorithm takes a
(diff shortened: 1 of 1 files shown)


mgiuca:
Hi @marcoscaceres, can you have a quick look at this (best to get this small change out the way so it doesn't show up in the diff of my big PR to change unbounded scope).

In the case of `start_url` it's just plain wrong (it references a variable "value" that's never defined). But in both cases, I think we do want to call this variable "value" rather than overloading/shadowing the variable name "start URL" or "scope URL", which would otherwise start off being a string and change type to URL part-way through the algorithm.

PR #647, opened 2018-01-31
[Breaking] "scope" member now defaults to parent of "start_url"

### LINKED ISSUE #550: Make unbounded navigation scope less broad
Check: https://bugs.chromium.org/p/chromium/issues/detail?id=688888

We don't necessarily want, say ```"start_url": "https://cnet.com/technology/"``` turn all of cnet.com into the PWA because scope is undeclared. We should probably restrict to closest surdir.

mgiuca:
This removes the concept of "unbounded scope", which was special behavior if "scope" was omitted. Now, if "scope" is omitted, it defaults to the parent directory of "start_url", which is the most sensible default.

This is a breaking normative change. Sites relying on the old behavior may find that the scope is now more restrictive. In particular, if the "start_url" is deep within the path hierarchy, the rest of the site may no longer be considered part of the app. Having said that, a major implementation (Chrome for Android) already had this policy, so this is expected to bring the spec in-line with current expectations, rather than break existing sites.

Closes #550.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/647.html" title="Last updated on Jan 31, 2018, 5:02 AM GMT (a665904)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/647/326c22e...mgiuca:a665904.html" title="Last updated on Jan 31, 2018, 5:02 AM GMT (a665904)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -747,6 +747,23 @@
         manifest is <a>applied</a>. The <a>navigation scope</a> of a manifest
         <var>manifest</var> is <var>manifest</var>["<a>scope</a>"].
       </p>
+      <div class="note" data-link-for="WebAppManifest">
+        <p>
+          If the <a>scope</a> member is not present in the manifest, it
+          defaults to the directory containing the <a>start_url</a> member. For
+          example, if <a>start_url</a> is <code>/pages/welcome.html</code>, and
+          <a>scope</a> is missing, the navigation scope will be
+          <code>/pages/</code> on the same origin. If <a>start_url</a> is
+          <code>/pages/</code> (the trailing slash is important!), the
+          navigation scope will be <code>/pages/</code>.
+        </p>
+        <p>
+          Developers should take care, if they rely on the default behaviour,
+          that all of the application's pages are in the same directory (or a
+          subdirectory) as the start URL. To be safe, explicitly specify
+          <a>scope</a>.
+        </p>
+      </div>
       <p>
         A string <var>targetURL</var> is said to be <dfn>within scope</dfn> of
         <a>navigation scope</a> <var>scopeURL</var> if the following algorithm
@@ -799,10 +816,6 @@
         scope</a> of the navigation scope of the application context's
         manifest, abort HTML's navigation algorithm with a
         <code>SecurityError</code>.
-      </p>
-      <p class="note" data-link-for="WebAppManifest">
-        A developer specifies the navigation scope via the <a>scope</a> member
-        (defaulting to the parent path of the <a>start_url</a> member).
       </p>
       <section data-link-for="DisplayModeType">
         <h3 id="navigation-scope-security-considerations">
--- a/index.html
+++ b/index.html
@@ -1762,12 +1762,14 @@
           given by the following algorithm. The algorithm takes a
           <a>USVString</a> <var>value</var>, a <a>URL</a> <var>manifest
           URL</var>, a <a>URL</a> <var>document URL</var>, and a URL <var>start
-          URL</var>. This algorithm returns a <a>URL</a> or
-          <code>undefined</code>.
+          URL</var>. This algorithm returns a <a>URL</a>.
         </p>
         <ol>
+          <li>Let <var>default</var> be a new <a>URL</a> using "." as
+          <var>input</var> and <var>start URL</var> as the <var>base</var> URL.
+          </li>
           <li>If <var>value</var> is the empty string, then return
-          <code>undefined</code>.
+          <var>default</var>.
           </li>
           <li>Let <var>scope URL</var> be a new <a>URL</a> using
           <var>value</var> as <var>input</var> and <var>manifest URL</var> as
@@ -1778,7 +1780,7 @@
               <li>
                 <a>Issue a developer warning</a>.
               </li>
-              <li>Return <code>undefined</code>.
+              <li>Return <var>default</var>.
               </li>
             </ul>
           </li>
@@ -1793,7 +1795,7 @@
                 same-origin</a> as <code>Document</code> of the <a>application
                 context</a>.
               </li>
-              <li>Return <code>undefined</code>.
+              <li>Return <var>default</var>.
               </li>
             </ol>
           </li>
@@ -1803,13 +1805,18 @@
                 <a>Issue a developer warning</a> that the start URL is not
                 <a>within scope</a> of the scope URL.
               </li>
-              <li>Return <code>undefined</code>.
+              <li>Return <var>default</var>.
               </li>
             </ol>
           </li>
           <li>Otherwise, return <var>scope URL</var>.
           </li>
         </ol>
+        <div class="note">
+          The default scope (if <code>scope</code> is omitted or an error) is
+          the <var>start URL</var>, with its filename, query, and fragment
+          removed.
+        </div>
       </section>
       <section>
         <h3>
--- a/index.html
+++ b/index.html
@@ -745,9 +745,7 @@
         A <dfn>navigation scope</dfn> is a [[!URL]] that represents the set of
         URLs to which an <a>application context</a> can be navigated while the
         manifest is <a>applied</a>. The <a>navigation scope</a> of a manifest
-        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"]. If this is
-        <code>undefined</code> (due to an error, or because it was not declared
-        in the manifest), the navigation scope is <dfn>unbounded</dfn>.
+        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"].
       </p>
       <p>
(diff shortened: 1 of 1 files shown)


mgiuca:
Anybody know why pr-preview hasn't kicked in yet? (There is no 647 page on pr-preview.s3.amazonaws.com.) I was hoping to be able to share this around with a link to the nicely formatted Diff page.

marcoscaceres:
That's odd... no idea why it failed. Maybe send again?

index.html
```diff
@@ -745,19 +745,31 @@ <h2>
...
         manifest is <a>applied</a>. The <a>navigation scope</a> of a manifest
-        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"]. If this is
-        <code>undefined</code> (due to an error, or because it was not declared
-        in the manifest), the navigation scope is <dfn>unbounded</dfn>.
+        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"].
       </p>
+      <div class="note" data-link-for="WebAppManifest">
+        <p>
+          If the <a>scope</a> member is not present in the manifest, it
+          defaults to the directory containing the <a>start_url</a> member. For
```
marcoscaceres: The problem is that there is not such thing as a "directory" in URLs (I know you mean path segments here)... it's still somewhat problematic, because the path segments are kinda meaningless or used to mean different things (unlike file system directories).

mgiuca: I know, I was trying to avoid using the word "directory" but I just ran with it. Only in non-normative text. I can change it but do you have any suggested wording?

> it's still somewhat problematic

What do you mean? The wording is problematic, or the concept? I think the concept of cutting off the last part of the path is very well defined on the web (it's the standard URL resolution algorithm). The *reason* resolution works this way is because it allows HTML pages to reference a file in the same directory just by name, which is why I described it this way.

mgiuca:
> That's odd... no idea why it failed. Maybe send again?

Hrr, GitHub doesn't let me create a new PR when one already exists. Also you've started commenting already :)

A shame, because the whole reason I waited until today to send the PR is so that the pr-preview Diff would look right.

index.html
marcoscaceres: >  I can change it but do you have any suggested wording?

We should follow terminology from the URL spec (i.e., path segment). 
 
> What do you mean? The wording is problematic, or the concept? 

Apologies, I accidentally missed this line: 
https://github.com/w3c/manifest/pull/647/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR1742

Which addresses the issue. There too, we might just point to the URL spec "single-dot path segment".

marcoscaceres:
As this is a significant change, we'd naturally want to get an ok from a few other implementers before we merge. 

@tobie, sorry to bother you, but for some reason we didn't get the preview generated? Any chance you could check what happened?

marcoscaceres:
@mgiuca, it MAY regenerate the pr-preview once you push another commit (I confirmed it's working, as you probably saw my "test" pull request). You could push an empty commit even, and then we can squash it later.

index.html
mgiuca: OK. I didn't want to s/directory/path segment, because a path segment is just one piece, not the whole prefix. (For example, if the path is `/x/y/z.html`, the scope is `/x/y`, not just `y`.)

I've reworded it to refer to path terminology, not directory.

> There too, we might just point to the URL spec "single-dot path segment".

Not sure what you mean. Do you think the actual algorithm should say "single-dot path segment" instead of "."? I don't think so, because "." is a specific string, while "single-dot path segment" is a pattern match.

marcoscaceres:
Excellent, thanks for also checking the URL stuff. Let’s get this in front of the various implementers and hopefully get consensus quickly.

mgiuca:
> Overall, with a few nits on terminology, this seems reasonable to me. I'd be happy to change have us change Gecko behaviour to match.

Is the Gecko behaviour in-line with the spec as it stands now? (i.e., you never navigate away from the app if the scope is unspecified). It would be good to make sure Mozilla engineering is happy to make this change before we commit to the spec change.

I have also contacted Apple to discuss, since Safari has shipped this in a public beta.

index.html
```diff
@@ -745,19 +745,30 @@ <h2>
...
       </p>
+      <div class="note" data-link-for="WebAppManifest">
+        <p>
+          If the <a>scope</a> member is not present in the manifest, it
+          defaults to the parent path of the <a>start_url</a> member. For
+          example, if <a>start_url</a> is <code>/pages/welcome.html</code>, and
+          <a>scope</a> is missing, the navigation scope will be
+          <code>/pages/</code> on the same origin. If <a>start_url</a> is
+          <code>/pages/</code> (the trailing slash is important!), the
+          navigation scope will be <code>/pages/</code>.
```
kenchris: So basically something like a base path or so. Isn't this defined somewhere?

If not, why not define it and create an algorithm, basically:

`base = (url) => url.substr(0, url.lastIndexOf('/') + 1)`

kenchris: Also what happens with `/hi/there/../cool/youcouldmakeit/'

index.html
```diff
@@ -745,19 +745,30 @@ <h2>
...
+        </p>
+        <p>
+          Developers should take care, if they rely on the default behaviour,
+          that all of the application's page URLs begin with the parent path of
```
kenchris: Even better if you could like to the "parent path" algorithm here

kenchris:
@mgiuca talk to Andreas Bovens from Mozilla - he is the PR.

index.html
mgiuca: > So basically something like a base path or so. Isn't this defined somewhere?

Yes, in the [URL standard](https://url.spec.whatwg.org/#url-parsing). This section is a non-normative explanation with examples, for readers who want to understand practically what the default scope will be, without going into the URL parsing algorithm (which is practically unreadable).

The real (normative) text is in the ["scope" member](https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/647.html#scope-member) section, where the default is defined as "a new URL using "." as input and *start URL* as the *base URL*". That invokes the [URL parser](https://url.spec.whatwg.org/#url-parsing) which then uses the standard resolution algorithm, the same as if you had `<a href=".">` to link to the containing directory of the current page. I did that because I don't think duplicating the resolution algorithm is a good idea, but it's a little arcane, hence the non-normative note.

If you want to explicitly define the logic here, instead of using the existing URL resolution algorithm, you need to:

1. Drop the query and fragment.
2. Drop everything after the last slash in the path.

The algorithm needs to operate on a URL object (not a string), so the path is represented as a list of segments, implicitly slash-separated. You need to be careful that the path ends with a blank segment, to represent a trailing slash. The algorithm would be this:

> 1. Let default be a new URL.
> 2. Set default’s username to start URL's username, default’s password to start URL's password, default’s host to start URL's host, default’s port to start URL's port, default’s path to a copy of start URL's path.
> 3. Remove default’s path’s last item, if any.
> 4. Append the empty string to default's path.

(This is basically a specialization of the URL Parser with *input* hard-coded as ".").

But I would rather not duplicate the existing logic.

(Aside: One of my main gripes about the URL standard is that almost the entire thing is a monolithic token-consume parser for the entire URL, rather than being broken into distinct concepts that you can easily point to, like a "resolve" algorithm which is what we need here. Formally, there is no such thing as a "relative URL" and there is no such thing as "resolution"; it's all just built-in functionality of the URL parser, which makes it hard to reference from outside, as we need to do here.)

index.html
mgiuca: Is this still needed, given the above? Technically, I do, because if you click on "scope" in that sentence, it links to the parent path algorithm defined in [scope member](https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/647.html#scope-member).

mgiuca:
@marcoscaceres Can you ping Andreas on this (or whoever is most appropriate on the product or eng side; i.e., who would be an approver for changing the implementation).

index.html
kenchris: I wonder if @annevk has some comments for this?

annevk: The language about "new URL" isn't quite correct. You actually need to invoke the URL parser. Otherwise I tend to agree with @mgiuca on the approach.

mgiuca: I agree with @annevk but I was trying to be consistent with the writing in the rest of this document (in fact on the very next line: "Let scope URL be a new URL using value as input and manifest URL as the base URL."). There are six existing instances of this "new URL using X as input" in the Manifest spec.

I'll do a pass over it afterwards and fix up the invocations of URL parsing, but I think it's more urgent to land this to keep implementation divergence from getting out of control.

kenchris: I am fine with this, do you want me to land this now?

kenchris:
OK, already approved by Marcos, so let's land this! :100:

Issue #636, opened 2018-01-09
Use case for appuninstall event


rmacklin:
I found a brief discussion in this issue #417 where an _uninstall_ event was discussed and deemed not useful:

> Regarding uninstall, I'm not sure this event makes a lot of sense

- https://github.com/w3c/manifest/issues/417#issuecomment-165773500

> You are probably right that uninstall is not that useful.

- https://github.com/w3c/manifest/issues/417#issuecomment-165776611

Along a similar line, I found this comment in issue #161:

> The architectural model that we've chosen for installable web apps is to not let the developer know if the app is installed or not (they should not care).

- https://github.com/w3c/manifest/issues/161#issuecomment-73245397

I'd like to offer a counterexample to these opinions, where I think it makes sense for the developer to care if the app is installed, and where an uninstall event would be useful.

Consider an app where private messages are shared between users. Because these messages can be considered sensitive data, the developer may decide it's better not to store these in IndexedDB for "regular" visitors. That way, if a user visited the app on a shared computer (e.g. at a library), their messages wouldn't be visible by another person who later inspected the IndexedDB contents in the browser developer tools. [1]

But when a user installs this app (either manually or in response to an install prompt), that would be a good indicator that the app can enable local data storage in IndexedDB. Users generally know, and even expect, that apps they install can store data locally for offline use. As a result, it seems unlikely that a user would install a web app on a shared computer; they're likely to only do this on more "trusted"/personal devices. In this way, the act of installing the web app can be viewed as the user consenting/opting in to local data storage. Enabling IndexedDB at this time seems natural and safe, and the `appinstall` event would allow this.

However, when a user uninstalls an app, they expect the data stored by that app to be deleted. An `appuninstall` event would allow the developer to hook into that event and clear their IndexedDB storage at that time.

[1] <sup> Yes, you could consider always storing the sensitive data in IndexedDB, and clearing it when the user logs out. However, I think that would only work if the user physically logs out on that computer. And that's probably not a safe assumption: Some users may have automatic session expiry (which doesn't require them to manually log out), or they may log out via other means (e.g. log back in on another computer and select to "log me out elsewhere"; in these cases the other computer's session would be invalidated, but it would not be possible to clear the leftover IndexedDB storage). So it still seems like not using IndexedDB for "regular" visitors is a reasonable choice from a privacy/security standpoint.</sup>

valioDOTch:
I don't see why this would be a good use-case.
[...]

PR #678, opened 2018-05-22
Added an Uninstallation section.

mgiuca:
The section recommends other cleanup activities besides removing the
application.

Addresses #636.


tomayac:
@mgiuca Could your `RECOMMENDED` become a stronger `SHOULD`, reflecting user expectations?

mgiuca:
@tomayac RECOMMENDED and SHOULD have the same meaning: https://www.ietf.org/rfc/rfc2119.txt

tomayac:
Wow, #TIL. To me, if something "should" be a certain way sounds stronger than if something is "recommended" to be a certain way. Luckily the [RFC](https://www.ietf.org/rfc/rfc2119.txt) has us non-native speakers covered. Thanks for teaching me. Sorry for the noise.

index.html
```diff
@@ -526,6 +526,21 @@ <h3>
           can be <a>installed</a> is left to implementers.
         </p>
       </section>
+      <section>
+        <h3>
+          Uninstallation
+        </h3>
+        <p>
+          User agents SHOULD provide a mechanism for the user to remove the
+          installed app from the system.
```
marcoscaceres: "from the system" might be overly specific... a UA can "install" a web app to, for example, the new tab page.

kenchris: Maybe just drop that part then... apart from that lgtm

PR #687, opened 2018-06-04
 Add note on back button requirements

### LINKED ISSUE #673: Display mode "standalone" inconsistently handled by browsers
For display mode `"standalone"`, the [spec](https://www.w3.org/TR/appmanifest/#dfn-display-mode) has the following to say (emphasis mine):

*"Opens the web application to look and feel like a standalone native application. This can include the application having a different window, its own icon in the application launcher, etc. In this mode, the user agent will exclude standard browser UI elements such as an URL bar, but **can include other system UI elements such as a status bar and/or system back button**."*

Now of course Edge and Chrome OS handle this differently **on desktop**, where Edge would [show a back button](https://www.youtube.com/watch?v=6fb-t9ffDvo&feature=youtu.be&t=745) (courtesy of @boyofgreen), and [Chrome OS wouldn't show a back button](https://cdn-images-1.medium.com/max/1200/1*YdAiADb46bjX9O6wNryhsg.jpeg) (courtesy of @kenchris).

**On mobile**, the general idea seems to be to not show any browser UI at all, which causes navigation issues for devices without a dedicated system ("hardware") back button as on Android, for example [in iOS](https://medium.com/@firt/progressive-web-apps-on-ios-are-here-d00430dee3a7#1edc) (see the screenshot, courtesy of @firtman).

The spec should probably add more guidance as to when to show a back button in `"standalone"` mode and when not; basing the decision on the presence or absence of a system ("hardware") back button seems like a reasonable default.

tomayac:
This closes #673.

***
<a href="https://pr-preview.s3.amazonaws.com/tomayac/manifest/pull/687.html" title="Last updated on Jun 5, 2018, 11:27 AM GMT (cc01b11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/687/0997b5a...tomayac:cc01b11.html" title="Last updated on Jun 5, 2018, 11:27 AM GMT (cc01b11)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -937,8 +937,8 @@
           <code>standalone</code> means that the application is <em>fully</em>
           responsible for handling its in-app navigation. This means that
           on platforms without a <em>system-level</em> back button the
-          application needs to make an <em>application-level</em> back button
-          available that respects the platform's user experience norms.
+          application needs to provide an <em>application-level</em> back button
+          that respects the platform's user experience norms.
         </p>
       </div>
       <p>
--- a/index.html
+++ b/index.html
@@ -936,9 +936,10 @@
           Setting the display mode to <code>fullscreen</code> or
           <code>standalone</code> means that the application is <em>fully</em>
           responsible for handling its in-app navigation. This means that
-          on platforms without a <em>system-level</em> back button the
-          application needs to provide an <em>application-level</em> back button
-          that respects the platform's user experience norms.
+          on platforms without a <em>system-level</em> back button and in cases
+          where backward navigation is necessary, the application needs to provide
+          an <em>application-level</em> back button that respects the platform's
+          user experience norms.
         </p>
       </div>
       <p>
--- a/index.html
+++ b/index.html
@@ -931,15 +931,15 @@
           ultimately falls back to <code>minimal-ui</code>.
         </p>
       </div>
-       <div class="note">
-          <p>
-            Setting the display mode to <code>fullscreen</code> or
-            <code>standalone</code> means that the application is <em>fully</em>
-            responsible for handling its in-app navigation. This means that
-            on platforms without a <em>system-level</em> back button the
-            application needs to make an <em>application-level</em> back button
-            available that respects the platform's user experience norms.
-          </p>
+      <div class="note">
+        <p>
+          Setting the display mode to <code>fullscreen</code> or
+          <code>standalone</code> means that the application is <em>fully</em>
+          responsible for handling its in-app navigation. This means that
+          on platforms without a <em>system-level</em> back button the
+          application needs to make an <em>application-level</em> back button
+          available that respects the platform's user experience norms.
+        </p>
       </div>
       <p>
         The <dfn>display modes values</dfn> defined by
--- a/index.html
+++ b/index.html
@@ -931,6 +931,16 @@
           ultimately falls back to <code>minimal-ui</code>.
         </p>
       </div>
+       <div class="note">
+          <p>
+            Setting the display mode to <code>fullscreen</code> or
+            <code>standalone</code> means that the application is <em>fully</em>
+            responsible for handling its in-app navigation. This means that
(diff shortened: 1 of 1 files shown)


kenchris:
LGTM, @marcoscaceres ?

marcoscaceres:
Seems ok, but I'm worried about getting into this game, because these things change as new browser versions get released. @dominickng, thoughts here?

dominickng:
Not every app needs a back button, so the _needs_ phrasing in this PR feels quite wrong.

More generally, PWAs should function without the crutch of system-level UI at all. If that means they need to include a back button, they should, and if they don't need one, they shouldn't provide one.

I think there's a bigger and more complex discussion about _duplicate_ back buttons as mentioned on #673 - when both the platform and app provide one. We could introduce new members to deal with this, and I believe @mgiuca is thinking about that. But this change doesn't do anything for that, and seems to amount to advising developers to use UA-detection for their UI which isn't good.

tomayac:
@dominickng Agree that not all apps need a back button, see https://github.com/w3c/manifest/pull/687/commits/cc01b115278bdbddb1ff15c348445561ff268670.

Also agree that double back buttons are a horrible experience and that user-agent sniffing can't be the answer.

Here's yet another issue for Twitter's desktop PWA ([on Chrome](https://developers.google.com/web/updates/2018/05/dpwa), here running on macOS), notice how it disables the back button based on window width:
 
screen shot 2018-06-05 at 13 29 23

screen shot 2018-06-05 at 13 29 19

I really think more guidance is required.

jackpalmer1:
@tomayac if you look in the video at https://youtu.be/6fb-t9ffDvo?t=762
I'm not sure why Microsoft is not opening the app actually in full screen. For games, it makes a lot of sense to open directly in fullscreen.

Here a problem arises. For Chrome, full screen is properly interpreted as taking up entire display area, which follows the spec. For Microsoft, it takes up only the entire area of the window effectively becoming standalone.

If my app manages the back button its self, I would not want the OS to manage it.
To implement the correct behavior across both browsers, when I detect the useragent as Chrome, in the PWA, I'll have to send manifest with the display property as standalone. When the useragent is Edge, I'd have to send another version of the manifest that has display property as fullscreen to give an expected experience.

Can we not make the spec strict so that if a developer says something the browser actually has to follow it, or at-least offer a strict version such as fullscreen-strict or standalone-strict. This hacky way I am having to do to provide a consistent with Edge is going to become a maintenance nightmare.

And maybe we should document the default behavior for all 3 major browsers for all properties (Chrome, Edge and Safari) because even I am confused despite reading all the documentation.

dominickng:
@tomayac I don't think this PR does anything for the double back button case. Twitter's hack is assuming that wide windows = desktop browser with a back button, and narrow windows = mobile browser without a back button. It's not a strictly correct assumption, but we

a) haven't giving developers better tools to work this out,
b) haven't shipped desktop PWAs on Mac yet
c) and it's a pretty reasonable assumption that users won't access mobile.twitter.com on desktop.

Note that what Twitter are doing here is exactly following the guidance proposed in the PR - try and work out what system UI is present and adjust your UX accordingly. So I'm not sure how adding the guidance is going to prevent the "issue" they have. While the need requirement has been softened, the PR is still essentially telling developers to use UA-detection. :)

Like I mentioned before, we will need a more active mechanism to help with multiple back buttons that's much broader than the fragment proposed here - and that is being fleshed out right now by @mgiuca. So I'm still not convinced we need this particular PR accepted - what we actually want to do is a bigger change to fix the actual problem being discussed - when do developers need their own back buttons.

@jackpalmer1 please don't duplicate post, especially since your comment here isn't relevant to the pull request under consideration. :)

tomayac:
Google-internally we agreed that this is an issue, but purely adding this note won't fix it. We will work on something more actionable and get back. Meanwhile, I'm closing the Pull Request.

PR #701, opened 2018-07-05
Breaking: Change requirement to block off-scope navigation.

### LINKED ISSUE #646: Clarify rule of navigating away from the navigation scope
Briefly discussed in #550 but spinning out into its own issue / PR.

@marcoscaceres @kenchris I think we want to rework the rule about navigating away from the navigation scope (as part of my work in #550 to remove the concept of "unbounded" scope).

There are currently two major cases: bounded scope (explicit `scope` in manifest) and unbounded scope (no `scope` in manifest). In the bounded case, you aren't allowed to navigate outside of scope at all, except the user agent is allowed to spawn a new top-level browsing context. In the unbounded case, you *are* allowed to navigate outside of the origin, but there is a recommendation that the UA show the location bar when off-origin (or just open a browser tab).

It makes no sense for these cases to be treated differently (especially as we move away from unbounded scope). But I think the right place to be is somewhere in the middle.

- It isn't clear to me what the practical UI implications of "a new top-level browsing context" means.
- A basic reading of the text suggests that if you have bounded scope, off-scope navigations must either fail, or bounce the user out to a new browser tab. Whereas, if you have an unbounded scope, in the event of an off-origin navigation, the UA may either:
  - a) Continue navigating within the app window, or
  - b) Stay within the app window but show some UI indicating that you're off-origin, along with some location bar, or
  - c) Bounce the user out to a new browser tab.

I think, regardless of bounded vs unbounded, (a) should be banned (because it is insecure, as the user will effectively be browsing the web with no knowledge of where they are), and (b) and (c) should be allowed. On Chrome for Android, we do (b) for both bounded and unbounded scope. Are we violating the spec? Or can we justify it in that showing the UI indicating you are off-origin counts as a "new top-level browsing context" (just not a normal browser window). On Chrome for Desktop, we plan to do (c).

I think the *normative* part of the spec (that applies to bounded scope; after #550 this will always apply) should be clear that (b) or (c) are both permitted.

Further problems. In the bounded case, the paragraph under §4 applies:

> The user agent MUST navigate the application context as per [HTML]'s [navigate algorithm](https://html.spec.whatwg.org/#navigate) with exceptions enabled. If the URL of the resource being loaded in the navigation is not within scope of the navigation scope of the application context's manifest, then the user agent MUST behave as if the application context is not allowed to navigate. This provides the ability for the user agent to perform the navigation in a different browsing context, or in a different user agent entirely. If during the handle redirects step of HTML's navigate algorithm the redirect URL is not within scope of the navigation scope of the application context's manifest, abort HTML's navigation algorithm with a `SecurityError`.

This has a few problems:

1. As discussed on #644, there is no "handle redirects" step of HTML's navigate algorithm. Not sure how to fix this.
2. "with exceptions enabled". I'm not sure why this is here. Turning on exceptions makes navigation automatically fail with a `SecurityError` if it's not allowed to navigate (i.e., if it goes out of scope). That nullifies the next sentence: "This provides the ability for the user agent to perform the navigation in a different browsing context". No, it doesn't; if exceptions were disabled, then the navigation algorithm explicitly allows the UA to open in a new top-level browsing context. So I think "with exceptions enabled" should be removed, and similarly, the text for aborting with `SecurityError` should be reworked.

I think we just need some language that says the UA MUST create a new top-level browsing context when navigating off-scope, and somehow explain that this doesn't have to be a normal browser window; it can be within the application window but has to show the location bar (or does that language have to be non-normative?)

mgiuca:
Previously, user agents were required to prevent off-scope navigation (or open it in a new top-level browsing context). Now, they are not allowed to do this, as it broke existing sites not designed to be run under these conditions.

Replaced with a recommendation to show the URL when off-scope.

Closes #646.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/701.html" title="Last updated on Jul 10, 2018, 3:01 AM GMT (ad78efa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/701/cc53c75...mgiuca:ad78efa.html" title="Last updated on Jul 10, 2018, 3:01 AM GMT (ad78efa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -818,12 +818,12 @@
         "!HTML#active-document">active document</a>'s <a data-cite=
         "!DOM#concept-document-url">URL</a> is not <a>within scope</a> of the
         navigation scope of the <a>application context</a>'s manifest, the user
-        agent SHOULD show prominent UI indicating the <a data-cite=
+        agent SHOULD show a prominent UI element indicating the <a data-cite=
         "!DOM#concept-document-url">document URL</a>, or at least its
         <a>origin</a>, including whether it is served over a secure connection.
         This UI SHOULD differ from any UI used when the <a data-cite=
         "!DOM#concept-document-url">document URL</a> is <a>within scope</a>, in
-        order to alert the user of the off-scope navigation.
+        order to make it obvious that the user is navigating off scope.
       </p>
       <div class="note">
         <p>
@@ -846,10 +846,10 @@
           Security considerations
         </h3>
         <p>
-          The above recommendation to show UI when the <a>application
-          context</a> is navigated to a <a>URL</a> that is not <a>within
-          scope</a> of the manifest is for security. It ensures that users are
-          always aware of which <a>origin</a> they are interacting with.
+          The above recommendation (to show some UI when the <a>application
+          context</a> is navigated to an out-of-scope <a>URL</a>) is for
+          security reasons. It ensures that users are always aware of which
+          <a>origin</a> they are interacting with.
         </p>
       </section>
       <section>
--- a/index.html
+++ b/index.html
@@ -836,9 +836,8 @@
           Unlike previous versions of this specification, user agents are no
           longer required or allowed to block off-scope navigations, or open
           them in a new <a>top-level browsing context</a>. This practice broke
-          a lot of sites that navigate to a URL on another origin (e.g., to
-          perform third-party authentication), which would then navigate back
-          to the originating site. See <a href=
+          some sites that navigate to an off-scope URL (e.g., to perform
+          third-party authentication). See <a href=
           "https://github.com/w3c/manifest/issues/646">Issue 646</a>.
         </p>
       </div>
--- a/index.html
+++ b/index.html
@@ -813,50 +813,44 @@
         "!SERVICE-WORKERS-1#scope-match-algorithm">Service Workers</a>. To
         avoid unexpected behavior, use a scope ending in a <code>/</code>.
       </div>
-      <div class="issue" title="🐒 patch">
+      <p>
+        If the <a>application context</a>'s <a data-cite=
+        "!HTML#active-document">active document</a>'s <a data-cite=
+        "!DOM#concept-document-url">URL</a> is not <a>within scope</a> of the
+        navigation scope of the <a>application context</a>'s manifest, the user
+        agent SHOULD show prominent UI indicating the <a data-cite=
+        "!DOM#concept-document-url">document URL</a>, or at least its
+        <a>origin</a>, including whether it is served over a secure connection.
+        This UI SHOULD differ from any UI used when the <a data-cite=
+        "!DOM#concept-document-url">document URL</a> is <a>within scope</a>, in
+        order to alert the user of the off-scope navigation.
+      </p>
+      <div class="note">
         <p>
-          Enforcing the navigation scope depends on [[!HTML]]'s navigate
-          algorithm. As such, the following monkey patches [[!HTML]]. <a href=
-          "https://www.w3.org/Bugs/Public/show_bug.cgi?id=27653">Bug 27653</a>
-          has been filed to address this.
+          Nothing prevents an <a>application context</a> from navigating to a
+          <a>URL</a> that is outside of the application's <a>navigation
+          scope</a>, while still having the <a>manifest</a> <a>applied</a> to
+          it.
+        </p>
+        <p>
+          Unlike previous versions of this specification, user agents are no
+          longer required or allowed to block off-scope navigations, or open
+          them in a new <a>top-level browsing context</a>. This practice broke
+          a lot of sites that navigate to a URL on another origin (e.g., to
+          perform third-party authentication), which would then navigate back
+          to the originating site. See <a href=
+          "https://github.com/w3c/manifest/issues/646">Issue 646</a>.
         </p>
       </div>
-      <p>
-        The [[!HTML]] <a>navigate algorithm</a> is monkey-patched as follows:
-      </p>
-      <ul>
-        <li>After the clause "If the <a data-cite=
-        "!HTML#source-browsing-context">source browsing context</a> is not <a>
-          allowed to navigate</a> <var>browsingContext</var>…", add: "or
-          <var>browsingContext</var> is an <a>application context</a>, and
-          <var>resource</var> is not <a>within scope</a> of the navigation
-          scope of <var>browsingContext</var>'s manifest."
-        </li>
-        <li>If, at any time, an <a>application context</a> is navigated to a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -813,39 +813,44 @@ <h2>
...
+        </p>
         <p>
-          Enforcing the navigation scope depends on [[!HTML]]'s navigate
-          algorithm. As such, the following algorithm monkey patches [[!HTML]].
-          <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27653">Bug
-          27653</a> has been filed to address this.
+          Unlike previous versions of this specification, user agents are no
+          longer required or allowed to block off-scope navigations, or open
+          them in a new <a>top-level browsing context</a>. This practice broke
+          a lot of sites that navigate to a URL on another origin (e.g., to
```
dominickng: Perhaps remove the "a lot of" qualifier?

dominickng: "to a URL on another origin" -> "to an off-scope URL" (since scope can be narrower than an origin)

index.html
```diff
@@ -813,39 +813,44 @@ <h2>
...
+          perform third-party authentication), which would then navigate back to
+          the originating site. See
```
dominickng: Remove everything from comma to full stop? Bug has more details.

index.html
```diff
@@ -1512,12 +1517,12 @@ <h3 id="applying">
...
+        <div class="note">
           <p>
-            Please note that the <a>start URL</a> is not necessarily the value
-            of the <a data-link-for="WebAppManifest"><code>start_url</code></a>
-            member: the user or user agent could have changed it when the
-            application was added to home-screen or otherwise bookmarked.
+            The <a>start URL</a> is not necessarily the value of the
+            <a data-link-for="WebAppManifest"><code>start_url</code></a> member:
+            the user or user agent could have changed it when the application
+            was added to home-screen or otherwise bookmarked.
```
dominickng: "was added to...." -> "installed"?

mgiuca: OK, moved this to #701 since it's kind of off-topic for this breaking change.

mgiuca:
I responded by completing all of your suggestions, @dominickng

index.html
```diff
@@ -813,39 +813,43 @@ <h2>
         "!SERVICE-WORKERS-1#scope-match-algorithm">Service Workers</a>. To
         avoid unexpected behavior, use a scope ending in a <code>/</code>.
       </div>
-      <div class="issue" title="🐒 patch">
+      <p>
+        If the <a>application context</a>'s <a data-cite=
+        "!HTML#active-document">active document</a>'s <a data-cite=
+        "!DOM#concept-document-url">URL</a> is not <a>within scope</a> of the
+        navigation scope of the <a>application context</a>'s manifest, the user
+        agent SHOULD show prominent UI indicating the <a data-cite=
```
kenchris: UI elements? UI just means user interface

index.html
```diff
@@ -813,39 +813,43 @@ <h2>
...
+        "!DOM#concept-document-url">document URL</a>, or at least its
+        <a>origin</a>, including whether it is served over a secure connection.
+        This UI SHOULD differ from any UI used when the <a data-cite=
+        "!DOM#concept-document-url">document URL</a> is <a>within scope</a>, in
+        order to alert the user of the off-scope navigation.
```
kenchris: I am not such a fan of "alert" 

in order to make it obvious to the user that they are navigating off-scope.

index.html
```diff
@@ -813,39 +813,43 @@ <h2>
...
+          <a>URL</a> that is outside of the application's <a>navigation
+          scope</a>, while still having the <a>manifest</a> <a>applied</a> to
+          it.
+        </p>
         <p>
-          Enforcing the navigation scope depends on [[!HTML]]'s navigate
-          algorithm. As such, the following algorithm monkey patches [[!HTML]].
-          <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27653">Bug
-          27653</a> has been filed to address this.
+          Unlike previous versions of this specification, user agents are no
```
kenchris: Note?

index.html
```diff
@@ -813,39 +813,43 @@ <h2>
...
           Security considerations
         </h3>
         <p>
-          It should not be possible to navigate the <a>top-level browsing
-          context</a> to somewhere outside the scope while the <a>manifest</a>
-          is <a>applied</a> to the <a>top-level browsing context</a>. That's
-          not to say that the web application cannot be navigated: just that
-          the set of URLs to which the manifest applies is restricted by the
-          <a>navigation scope</a>.
+          The above recommendation to show UI when the <a>application
```
kenchris: *is* to show?

kenchris: No, can you split up the sentence instead?

security -> security reasons?

index.html
mgiuca: Already in a "NOTE" block (don't like redundantly writing "Note" everywhere).

mgiuca:
Responded to @kenchris comments.

index.html
kenchris: ok, great

mgiuca:
Dom and I have been talking and want to proceed with this. We're getting a lot of requests to fix our broken navigation within PWAs.

Chrome has already implemented this on Android, and we're committed to changing our Desktop implementation to match. However, none of the other major browsers do this. So we need a commitment from at least one other browser (Edge, Firefox, Safari) before we land this. @boyofgreen @marcoscaceres @davidquesada please give us your thoughts.

As we've discussed on #646, even though this change breaks the existing spec (which other browsers — including Chrome on Desktop — are implementing), what we're really doing is bringing PWAs in line with normal HTML navigations. Currently, user agents following the Manifest spec are forced to break the normal behaviour of target=self links on existing websites.

marcoscaceres:
@marcoscaceres I'm personally supportive with the change, but I can't commit to implementation until next year 😢 Meeting with my colleagues from Mozilla in December to take stock of where we are at with a product strategy.

mgiuca:
Hi @marcoscaceres. Thanks. I'll take your support and understand that you can't commit to implementing in Firefox until next year.

I haven't heard back from any of the others. I'd like to move forward with this change regardless because the current behaviour is untenable for us (we are constantly receiving reports about this).

I've compiled a document summarizing the issue with a list of reports we've gotten internally, and some external bug reports:
https://docs.google.com/document/d/1ZFItpUruPclnYechVkgXz9lVEwAK2Tuuo-592ohPpDo/edit

Since I've already got an approval from Ken, I'm going to merge this later today unless there are objections.

marcoscaceres:
Go for it. Just please reach out to the Apple folks and make sure they are aware. I'm sure they are probably getting similar reports.

PR #710, opened 2018-08-21
Abstracting selecting an icon.

jakearchibald:
I'm using a similar "pick the best icon from this set" pattern in background fetch, and I'm wanting to use definitions from this spec.

This PR abstracts the prose for selecting an `ImageResource` from a list.

***
<a href="https://pr-preview.s3.amazonaws.com/jakearchibald/manifest/pull/710.html" title="Last updated on Aug 21, 2018, 9:50 AM GMT (162ebb0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/710/60a5156...jakearchibald:162ebb0.html" title="Last updated on Aug 21, 2018, 9:50 AM GMT (162ebb0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1802,66 +1802,16 @@
         </h3>
         <p>
           The <dfn>icons</dfn> member is an <a>array</a> of
-          <a>ImageResource</a>s that can serve as iconic representations of the
-          web application in various contexts. For example, they can be used to
-          represent the web application amongst a list of other applications,
-          or to integrate the web application with an <abbr title=
-          "operating system">OS</abbr>'s task switcher and/or system
-          preferences.
-        </p>
+          <a>ImageResource</a>s.
         <p>
-          If there are multiple equally appropriate icons in <var>icons</var>,
-          a user agent MUST use the last one declared in order at the time that
-          the user agent collected the list of <var>icons</var>. If the user
-          agent tries to use an icon but that icon is determined, upon closer
-          examination, to in fact be inappropriate (e.g. because its content
-          type is unsupported), then the user agent MUST try the
-          next-most-appropriate icon as determined by examining the
-          <a>ImageResource</a>'s members.
+        <p>
+          The user agent MAY <a>select an icon</a> from this array to serve as
+          iconic representations of the web application in various contexts. For
+          example, they can be used to represent the web application amongst a
+          list of other applications, or to integrate the web application with
+          an <abbr title= "operating system">OS</abbr>'s task switcher and/or
+          system preferences.
         </p>
-        <div class="example">
-          <p>
-            In the following example, the developer has made the following
-            choices about the icons associated with the web application:
-          </p>
-          <ul>
-            <li>The developer has included two icons at the same size, but in
-            two different formats. One is explicitly marked as WebP through the
-            <code>type</code> member. If the user agent doesn't support WebP,
-            it falls back to the second icon of the same size. The media type
-            of this icon can then be either determined via a HTTP header, or
-            can be sniffed by the user agent once the first few bytes of the
-            icon are received.
-            </li>
-            <li>The developer wants to use an SVG for greater than or equal to
-            257x257px. She has found that the SVG file looks too blurry at
-            small sizes, even on high-density screens. To deal with this
-            problem, she's included an SVG icon that is only used when the
-            dimensions are at least 257px. Otherwise, the user agent uses the
-            ICO file (hd_hi.ico), which includes a gamut of raster icons
-            individually tailored for small display sizes.
-            </li>
-          </ul>
-          <pre class="example json">
-            {
-              "icons": [
-                {
-                  "src": "icon/lowres.webp",
-                  "sizes": "48x48",
-                  "type": "image/webp"
-                },{
-                  "src": "icon/lowres",
-                  "sizes": "48x48"
-                },{
-                  "src": "icon/hd_hi.ico",
-                  "sizes": "72x72 96x96 128x128 256x256"
-                },{
-                  "src": "icon/hd_hi.svg",
-                  "sizes": "257x257"
-                }]
-            }
-          </pre>
-        </div>
       </section>
       <section>
         <h3>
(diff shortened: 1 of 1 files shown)


jakearchibald:
cc @dbaron - this is an effort to merge together icon usage in specs.

marcoscaceres:
I'm not thrilled that we are using the document format for APIs. I think we should spin off an new `ImageResource` spec that we can use for APIs independently on the manifest spec (we can then build on the stuff in the Notifications API and in the Credential Management API, and in the Payment Handler spec... with the aim of them hopefully using `ImageResource` if possible, as well as the appropriate security model for these `ImageResource`s as defined in Credential Management spec). The `ImageResource` as currently specified is not really suitable for reuse in other specs, because it has `purpose` and `platform` members. 

I think we should seriously consider reverting all the WebIDL changes in the spec, as I think we made a mistake in using WebIDL to define a document format :(

mgiuca:
The "adaptive icon PR" @kenchris mentions is #657. I don't think this clashes with that. `ImageResource` isn't being extended there, we're just adding a new "purpose".

@jakearchibald Will Service Worker be making use of the `ImageResource` dictionary defined here? Including "purpose" and "platform"? Otherwise it might not make sense to do this, as Marcos says.

@marcoscaceres Why is WebIDL not appropriate for defining a document format?

It defines a JSON dictionary structure much more succinctly and readably than the parser algorithms we had previously specified. I don't think there's a problem with using the same definition for both interfaces and JSON. Even if it doesn't make sense in service workers to re-use `ImageResource`, imagine a hypothetical API where you can dynamically register new icons for the manifest:

```webidl
void RegisterIcon(ImageResource icon);
```

It would make sense to re-use `ImageResource` dictionary for that method, no?

jakearchibald:
@mgiuca 

> Will Service Worker be making use of the ImageResource dictionary defined here? Including "purpose" and "platform"?

Hmm no, they don't really make sense. I need src & size, and the hand-wavey selection mechanism (the fetching is different in the background fetch case).

Yeah, maybe this needs to be somewhere else. Ugh I was trying to avoid something as huge as spinning up a new spec 😀.

Anyone know the process for this? I imagine it's different to proposing a new feature.

marcoscaceres:
> @marcoscaceres Why is WebIDL not appropriate for defining a document format?

IDL defines type coercion rules, so when one does:

```
{
   orientation: "bananas",
}
```

That should "throw" a TypeError... but throwing doesn't mean anything in our format.

Similarly, per IDL:

```JS
{
   name: ["this gets coerced to a string"],
}
```

And so on... 

@jakearchibald wrote:
 > Anyone know the process for this? I imagine it's different to proposing a new feature.

We can just quickly spin something up at the WICG. All we need is an index.html file and a collaborative spirit :) When we are happy, we move it back into Web Platform and just explain that we sprouted it from this spec.

jakearchibald:
Here we go https://discourse.wicg.io/t/new-spec-request-select-most-appropriate-image/2964

mgiuca:
@marcoscaceres:

> That should "throw" a TypeError... but throwing doesn't mean anything in our format.

Hmm, yeah that's the one thing that bugs me. We can write a top-level rule that says "if the ECMAScript to IDL conversion throws a TypeError, the manifest is invalid", so it isn't meaningless that the conversion throws a TypeError outside of an actual JavaScript execution context. BUT the problem is that we don't want to trash the whole manifest if one member is invalid. If one member throws a TypeError, we generally want to ignore that member, not the whole thing.

It seems that this isn't an issue with using IDL in a document format, but rather that IDL's conversion algorithm doesn't have a "non-strict mode" (that I know of). If there was a way to say "convert an ECMAScript object to IDL, with any TypeErrors in a dictionary member causing that member to be set as undefined". Or thereabouts. That would be perfectly usable in this spec.

I also wanted to say I'm not sure it's worth breaking the current icon selection language out into its own spec. At the moment, it's just too vague to be useful. The only normative text is:

> If there are multiple equally appropriate icons in icons, a user agent MUST use the last one declared in order at the time that the user agent collected the list of icons. If the user agent tries to use an icon but that icon is determined, upon closer examination, to in fact be inappropriate (e.g. because its content type is unsupported), then the user agent MUST try the next-most-appropriate icon as determined by examining the ImageResource's members.

"appropriate" is never defined. The only real normative requirement here is that ties (of an unspecified evaluation function) are broken by which comes last. Doesn't seem worth trying to make this weak requirement available across specs. Though perhaps this could be an opportunity to better define "appropriate" icon selection.

jakearchibald:
The dictionary and fetching rules would also go into the spec. But I agree the spec would be otherwise small and vague.

marcoscaceres:
> BUT the problem is that we don't want to trash the whole manifest if one member is invalid. If one member throws a TypeError, we generally want to ignore that member, not the whole thing.

Yes, exactly. So to do that, we would need to implement a custom IDL binding layer. That's no feasible. 

Our (Gecko's) implementation just runs the old processing rules:
https://github.com/mozilla/gecko-dev/blob/master/dom/manifest/ManifestProcessor.jsm#L65

We just do simple processing, it doesn't go through the IDL binding layer. Does Chrome's implementation go through the IDL layer? 

> It seems that this isn't an issue with using IDL in a document format, but rather that IDL's conversion algorithm doesn't have a "non-strict mode" (that I know of). 

That's really not feasible. We don't need to fork our IDL biding layer for this format. 

I think we need to go and figure out https://github.com/whatwg/infra/issues/159

marcoscaceres:
> The dictionary and fetching rules would also go into the spec. But I agree the spec would be otherwise small and vague.

Small and vague is good - but with precision on the fetching aspects. Just having the dictionary would be great.

marcoscaceres:
Cc'ing @rsolomakhin for Payment Handlers. I'd like to help with the icon effort when I finish Payment Request. I'm hoping that will be by TPAC. Maybe we can have a meeting at the TPAC Tech Plenary on Wednesday about this with Annevk and Mike for Credential Management?

marcoscaceres:
(I didn't Mike and Anne on purpose... we can formally arrange things once we have a clear understanding of what we want to do)

marcoscaceres:
> @marcoscaceres : Sounds good to work at TPAC, but keep in mind that I'm there only for Monday and Tuesday.

That should be ok, I think.

aarongustafson:
Yesterday (at TPAC) we agreed to move `ImageResource` out to its own spec. @rayankans and I will be tackling that. Once that happens we can look to incorporate some of these suggestions into that.

marcoscaceres:
Closing, as [ImageResource](http://github.com/w3c/image-resource) moved to its own repo.

PR #714, opened 2018-08-28
Fix directive name in CSP figure

christianliebel:
The figure shows `image-src` as a CSP directive, but it’s actually called `img-src`.


kenchris:
@marcoscaceres you want the "Rebase and merge" right?

marcoscaceres:
Squash is usually best.

marcoscaceres:

Issue #733, opened 2018-11-07
Validate ReSpec in CI


marcoscaceres:
We should steal this: 
https://github.com/w3c/webdriver/pull/1339/files#diff-b67911656ef5d18c4ae36cb6741b7965

Then Travis will fail if there are ReSpec warnings/errors.

marcoscaceres:
Got this working a while back...

PR #739, opened 2018-11-13
chore(.travis): check spec for errors/warnings

marcoscaceres:
Part of #733 

This change (choose one):

* [x] Is a "chore" (metadata, formatting, fixing warnings, etc).

Commit message:

Check the spec on Travis, meaning we can drop the need to manually check for the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/739.html" title="Last updated on Nov 13, 2018, 6:02 AM GMT (6abed2f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/739/16c5d13...6abed2f.html" title="Last updated on Nov 13, 2018, 6:02 AM GMT (6abed2f)">Diff</a>

### DIFF
--- a/tools/validate.js
+++ b/tools/validate.js
@@ -5,6 +5,10 @@ const { exec } = require("child_process");
 const handler = require("serve-handler");
 const http = require("http");
 
+/**
+ *
+ * @param {string} cmd A string representing a shell command.
+ */
 function toExecutable(cmd) {
   return {
     get cmd() {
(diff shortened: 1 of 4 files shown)


tools/validate.js
```diff
@@ -0,0 +1,58 @@
...
+"use strict";
+const { exec } = require("child_process");
+const handler = require("serve-handler");
+const http = require("http");
+
+/**
+ *
+ * @param {string} cmd A string representing a shell command.
+ */
+function toExecutable(cmd) {
```
mgiuca: This name is a bit confusing. ("It's generating an executable file?")

How about CommandRunner? Thus, `CommandRunner(cmd).run()`.

marcoscaceres: Good suggestion. Will change.

marcoscaceres: Will just make it a class called `ShellCommand`.

marcoscaceres:
> FYI, I don't know anything about Node or Travis. It looks correct, but I'm not an expert.

That's ok :) Feedback was useful. I verified it fails and passes - so should be good.

PR #741, opened 2018-11-15
Breaking: Ignore any icon with no recognised purpose.

### LINKED ISSUE #728: Icon matching algorithm: unrecognised purposes can result in "any"
https://w3c.github.io/manifest/#purpose-member

Currently, if the user agent doesn't recognise any of the purposes in the list, it defaults to `"any"`. Instead, this should be an error (i.e., the icon should be ignored). The default of `"any"` should only be applied if purpose is the empty string or missing. Otherwise, it is still permitted to have unrecognised purposes in the list, but the UA must only treat the icon as one of the recognised purposes.

mgiuca:
Closes #728.

This change (choose one):

* [ ] Adds new normative behavior

* [ ] Safari (link to issue)
* [X] Chrome (https://crbug.com/905561)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

Breaking: Ignore any icon with no recognised purpose.

Previously, if all of an icon's purposes were unrecognised, it could be
used for any purpose. This causes forwards-compatibility issues, since
if we add a new purpose (such as the recently added 'maskable'), old
user agents will interpret icons with that new purpose as 'any', which
is not intended.

Icons with no stated purpose, or whose purpose is the empty string, are
still treated as 'any'.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/741.html" title="Last updated on Nov 16, 2018, 4:01 AM GMT (586d82b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/741/176d3e2...mgiuca:586d82b.html" title="Last updated on Nov 16, 2018, 4:01 AM GMT (586d82b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2522,7 +2522,7 @@
           <a>purpose</a>.
         </p>
         <p class="note">
-          For example, an icon with purpose "<a>badge</a>" may be used as a
+          For example, an icon with purpose "<a>badge</a>" could be used as a
           badge or pinned icon that is visually distinct, in color or form,
           from an application's launch icon. The user agent uses the value of
           the <a>purpose</a> member as a hint to determine where and how an
@@ -2557,13 +2557,12 @@
           </dd>
         </dl>
         <p class="note">
-          If an icon contains multiple purposes, it can be used for any of
-          those purposes. If all of the stated purposes are not recognized, the
+          If an icon contains multiple purposes, it could be used for any of
+          those purposes. If none of the stated purposes are recognized, the
           icon is totally ignored. For example, if an icon has purpose
-          <code>"badge fizzbuzz"</code>, then it may be used as a badge, but if
-          an icon has purpose <code>"fizzbuzz"</code>, then it must be ignored.
-          This allows new purposes to be added to the standard in the future
-          without them being interpreted as normal icons by older user agents.
+          <code>"badge fizzbuzz"</code>, then it could be used as a badge, but
+          if an icon has just the purpose <code>"fizzbuzz"</code>, then it will
+          be ignored.
         </p>
         <p>
           The steps for <dfn>processing the <code>purpose</code> member of an
--- a/index.html
+++ b/index.html
@@ -2518,11 +2518,11 @@
           When an <a>ImageResource</a> is used as an <dfn>icon</dfn>, a
           developer can hint that the image is intended to serve some special
           purpose in the context of the host OS (i.e., for better integration).
-          User agents SHOULD NOT use an icon other than for its stated purpose.
+          User agents SHOULD NOT use an icon other than for its stated <a>purpose</a>.
         </p>
         <p class="note">
-          For example, an icon with purpose <code>"badge"</code> may be used as
-          a <a data-lt="badge purpose">badge</a> or "pinned" icon that is
+          For example, an icon with purpose "<a>badge</a>" may be used as
+          a <a data-lt="badge purpose">badge</a> or "<a>pinned</a>" icon that is
           visually distinct, in color or form, from an application's launch
           icon. The user agent uses the value of the <a>purpose</a> member as a
           hint to determine where and how an <a>ImageResource</a> is displayed.
--- a/index.html
+++ b/index.html
@@ -2518,7 +2518,8 @@
           When an <a>ImageResource</a> is used as an <dfn>icon</dfn>, a
           developer can hint that the image is intended to serve some special
           purpose in the context of the host OS (i.e., for better integration).
-          User agents SHOULD NOT use an icon other than for its stated <a>purpose</a>.
+          User agents SHOULD NOT use an icon other than for its stated
+          <a>purpose</a>.
         </p>
         <p class="note">
           For example, an icon with purpose "<a>badge</a>" may be used as
--- a/index.html
+++ b/index.html
@@ -2569,7 +2569,7 @@
           The steps for <dfn>processing the <code>purpose</code> member of an
           image</dfn> are given by the following algorithm. The algorithm takes
           an <a>ImageResource</a> <var>image</var>. This algorithm will return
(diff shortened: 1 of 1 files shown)


mgiuca:
Thanks for looking, Ken.

I just pushed a small update.

Since this is a (fairly benign) breaking change, I'll wait for Marcos as well, and try to get at least one other browser to commit to changing their parser.

index.html
```diff
@@ -2518,14 +2518,16 @@ <h3>
...
-        <p>
-          For example, as a <a data-lt="badge purpose">badge</a> or "pinned"
-          icon that is visually distinct, in color or form, from an
-          application's launch icon. The user agent uses the value of the
-          <a>purpose</a> member as a hint to determine where and how an
-          <a>ImageResource</a> is displayed. Unless declared otherwise by the
-          developer, a user agent can use an icon for <a>any purpose</a>.
+        <p class="note">
+          For example, an icon with purpose <code>"badge"</code> may be used as
+          a <a data-lt="badge purpose">badge</a> or "pinned" icon that is
```
```suggestion
          a <a data-lt="badge purpose">badge</a> or "<a>pinned</a>" icon that is
```

index.html
```diff
@@ -2518,14 +2518,16 @@ <h3>
           When an <a>ImageResource</a> is used as an <dfn>icon</dfn>, a
           developer can hint that the image is intended to serve some special
           purpose in the context of the host OS (i.e., for better integration).
+          User agents SHOULD NOT use an icon other than for its stated purpose.
```
marcoscaceres: Do we have a concept of "purpose" defined? 

```suggestion
          User agents SHOULD NOT use an icon other than for its stated <a>purpose</a>.
```

index.html
```diff
@@ -2554,42 +2556,52 @@ <h3>
             The user agent is free to display the icon in any context.
           </dd>
         </dl>
+        <p class="note">
+          If an icon contains multiple purposes, it can be used for any of
```
marcoscaceres: Maybe: 
> used by the user agent for any of

Otherwise, it's a little unclear if the note is speaking to developers.

index.html
```diff
@@ -2554,42 +2556,52 @@ <h3>
...
+          those purposes. If all of the stated purposes are not recognized, the
+          icon is totally ignored. For example, if an icon has purpose
```
marcoscaceres: This is a bit contradictory with the SHOULD above. Maybe the above SHOULD needs to be MUST, otherwise this doesn't hold true.

index.html
```diff
@@ -2554,42 +2556,52 @@ <h3>
...
+          <code>"badge fizzbuzz"</code>, then it may be used as a badge, but if
+          an icon has purpose <code>"fizzbuzz"</code>, then it must be ignored.
```
marcoscaceres: This note is sounding very much like conformance requirements - which are already stated above. It should be phased like it's describing behavior to a developer, not giving conformance requirements to an implementer.

index.html
mgiuca: Actually, I accepted this but I realised it doesn't make sense; "pinned" is not a concept in this doc, it's just being used as a word in this sentence that's quoted for emphasis. (I didn't change this sentence but I'll update it now so it doesn't look like a formal concept.)

marcoscaceres:
```suggestion
          a <a data-lt="badge purpose">badge</a> or "pinned" icon that is
```

marcoscaceres: Hopefully the above lets it be reverted back. Sorry about that.

index.html
mgiuca: No, they are talking about slightly different things.

The SHOULD above:

> User agents SHOULD NOT use an icon other than for its stated purpose.

Is saying you shouldn't use a "maskable" icon (i.e., a valid purpose) as a normal icon. Since this is UI, we can't upgrade this to a MUST.

Whereas this non-normative note:

> If all of the stated purposes are not recognized, the icon is totally ignored.

Is saying that the user agent must not under any circumstances consider an icon whose purpose is not in {"any", "badge", "maskable"}. The corresponding normative text for this is the algorithm below, which converts icons without any recognised purpose into failure.

index.html
mgiuca: Fixed. I assume this is basically asking to change phrasing like "the user agent must" to "the user agent will", basically describing to a developer what would happen in certain situations assuming a conformant user agent, rather than describing conformance. Also avoiding RFC 2119 words.

mgiuca:
(Fantastic! Your Respec validator caught a Respec error on the suggestion commit and blocked the merge.)

Updated in response to review.

index.html
mgiuca: Also, I removed the final sentence "This allows new purposes to be added to the standard in the future without them being interpreted as normal icons by older user agents." which is really more an explanation of _why_ we made it work this way, neither requirements nor an explanation of _how_ it works, so I took that out. Unless you think this is still useful?

index.html
marcoscaceres: Ok, but then the algorithm takes care of dropping the unknown purposes on the floor. Doesn't that mean the SHOULD can go away? The reasoning being that a conforming user agent would have dropped the icons with unknwn purposes, and will only be left with a list of known icons to display.

marcoscaceres: Actually, never mind. You are right that it's about stated purpose.

index.html
marcoscaceres: No, I think it's good now... I should write a ReSpec validation rule to maybe suggest alternative to  RFC 2119 keywords in Notes.

mgiuca:
I forgot to land this. Doing so now (after I had another look over it).

(I started seeing our [test site](https://killer-marmot.appspot.com/web/)'s adaptive icon showing up on desktop platforms as a square. We need to fix this in the implementations.)

PR #743, opened 2018-11-29
Rewrite dir member.

### LINKED ISSUE #742: Remove "steps to programmatically determine the directionality of a member"
Following up from https://github.com/w3c/manifest/issues/676#issuecomment-442658368, this seems to be unnecessary.

We currently have this normative text:

> When the base direction is "`auto`" the user agent *MUST* run the steps to programmatically determine the directionality of a member - and use the resulting text-direction value to assist in displaying the value of the member.
>
> The **steps to programmatically determine the directionality of a member** are as follows. The algorithm takes the *value* of a member.
>
> 1. Find the first character (in logical order) of the *value* that is of bidirectional character type L, AL, or R [[BIDI](https://www.unicode.org/reports/tr9/tr9-37.html)].
> 2. If such a character is found and it is of bidirectional character type AL or R, return "`rtl`".
> 3. Otherwise, return "`ltr`".

This seems entirely redundant for the text in [BIDI](http://unicode.org/reports/tr9/#P2):

> P2. In each paragraph, find the first character of type L, AL, or R while skipping over any characters between an isolate initiator and its matching PDI or, if it has no matching PDI, the end of the paragraph.
>
> ...
>
> P3. If a character is found in P2 and it is of type AL or R, then set the paragraph embedding level to one; otherwise, set it to zero.
>
> Whenever a higher-level protocol specifies the paragraph level, rules P2 and P3 may be overridden

Manifest is overriding rules P2 and P3, legitimately when `dir` is "`left`" or "`right`", but when `dir` is "`auto`", we unnecessarily override it with almost the exact same rule, only less robust (since we do not take isolates into account).

I suggest we drop this text and simply say that we only override BIDI when `dir` is "`left`" or "`right`".

mgiuca:
Closes #742

This change (choose one):

* [ ] Adds new normative behavior

Commit message:

`dir`=`auto` no longer explicitly overrides the algorithm in [BIDI]. Also the whole thing more formally uses the BIDI infrastructure.

Note: I think this is actually making a slight behavioural change which we may not be OK with (but assuming it brings it more in line with actual implementations): if the member is embedded within another string (such as "Do you want to install %s?"), the new text lets the BIDI algorithm take the wheel, whereas the old text implies (without formally stating) that it would be embedded LTR or RTL depending on the first strong character. I'll have to think on this more.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/743.html" title="Last updated on Nov 29, 2018, 4:48 AM GMT (103c2dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/743/176d3e2...mgiuca:103c2dd.html" title="Last updated on Nov 29, 2018, 4:48 AM GMT (103c2dd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1657,36 +1657,24 @@
             <dfn>auto</dfn>
           </dt>
           <dd>
-            <a data-lt=
-            "steps to programmatically determine the directionality of a member">
-            programmatically determined</a> from the value of the member.
+            no explicit directionality.
           </dd>
         </dl>
-        <p>
+        <p data-link-for="TextDirectionType">
           When displaying the <a>directionality-capable members</a> to an
-          end-user, the user agent MUST use the <a>base direction</a> to
-          compute directional runs and layout or position text correctly in
-          text containing mixed-direction sequences [[!BIDI]]. When the <a>base
-          direction</a> is "<a data-link-for="TextDirectionType">auto</a>" the
-          user agent MUST run the <a>steps to programmatically determine the
-          directionality of a member</a> - and use the resulting
-          <a>text-direction value</a> to assist in displaying the value of the
-          member.
-        </p>
-        <p>
-          The <dfn>steps to programmatically determine the directionality of a
-          member</dfn> are as follows. The algorithm takes the <var>value</var>
-          of a member.
+          end-user, if the <a>base direction</a> is <a>ltr</a> or <a>rtl</a>:
         </p>
         <ol>
-          <li>Find the first character (in logical order) of the
-          <var>value</var> that is of bidirectional character type L, AL, or R
-          [[!BIDI]].
+          <li data-link-for="TextDirectionType">If the member is being
+          displayed in a paragraph by itself, the user agent MUST override Rule
+          P3 of [[!BIDI]], setting the paragraph embedding level to 0 if the
+          <a>base direction</a> is <a>ltr</a>, or 1 if the <a>base
+          direction</a> is <a>rtl</a>.
           </li>
-          <li>If such a character is found and it is of bidirectional character
-          type AL or R, return <code>"rtl"</code>.
-          </li>
-          <li>Otherwise, return <code>"ltr"</code>.
+          <li data-link-for="TextDirectionType">Otherwise, the user agent MUST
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1657,36 +1657,24 @@ <h3>
             <dfn>auto</dfn>
           </dt>
           <dd>
-            <a data-lt=
-            "steps to programmatically determine the directionality of a member">
-            programmatically determined</a> from the value of the member.
+            no explicit directionality.
```
```suggestion
            No explicit directionality.
```

mgiuca:
> Note: I think this is actually making a slight behavioural change which we may not be OK with (but assuming it brings it more in line with actual implementations): if the member is embedded within another string (such as "Do you want to install %s?"), the new text lets the BIDI algorithm take the wheel, whereas the old text implies (without formally stating) that it would be embedded LTR or RTL depending on the first strong character. I'll have to think on this more.

I thought a lot about this case. I think this is making a slight normative change, to no longer require, for `dir: auto`, user agents to wrap the string in an LTR or RTL embedding.

- a) it only makes things more permissive (any existing user agents are free to keep wrapping strings in an embedding, which is good practice for any user-supplied strings anyway), and
- b) this only affects extremely rare edge cases (wrapping a multi-directional string in another multi-directional string).

I think it's better to proceed; it's better to treat this like a normal string than impose special rules on how it must be renderer — rules that not only does the user agent need to follow, but it also needs to ensure the OS follows. (Consider generating a [.desktop file](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s06.html) on Linux, and needing to wrap all strings in "[LRE]...[PDF]" or "[RLE]...[PDF]" to force the OS to follow the above rule.)

I'm going to keep this marked as "editorial change" since it doesn't break any existing behaviour but makes it (slightly) more permissive and easier to implement. (Also note that Chrome currently doesn't implement `dir` at all; I'm not sure about the state of the other browsers.)

index.html
mgiuca: All of these start with a lowercase letter. For consistency, I made them all uppercase.

marcoscaceres:
Oh, boo! The authentication on the server is failing. Ok to merge tho.

PR #750, opened 2018-12-10
Added CatchTypeError to all fields of WebAppManifest and other places.

### LINKED ISSUE #611: Describe the manifest data structure
There isn't really a formal specification of the manifest data structure. Instead, each member has its own section which describes the format of that member in prose. This makes it hard to discuss (e.g., in Ken's [shortcut proposal](https://gist.github.com/kenchris/0acec2790cd38dfdff0a7197ff00d1de), we'd like to be able to say "`sequence<IconInfo> icons;`" but there's no name for "the dictionary that describes a Manifest icon" because it's described in prose).

I think we should add a WebIDL section that contains a dictionary definition for the full manifest, as a first step (I can do it).

As a second step, I think this would allow us to simplify the language in each individual member. The "steps for processing a manifest" algorithm could be replaced with "Parse the JSON into an IDL dictionary value, then [convert it into an ECMAScript object](https://www.w3.org/TR/WebIDL-1/#es-dictionary)" (which automatically applies all of the IDL type conversion rules).

Then, for example, this text:

> The steps for processing the short_name member is given by the following algorithm. The algorithm takes a manifest as an argument. This algorithm returns a string or undefined.
>
> 1. Let value be the result of calling the [[GetOwnProperty]] internal method of manifest with argument "`short_name`".
> 2. If Type(*value*) is not "string":
>     1. If Type(*value*) is not "undefined", issue a developer warning that the type is not supported.
>     2. Return `undefined`.
> 3. Otherwise, Trim(*value*) and return the result.

could be replaced by this (assuming `short_name` is a `DOMString`):

> The steps for processing the short_name member is given by the following algorithm. The algorithm takes a manifest [referring to the processed ECMAScript object] as an argument. This algorithm returns a string or undefined.
>
> 1. Let value be the result of calling the [[GetOwnProperty]] internal method of manifest with argument "`short_name`".
> 2. If Type(*value*) is not "string", return *value*.
> 3. Otherwise, Trim(*value*) and return the result.

Were it not for the call to Trim, this would make almost all the processing steps trivial. Unfortunately, almost all of them do Trim their strings, which makes things more complicated. Perhaps we could make a global rule that Manifest strings are trimmed, then we wouldn't really need any processing steps at all.

mgiuca:
**DO NOT SUBMIT** just yet. Depends upon heycam/webidl#597 so I'm just posting this as a request for comments.

Closes #611.

This change (choose one):

**However**, while this is marked as breaking, I suspect this reverts a previous breakage back to the previous behaviour (members with type errors invalidate just the member, not the whole manifest). I'll have to investigate the old text and other implementations. The goal of this is to *improve* spec compatibility with implementations.

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

This restores the previous behaviour (before we switched to WebIDL)
where individual elements being a type error would not invalidate the
whole manifest.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/750.html" title="Last updated on Dec 10, 2018, 6:24 AM GMT (7999f5f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/750/3ea8093...mgiuca:7999f5f.html" title="Last updated on Dec 10, 2018, 6:24 AM GMT (7999f5f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1580,24 +1580,25 @@
       </h2>
       <pre class="idl">
           dictionary WebAppManifest {
-             TextDirectionType dir = "auto";
-             DOMString lang;
-             USVString name;
-             USVString short_name;
-             USVString description;
-             sequence&lt;ImageResource&gt; icons;
-             sequence&lt;ImageResource&gt; screenshots;
-             sequence&lt;USVString&gt; categories;
-             DOMString iarc_rating_id;
-             USVString start_url;
-             DisplayModeType display = "browser";
-             OrientationLockType orientation;
-             USVString theme_color;
-             USVString background_color;
-             USVString scope;
-             ServiceWorkerRegistrationObject serviceworker;
-             sequence&lt;ExternalApplicationResource&gt; related_applications;
-             boolean prefer_related_applications = "false";
+             [CatchTypeError] TextDirectionType dir = "auto";
+             [CatchTypeError] DOMString lang;
+             [CatchTypeError] USVString name;
+             [CatchTypeError] USVString short_name;
+             [CatchTypeError] USVString description;
+             [CatchTypeError] sequence&lt;[CatchTypeError] ImageResource&gt; icons;
+             [CatchTypeError] sequence&lt;[CatchTypeError] ImageResource&gt; screenshots;
+             [CatchTypeError] sequence&lt;[CatchTypeError] USVString&gt; categories;
+             [CatchTypeError] DOMString iarc_rating_id;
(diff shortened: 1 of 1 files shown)


marcoscaceres:
In 32b497c4ca34c9069aa5e545126c252724f370fe, I (finally) got rid of the IDL and converted the "processed manifest" to an (Infra) ordered map.

Issue #693, opened 2018-06-14
Add a way to query whether there is a UA-provided back button


mgiuca:
As discussed in #673, developers of "display: standalone" (or "minimal-ui") apps cannot be sure of whether a back button will be provided by the user agent (meaning, the browser **or** operating system **or** hardware). For some apps, this is fine, but for others, not having a back button makes it hard to navigate, so they need to add an in-app back button. The problem is, if there is already one provided by the system, this can be redundant, wasting space and looking silly.

At the extreme, this can lead to the "double back button" problem, seen here in the Twitter PWA currently [installable from the Microsoft Store](https://www.microsoft.com/en-us/p/twitter/9wzdncrfj140) on Windows 10 (this internally uses https://mobile.twitter.com/):

![image](https://user-images.githubusercontent.com/228433/41394569-c44a3330-6fed-11e8-9ea9-13f72deb39b5.png)

We can discuss ways to let developers *control* whether a back button is shown in #673; here I want to propose that we give developers a way to *query* whether a back button is shown. Thus Twitter and other apps can conditionally show an in-app back button, depending on whether there is already a system one. Giving a standard way to check this means we can avoid problematic user-agent detection, or viewport-size detection (which is just as bad for compatibility --- that's what Twitter does today).

## Proposal

I propose a [CSS media feature](https://www.w3.org/TR/mediaqueries-4/#mq-features) which says whether a back button is currently provided by the user agent. The precedent for this is the [`display-mode`](https://www.w3.org/TR/appmanifest/#the-display-mode-media-feature) media feature already in the Manifest spec.

The proposed feature is named `navigation-controls` with possible values {`'none'` | `'back-button'`}, (for example, allowing it to be extended later with `'back-and-forward'`), but it is designed to be used in a [Boolean context](https://www.w3.org/TR/mediaqueries-4/#mq-boolean-context) so all future values would imply the presence of the back button. Thus, you could use it like this to show a back button if-and-only-if there isn't already a system-provided one:

```css
@media navigation-controls {
  #go-back {
    display: none;
  }
}
```

(Note: I use an enum instead of a Boolean due to [this note in Media Queries](https://www.w3.org/TR/mediaqueries-4/#grid): "The &lt;mq-boolean&gt; type exists only for legacy purposes. If this feature were being designed today, it would instead use proper named keywords for its values.")

[...]

PR #766, opened 2019-05-17
Add navigation_ui property (fixes #765).

### LINKED ISSUE #765: Add Manifest property to hint whether the navigation UI should be visible
In the same spirit of: https://fullscreen.spec.whatwg.org/#enumdef-fullscreennavigationui

This will allow websites that are in fullscreen, standalone or other compatible modes to hint the UA as to whether they would like the navigation UI to show up. We could re-use the same values as the fullscreen spec ("auto"/"show"/"hide").

An obvious benefit would be for websites that use the fullscreen display mode to be able to benefit from this feature from the fullscreen API but we could also have UA allowing this setting for standalone applications.

mounirlamouri:
Closes #765 

This change (choose one):

* [ ] Safari (link to issue)
* [X] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message: add navigation_ui property

Adds hint in the manifest regarding the visibility of the navigation ui.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/766.html" title="Last updated on May 17, 2019, 7:04 PM UTC (54d40b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/766/0349be6...54d40b8.html" title="Last updated on May 17, 2019, 7:04 PM UTC (54d40b8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1063,6 +1063,62 @@
       </p>
       <section>
         <h3>
+          Navigation UI
+        </h3>
+        <p>
+          A <dfn>navigation UI mode</dfn> represents whether the web application
+          would like the user agent to display a navigation UI, reducing the
+          screen space used by the content. The <a>navigation UI mode</a> MUST
+          be ignored when the <a>display mode</a> is <a>browser</a>. The user
+          agent MAY ignore the <a>navigation UI mode</a> for other <a>display
+          mode</a>.
+        </p>
+        <p>
+          When the <a data-link-for="WebAppManifest">navigation_ui</a> member is
+          missing, or if there is no valid <a data-link-for=
+          "WebAppManifest">navigation_ui</a> member, the user agent uses the
+          <a data-link-for="NavigationUIModeType">auto</a> <a>navigation ui
+          mode</a> as the <a>navigation ui mode</a>.
+        </p>
+        <p>
+          The <a data-link-for="NavigationUIModeType">auto</a> <a>navigation ui
+          mode</a> allows the user agent to use either <a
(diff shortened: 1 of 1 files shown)


marcoscaceres:
The proposal seems fine... I guess we might want to be a bit more suggestive about what it might include? Like, obviously the ability to navigate back, forward, and reload?

mounirlamouri:
> The proposal seems fine... I guess we might want to be a bit more suggestive about what it might include? Like, obviously the ability to navigate back, forward, and reload?

In practice, at least for fullscreen, this is only used on mobile and it's usually backward and back to homescreen. I'm happy to add some guidance around this to be a bit more specific if that works for you.

mounirlamouri:
@marcoscaceres ping :)

g-ortuno:
How does this fit with `minimal-ui`, #693, and #673?

mounirlamouri:
The spec suggests that if `navigation_ui` is set, it should be respected and if it's not set, the default behaviour should be followed. This would apply to `minimal_ui`  as much as to `fullscreen` or `standalone`. On mobile, it's likely that a UA would show the navigation bar with `minimal_ui` but the website could have a way to opt out. Regarding the inconsistencies on desktop, this would offer one more toggle that websites can use to increase consistency and guarantee a certain level of experience.

marcoscaceres:
This proposal didn't seem to go anywhere. @mounirlamouri, can we close this?

mounirlamouri:
As far as I know, Chrome is no longer planning in adding this feature. I will close.

Issue #582, opened 2017-05-23
App shortcuts / actions


hemanth:
On native apps, we have an option to provide app shortcuts on long press on the app icon, for example:

![image](https://cloud.githubusercontent.com/assets/18315/26338473/e98f8888-3f9d-11e7-808e-f9bce251f7ff.png)

I feel the same would be very useful for PWAs and it would be easy to mention them in the mainfest?

//cc @slightlyoff

hemanth:
Maybe something like below makes sense?

```
contextMenu: [{
   url: '..',
   icon: '..',
   description: '...'
}]
```

marcoscaceres:
probably just "label": ... can someone dig up how this registration is done in Android itself (via Java or Xml or whatever)

marcoscaceres:
Also, icon: '..', should be an "image object", as per the spec.

marcoscaceres:
Btw, what do the little double bars do? Is that to drag things into different order?

hemanth:
Double bars will help us to add that action to home screen.

![image](https://cloud.githubusercontent.com/assets/18315/26395235/12c4f1ee-408d-11e7-9be0-f134bb3b0d5a.png)

`shortcut.xml`:

```xml
<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
  <shortcut
    android:shortcutId="compose"
    android:enabled="true"
    android:icon="@drawable/compose_icon"
    android:shortcutShortLabel="@string/compose_shortcut_short_label1"
    android:shortcutLongLabel="@string/compose_shortcut_long_label1"
    android:shortcutDisabledMessage="@string/compose_disabled_message1">
    <intent
      android:action="android.intent.action.VIEW"
      android:targetPackage="com.example.myapplication"
      android:targetClass="com.example.myapplication.ComposeActivity" />
    <!-- If your shortcut is associated with multiple intents, include them
         here. The last intent in the list is what the user sees when they
         launch this shortcut. -->
    <categories android:name="android.shortcut.conversation" />
  </shortcut>
  <!-- Specify more shortcuts here. -->
</shortcuts>
```

kenchris:
Proposal from @mgiuca inspired by earlier proposal from me. I think this is pretty nice and the direction we should go:

https://docs.google.com/document/d/1WzpCnpc1N7WjDJnFmj90-Z5SALI3cSPtNrYuH1EVufg/edit

sheshbabu:
[...]

PR #768, opened 2019-06-14
Add shortcuts feature to the explainer and spec

aarongustafson:
This change (choose one):

* [ ] Safari ([link to issue](https://bugs.webkit.org/show_bug.cgi?id=201964))
* [x] Chrome ([link to issue](https://bugs.chromium.org/p/chromium/issues/detail?id=955497))
* [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1582341)
* [x] Edge ([public signal](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/Shortcuts/explainer.md))
* [x] Samsung Internet ([public signal](https://github.com/w3c/manifest/pull/768#issuecomment-532941900))

Commit message:

Following discussion on issue #582, adding abbreviated shortcut documentation to [the explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/Shortcuts/explainer.md).

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Nov 4, 2019, 5:58 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Faarongustafson%2Fmanifest%2F59eaa7a76963bd58cc39cec549f578efb5f2a9eb%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/aarongustafson/manifest/59eaa7a76963bd58cc39cec549f578efb5f2a9eb/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/manifest%23768.)._
</details>


aarongustafson:
Let me know how this looks @marcoscaceres.

siusin:

aarongustafson:
@marcoscaceres @raymeskhoury @kenchris I’ve merged shortcuts into the spec and explainer. I think I’ve managed to catch most of the nuance in the spec-authoring process, but please let me know if I’ve missed anything.

A few notes:

* For clarity, I’ve changed `ShortcutInfo` to `ShortcutItem`. I think this makes the spec read a lot better.
* I’ve abstracted the `params` property of `ShortcutItem` to a new abstract dictionary `RequestParams` as it is something that can be reused in many contexts going forward (including Web Share Target).

Thanks for letting me take a crack at this. Hopefully I haven’t broken things too badly. Respec seems generally happy with me right now.

aarongustafson:
@marcoscaceres I’m trying to figure out why the build is failing, so I tried to run `respec-validator` locally, but it is failing because `respec2html` isn’t published. Thoughts?

index.html
```diff
@@ -1495,12 +1495,12 @@ <h3>
           <var>manifest</var>["<a>categories</a>"].
           </li>
           <li>Set <var>manifest</var>["<a>icons</a>"] to the result of running
-          <a>processing ImageResource members</a> given
+          <a>processing <code>ImageResource</code> members</a> given
```
kenchris: I believe we can use backtick instead of <code> in respec today

index.html
```diff
@@ -2362,6 +2367,94 @@ <h3>
...
+      <section>
+        <h3>
+          <code>shortcuts</code> member
+        </h3>
+        <p>
+          The <dfn>shortcuts</dfn> member is an <a>array</a> of
+          <a>ShortcutItem</a>s that provide access to key tasks within a web
+          application. For example, they can be used to link directly to a
+          user’s timeline within a social media application or their recent
+          orders in an ecommerce context.
```
kenchris: e-commerce?

index.html
```diff
@@ -2945,6 +3038,167 @@ <h3>
...
+        <h3>
+          <dfn>Launching a shortcut</dfn>
+        </h3>
+        <p>
+          When <a>ShortcutItem</a> <var>shortcut</var> having
+          <a>WebAppManifest</a> <var>manifest</var> is invoked, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be a copy of <var>shortcut.url</var>.
```
kenchris: I dont think we need to say 'copy' @marcoscaceres

index.html
```diff
@@ -2945,6 +3038,167 @@ <h3>
...
+          </li>
+          <li>Let <var>header list</var> be a new empty <a data-xref-for=
+          "request">header list</a>.
+          </li>
+          <li>If <var>shortcut.params</var>:
```
kenchris: refer to "exists" from infra spec

kenchris:
cc @mgiuca and @jakearchibald regarding the `RequestParams`

index.html
aarongustafson: I’ll test. I wasn’t sure if markdown was supported now.

index.html
aarongustafson: Fixing now. Is there a style guide and favored dictionary for specs?

index.html
aarongustafson: I cribbed this from [the proposed Web Share Target spec](https://wicg.github.io/web-share-target/#launching-the-web-share-target).

index.html
aarongustafson: Done.

index.html
aarongustafson: It is.

kenchris: Check the wake lock spec for use of modern Respec features :-)

index.html
kenchris: I don't believe so, but US English is the norm

index.html
```diff
@@ -2368,6 +2373,98 @@ <h3>
...
       </section>
+      <section>
+        <h3>
+          <code>shortcuts</code> member
+        </h3>
+        <p>
+          The <dfn>shortcuts</dfn> member is an <a>array</a> of
+          <a>ShortcutItem</a>s that provide access to key tasks within a web
+          application. For example, they can be used to link directly to a
+          user’s timeline within a social media application or their recent
```
mgiuca: nit: There's a fancy apostrophe here; I think this should just be a regular ASCII apostrophe (maybe search the document for other instances of this).

index.html
mgiuca: I'm not sure we need it there either.

index.html
```diff
@@ -2368,6 +2373,98 @@ <h3>
...
+          <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ShortcutItem</a>&gt;. For
+          each <var>shortcut</var> (<a>ShortcutItem</a>) in the sequence, set
+          <var>shortcut.icons</var> to the result of running <a>processing
+          `ImageResource` members</a> given <var>shortcut.icons</var> and
+          <var>manifest URL</var>.
+        </p>
+        <p>
+          A user agent MUST expose shortcuts via interactions that are
+          consistent with exposure of an application icon’s context menu in the
```
mgiuca: here too

index.html
```diff
@@ -2368,6 +2373,98 @@ <h3>
...
+          <var>shortcuts</var> as an argument. This algorithm returns an
+          <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ShortcutItem</a>&gt;. For
+          each <var>shortcut</var> (<a>ShortcutItem</a>) in the sequence, set
+          <var>shortcut.icons</var> to the result of running <a>processing
+          `ImageResource` members</a> given <var>shortcut.icons</var> and
+          <var>manifest URL</var>.
+        </p>
+        <p>
+          A user agent MUST expose shortcuts via interactions that are
```
mgiuca: These MUSTs should be SHOULDs. We can't mandate what the user agent does with its own (or the OS's) UI. From a spec point of view, we treat the UA and OS as a single entity --- if the UA wants to invent a whole new mechanism for displaying the icon that's different to how the OS normally does it, well that's fine. It could also display these things inside the browser UI, not the OS at all.

We also can't impose a MUST on the ordering; the user agent could provide the list of shortcuts to the OS and then the OS may sort them alphabetically, or sort them based on how often the user presses each one.

index.html
```diff
@@ -2945,6 +3042,168 @@ <h3>
...
+          </li>
+        </ol>
+      </section>
+    </section>
+    <section>
+      <h2>
+        <dfn>RequestParams</dfn>
+      </h2>
+      <pre class="idl">
+        dictionary RequestParams {
```
mgiuca: Why is this an empty dictionary? I guess you intend this to contain arbitrary key/values (not specifying which ones)? I don't think a dictionary is appropriate here. This should be an IDL [record](https://heycam.github.io/webidl/#idl-record).

However, I don't think we want RequestParams at all --- see comment above.

index.html
```diff
@@ -2945,6 +3042,168 @@ <h3>
...
+          <a>WebAppManifest</a> <var>manifest</var> is invoked, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be a copy of <var>shortcut.url</var>.
+          </li>
+          <li>Let <var>header list</var> be a new empty <a data-xref-for=
+          "request">header list</a>.
+          </li>
+          <li>If <var>shortcut.params</var> <a data-cite="infra/#list-contain">
```
mgiuca: Why do we need `shortcut.params`? This doesn't let the user, user agent, or site dynamically inject parameters into the URL, does it? Like, it seems that the params are a static set of key/value pairs defined in the manifest alongside `url`. Thus, why do we need to specify this as a separate dictionary that gets encoded into a query string? Why not just let the manifest specify query parameters in the `url` directly?

As an example, say you had this:

```json
"shortcuts": [
  {
      "name": "Subscriptions",
      "url": "/subscriptions",
      "params": {
        "sortby": "ascending"
      }
  }
]
```

why not just express it as this:

```json
"shortcuts": [
  {
      "name": "Subscriptions",
      "url": "/subscriptions?sortby=ascending"
  }
]
```

aarongustafson: Fair point. I think this was a vestige of when we had considered supporting POST, etc. and could likely be done away with assuming everyone agrees that shortcuts should only support GET requests.

index.html
aarongustafson: Ah cool, I was unaware of `record`. I figured there had to be a similar generic, but had not found it yet. Thanks!

index.html
aarongustafson: This seems like a duplicate of the other one (MUST -> SHOULD). Am I missing something?

aarongustafson:
@mgiuca You said "I'm not sure we need it there either." but I’m not sure what "it" is. Is it a reference to the shortcut parameters?

index.html
aarongustafson: NM, GitHub comments suck sometimes. This was in reference to "A user agent MUST render the shortcuts" right?

aarongustafson:
Ok, I made all of the changes, save the removal of `shortcut.params`. I just want confirmation that we have no desire to allow verbs other than GET before I remove it (and possibly have to add it back in).

+ @mgiuca @kenchris @marcoscaceres

index.html
mgiuca: Yes, this is another MUST -> SHOULD.

index.html
mgiuca: Oh OK, I didn't know there was consideration to support POST. Now that you mention it, why was it decided not to? Is it "maybe in the future" or we have an active reason not to?

I agree it would be a little weird to click a shortcut and have it perform some side effect, and it's probably a better user experience to take you to a GET page that asks if you want to perform the side effect first. So I'm cool with disallowing POST but just wanted to get the rationale.

mgiuca:
I am OK with saying we'd never want POST.

Another thing to consider is that POST requests _can_ take query parameters (it's just that HTML forms with action=POST do not produce them). So we could add POST in the future without needing a separate params dict.

christianliebel:
The platform conventions for shortcut icons seem to differ quite a bit. For instance, iOS shortcuts use icon masks; Windows jump list entries have colored icons, etc.

Is there a plan on how to deal with this? Should a user agent on IOS create an icon mask from a given image? And/or should`ImageResource`’s `platform` member be evaluated for that scenario?

I assume this question will arise very quickly, so it would be great if the spec could be a bit clearer on this.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+        </h3>
+        <p>
+          When <a>ShortcutItem</a> <var>shortcut</var> having
+          <a>WebAppManifest</a> <var>manifest</var> is invoked, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be <var>shortcut.url</var>.
+          </li>
+          <li>Let <var>browsing context</var> be the result of creating a new
```
christianliebel: What does "new top-level browsing context" mean here? Would it create a new window in all cases? On desktop platforms, it would rather be common to reuse one of the existing windows (or launch the app with given parameters when it’s closed).

I think these steps are okay if the application isn't open. During runtime, I'd suggest raising an `onshortcutclick` event on a service worker which can then override the default behavior (e.g. by navigating an existing client window or opening a new one).

aarongustafson:
> The platform conventions for shortcut icons seem to differ quite a bit. For instance, iOS shortcuts use icon masks; Windows jump list entries have colored icons, etc.
> 
> Is there a plan on how to deal with this? Should a user agent on IOS create an icon mask from a given image? And/or should`ImageResource`’s `platform` member be evaluated for that scenario?

I feel like that might be a good thing to consider with respect to `imageResource` in a broader sense. I know some launchers also support different shapes (circle, squircle, etc.)

index.html
aarongustafson: In most cases it would be a fresh launch of the PWA or (potentially) a fresh tab within the context of a PWA (if tabs are supported). There is separate work ongoing with respect to the [launch event](https://github.com/WICG/sw-launch) that feels like it could enable what you are looking for with respect to the service worker.

NotWoods:
The different launcher shapes are handled by `purpose: maskable` for image resources right now. Maybe a similar field could be used for `ShortcutItem`s? Depending on the specifics of the platform, another option could be to only use the alpha layer of the shortcut (for iOS, Android) and then use full color elsewhere (Windows).

aarongustafson:
> The different launcher shapes are handled by `purpose: maskable` for image resources right now. Maybe a similar field could be used for `ShortcutItem`s? Depending on the specifics of the platform, another option could be to only use the alpha layer of the shortcut (for iOS, Android) and then use full color elsewhere (Windows).

Shortcuts piggybacks on `ImageResource`, so that would be covered.

mgiuca:
Yeah, since shortcut uses `ImageResource`, maskable should cover any platform-specific needs with regards to conforming all icons to a specific shape. We don't currently have a solution for platforms that want icons to be all monochrome. It's possible there's a more pressing need for that in relation to shortcuts than generic app icons (since some platforms may have monochrome shortcut icons).

If we want to go down that path, I would propose a new `purpose: "monochrome"` for ImageResource (that would be separate to this PR). I want to avoid any ability for developers to supply an "iOS icon" and a "Windows icon", etc, because then we will break away from the Web platform's platform neutrality and start seeing apps with poor or no icons on other platforms. (Devs can already technically do this by user-agent-detecting when they serve their manifest, but I want to avoid giving them an explicit feature to do this.)

To do this, we should start a separate issue, starting with an analysis of what shortcut icon guidelines there are on each major platform so we can see whether "monochrome" is necessary and sufficient. Anyone care to do that?

That issue aside, I still think we want to drop params, but happy to be overruled in further discussion. Has there been any more (at TPAC for instance)?

beverloo:
Would the `badge` purpose be sufficient? It specifically leaves such flexibility as we have the same constraints in other UI surfaces:

> A user agent can present this icon where space constraints and/or color requirements differ from those of the application icon.

aarongustafson:
Bug files with Webkit. No commitment yet.

marcoscaceres:
Filed Firefox bug... seems we can't support this on Android, but doing it on Windows might be an option https://bugzilla.mozilla.org/show_bug.cgi?id=1582341

hjrchung:
Samsung Internet will also implement this feature.

mgiuca:
I think the discussion about the platform-specific icon requirements is off-topic of this PR since no matter the outcome of that, we don't need to address it here. Spun off to #795 to continue discussion.

aarongustafson:
Looping in @slightlyoff for feedback on v1.

aarongustafson:
@mgiuca Are there any outstanding issues with respect to this PR or are we good to merge?

explainer.md
```diff
@@ -194,6 +194,62 @@ TBW: using description and screenshots.
 ## Theme color and background color
 TBW...
 
+## Adding shortcuts
+Numerous operating systems grant native applications the ability to add menu items to the app icon itself. These often provide quick access to key tasks for an app. Typically, these are exposed via a right click, long tap, or a similar context menu-triggering action. For web applications, you can define a set of shortcuts to be exposed when the app is installed. Each shortcut item must have a name and a target URL. You may also include additional information, such as a shorter name, a description for the action, one or more icons, and parameters to be passed along to the target URL.
```
mgiuca: "and parameters to be passed along to the target URL"

No longer true; delete this.

index.html
```diff
@@ -2368,6 +2373,95 @@ <h3>
...
+          <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ShortcutItem</a>&gt;. For
+          each <var>shortcut</var> (<a>ShortcutItem</a>) in the sequence, set
+          <var>shortcut.icons</var> to the result of running <a>processing
+          `ImageResource` members</a> given <var>shortcut.icons</var> and
+          <var>manifest URL</var>.
+        </p>
+        <p>
+          A user agent SHOULD expose shortcuts via interactions that are
+          consistent with exposure of an application icon’s context menu in the
```
mgiuca: nit: This right quote should be a straight apostrophe.

index.html
```diff
@@ -2368,6 +2373,95 @@ <h3>
...
+          host operating system (e.g., right click, long press). A user agent
+          SHOULD render the shortcuts in the same order as they are provided in
+          the manifest. A user agent SHOULD represent the shortcuts in a manner
+          consistent with exposure of an application icon’s context menu in the
```
mgiuca: Same here.

index.html
```diff
@@ -2368,6 +2373,95 @@ <h3>
...
+            agent would present
+            <samp>https://example.com/icons/play-later.svg</samp> next to the
+            text. When launched, the user agent would instantiate a new
+            <a>top-level browsing context</a> and navigate to
+            <samp>https://example.com/play-later</samp>.
+            </li>
+            <li>The first shortcut would be displayed with the text
+            "Subscriptions". When launched, the user agent would instantiate a
+            new <a>top-level browsing context</a> and navigate to
+            <samp>https://example.com/subscriptions?sort=desc</samp>.
```
mgiuca: The `sort=desc` part is no longer true since you removed params.

Maybe just add `?sort=desc` to the `"url"` in the example (without changing this sentence), to show that you can use query parameters in your URL.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+          USVString description;
+          required USVString url;
+          sequence&lt;ImageResource&gt; icons;
+        };
+      </pre>
+      <p>
+        Each <a>ShortcutItem</a> represents a link to a key task or page within
+        a web app. A user agent can use these values to assemble a context menu
+        to be displayed by the operating system when a user engages with the
+        web app’s icon.
```
mgiuca: Quote -> apostrophe

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+        <p>
+          The steps for <dfn>processing the <code>url</code> member of a
+          shortcut</dfn> are given by the following algorithm. The algorithm
+          takes a <a>ShortcutItem</a> <var>url</var>, and a <var>manifest
+          URL</var>, which is the <a>URL</a> from which the <var>manifest</var>
+          was fetched. This algorithm will return a <a>URL</a> or
+          <code>undefined</code>.
+        </p>
+        <ol>
+          <li>Let <var>value</var> be <var>url</var>
```
mgiuca: No need to create a *value* when you already have a variable *url*.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+          </li>
+          <li>If <a>Type</a>(<var>value</var>) it not <a>String</a>, return
```
mgiuca: I don't think this is necessary. (The WebIDL already guarantees that `url` is a string.) You can delete this.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+        <h3>
+          <code>url</code> member
+        </h3>
+        <p>
+          The <dfn>url</dfn> member of a <a>ShortcutItem</a> is a <a>URL</a> to
+          which a user agent should navigate when the associated shortcut is
+          activated.
+        </p>
+        <p>
+          The steps for <dfn>processing the <code>url</code> member of a
```
mgiuca: Given my comments below, this doesn't need to be a separate steps-for. In fact, these steps aren't called from the steps for processing ShortcutInfo, so the URL would never be parsed.

Delete this section, and just say "parse *shortcut*.`["url"]` using *manifest URL* as the base URL" up in the steps for processing ShortcutInfo.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+        </p>
+        <ol>
+          <li>Let <var>url</var> be <var>shortcut.url</var>.
+          </li>
+          <li>Let <var>browsing context</var> be the result of creating a new
+          <a>top-level browsing context</a>.
+          </li>
+          <li>
+            <a>Navigate</a> <var>browsing context</var> to a new <a>request</a>
+            whose url is <var>url</var>, header list is <var>header list</var>,
```
mgiuca: What is *header list*? That isn't mentioned elsewhere in the document, so I think it should be dropped from this sentence.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+            and body is <var>body</var>.
```
mgiuca: Same with *body*.

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be <var>shortcut.url</var>.
+          </li>
+          <li>Let <var>browsing context</var> be the result of creating a new
+          <a>top-level browsing context</a>.
+          </li>
+          <li>
+            <a>Navigate</a> <var>browsing context</var> to a new <a>request</a>
```
mgiuca: *request* doesn't need to be a variable since it's never referred to again.

I don't think you really need to create a request here. Can you just say "Navigate *browsing context* to *url*"?

index.html
```diff
@@ -2945,6 +3039,118 @@ <h3>
...
+        <h3>
+          <dfn>Launching a shortcut</dfn>
+        </h3>
+        <p>
+          When <a>ShortcutItem</a> <var>shortcut</var> having
+          <a>WebAppManifest</a> <var>manifest</var> is invoked, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be <var>shortcut.url</var>.
```
mgiuca: nit: shortcut.url should not be formatted as a variable, but as an object, like you see in [processing a manifest](https://www.w3.org/TR/appmanifest/#dfn-processing-a-manifest): `<var>shortcut</var>["<a>url</a>"]`.

index.html
mgiuca: Yep, the `launch` event was [specifically designed with this feature in mind](https://github.com/WICG/sw-launch/blob/master/explainer.md#background).

.travis.yml
```diff
@@ -5,7 +5,7 @@ node_js:
 
 branches:
   only:
-    - gh-pages
+    - master
```
mgiuca: This change seems incorrect and unrelated. We don't have a `master` branch; it's `gh-pages` (dating back to when GitHub forced you to use that branch name if you wanted to host content on github.io).

explainer.md
aarongustafson: Done

index.html
aarongustafson: Done.

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Done

index.html
aarongustafson: Fixed

.travis.yml
aarongustafson: I don’t remember why I changed this. Reverted.

index.html
```diff
@@ -2403,6 +2408,98 @@ <h3>
...
         </p>
       </section>
+      <section>
+        <h3>
+          <code>shortcuts</code> member
+        </h3>
+        <p>
+          The <dfn>shortcuts</dfn> member is an <a>array</a> of
+          <a>ShortcutItem</a>s that provide access to key tasks within a web
+          application. For example, they can be used to link directly to a
```
mgiuca: The "for example" sentence should be in a note paragraph as it is non-normative.

index.html
```diff
@@ -2403,6 +2408,98 @@ <h3>
...
+          user's timeline within a social media application or their recent
+          orders in an e-commerce context.
+        </p>
+        <p>
+          A developer may provide as many shortcuts as they like, but the user
```
mgiuca: Shouldn't use the keyword "may" when this is not a normative requirement.

I would just drop the first clause and rewrite this sentence. Suggestion: "How shortcuts are presented, and how many of them are shown to the user, is at the discretion of the user agent and/or operating system."

index.html
```diff
@@ -2403,6 +2408,98 @@ <h3>
...
+          agent and operating system will decide how many are presented to the
+          user. For this reason, it is recommended that developers order their
```
mgiuca: This sentence is non-normative and should be in a note. Also you can't use the keyword "recommended" when giving instructions to the developer (since this spec governs how UAs work, not how developers should behave). Reword as "Developers are encouraged to order their ...".

index.html
```diff
@@ -2403,6 +2408,98 @@ <h3>
...
+        </p>
+        <p>
+          A user agent SHOULD expose shortcuts via interactions that are
+          consistent with exposure of an application icon's context menu in the
+          host operating system (e.g., right click, long press). A user agent
+          SHOULD render the shortcuts in the same order as they are provided in
+          the manifest. A user agent SHOULD represent the shortcuts in a manner
+          consistent with exposure of an application icon's context menu in the
+          host operating system. A user agent MAY truncate the list of
+          shortcuts presented in order to remain consistent with conventions of
```
mgiuca: Perhaps add "... with conventions **or limitations** of  ..." (since it may not just be conventions that constrain the UA, but a hard limit).

index.html
```diff
@@ -2953,6 +3050,98 @@ <h3>
...
+        web app's icon.
+      </p>
+      <section data-dfn-for="ShortcutItem" data-link-for="ShortcutItem">
+        <h3>
+          <code>name</code> member
+        </h3>
+        <p>
+          The <dfn>name</dfn> member of a <a>ShortcutItem</a> is a
+          <a>string</a> that represents the name of the shortcut as it is
+          usually displayed to the user in a context menu
```
mgiuca: Period.

index.html
```diff
@@ -2953,6 +3050,98 @@ <h3>
...
+        </p>
+      </section>
+      <section data-dfn-for="ShortcutItem" data-link-for="ShortcutItem">
+        <h3>
+          <code>description</code> member
+        </h3>
+        <p>
+          The <dfn>description</dfn> member of a <a>ShortcutItem</a> is a
+          <a>string</a> that allows the developer to describe the purpose of
+          the shortcut. User agents may choose to expose this information to
```
mgiuca: Replace "may choose to" with "MAY". (i.e., make this a normative requirement)

index.html
```diff
@@ -2953,6 +3050,98 @@ <h3>
...
+        </h3>
+        <p>
+          The <a>icons</a> member of an <a>ShortcutItem</a> member is an
+          <a>array</a> of <a>ImageResource</a>s that can serve as iconic
+          representations of the shortcut in various contexts.
+        </p>
+      </section>
+      <section>
+        <h3>
+          <dfn>Launching a shortcut</dfn>
```
mgiuca: This isn't linked from anywhere.

I think it would be good if the main paragraph above that introduces all the normative requirements had a sentence to the effect of "When the user invokes a shortcut from the operating system menu, the user agent SHOULD run **Launching a shortcut**".

index.html
```diff
@@ -2953,6 +3050,98 @@ <h3>
...
+          assistive technology.
+        </p>
+      </section>
+      <section data-dfn-for="ShortcutItem" data-link-for="ShortcutItem">
+        <h3>
+          <code>url</code> member
+        </h3>
+        <p>
+          The <dfn>url</dfn> member of a <a>ShortcutItem</a> is a <a>URL</a> to
+          which a user agent should navigate when the associated shortcut is
```
mgiuca: You can't use the keyword "should" here, unless you want to introduce a normative requirement. (But since you already give normative text for this in the algorithm below, I think just reword this.) "the URL that opens when the shortcut is activated."

PR #775, opened 2019-08-06
fix: purpose member type check + empty string handling

### LINKED ISSUE #771: purpose member type check missing
When processing purpose, the spec assumes that the value will be a String. However, it could be any other JSON type. 

@dominickng or @mgiuca, could you kindly please check what Chrome does when given:

```JSON
{
    "icons": [{"src": "test", "sizes": "400x400", "purpose": [] }]
}
``` 

Does it convert purpose to "any"? or does the icon get ignored (failure)? 

Also, probably too late... but `purpose` should have been a List :( When we do canonicalization, we return it as a list. Anyway, no big deal.

marcoscaceres:
Closes #771 

This change (choose one):

* [ ] Safari (link to issue)
* [x] Chrome - current behavior as per #771 
* [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1559412)
* [ ] Edge (public signal)

Commit message:
Clarifies behavior for purpose member type check + empty string handling (to match Chrome):

 * empty string = treat as "any".
 * type check fails = treat as "any".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/775.html" title="Last updated on Aug 6, 2019, 10:55 AM UTC (61a8937)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/775/c10cbdd...61a8937.html" title="Last updated on Aug 6, 2019, 10:55 AM UTC (61a8937)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2547,42 +2547,38 @@
           be ignored.
         </p>
         <p>
-          The steps for <dfn>processing the <code>purpose</code> member of an
-          image</dfn> are given by the following algorithm. The algorithm takes
-          an <a>ImageResource</a> <var>image</var>. This algorithm will return
-          a <a>set</a> or failure.
+          The steps for <dfn>processing the `purpose` member of an image</dfn>
+          are given by the following algorithm. The algorithm takes an
+          <a>ImageResource</a> |image:ImageResource|. This algorithm returns a
+          [=set=] or failure.
         </p>
         <ol>
-          <li>If <a>Type</a>(<var>image</var>["purpose"]) it not String, or
-          <var>image</var>["purpose"] consists solely of <a>ascii
-          whitespace</a>, then return the <a>set</a> « "any" ».
+          <li>If [=Type=](|image|["purpose"]) it not String, or
+          |image|["purpose"] consists solely of [=ascii whitespace=], then
+          return the [=set=] « "any" ».
           </li>
-          <li>Let <a>list</a> <var>keywords</var> be the result of <a>split on
-          ASCII whitespace</a> <var>image</var>["purpose"].
+          <li>Let |keywords:list&lt;string&gt;| be the result of [=split on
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2547,44 +2547,38 @@ <h3>
...
             <ol>
-              <li>Set <var>keyword</var> to <var>keyword</var>,
-              <a>lowercased</a>.
+              <li>Set |canonicalKeyword| to [=lowercased=] |keyword|.
               </li>
-              <li>If <var>keyword</var> is not one of the <a>icon purposes</a>,
-              or <var>set</var> <a>contains</a> <var>keyword</var>, then
-              <a>issue a developer warning</a> and <a>continue</a>.
+              <li>If |canonicalKeyword| is not one of the [=icon purposes=], or
+              |purposes| [=contains=] |keyword|, then [=issue a developer
```
dominickng: FYI: Chrome actually will happily give you back duplicates if we see a valid purpose that is already in the set (we return an array, not a set). It's probably right that the spec doesn't allow duplicates, and we should fix that in Chrome.

marcoscaceres: > Chrome actually will happily give you back duplicates 

When you say "will happily give you back duplicates", you mean internally at the browser level, right?

marcoscaceres: > we return an array, not a set

We do the same... can convert it to list, as it's cleaner.

marcoscaceres: Actually, will leave it... as sets are already lists, so it's an internal implementation detail.

marcoscaceres:
Thanks for the speedy review @kenchris and @dominickng!

index.html
dominickng: Yes, I mean internally in the browser here, sorry for the lack of clarity. :)

Issue #784, opened 2019-09-09
Clarify behaviour of a page linking to a manifest that is not within scope of that manifest


mgiuca:
Coming across from [Chrome issue 961525](https://crbug.com/961525).

Consider the case where a page (e.g., https://stormy-pincushion.glitch.me) links to a manifest whose scope does not enclose the page that originally linked the manifest (e.g., "scope": "/why"). Should this be considered a valid manifest for the page?

The spec doesn't really provide any guidance on this. Since the details in the spec around installation are pretty fuzzy, it's hard to see exactly how to fit this in, but I think we'd like to essentially _ignore_ the manifest entirely in this situation. The manifest does not apply to this page, therefore it shouldn't influence the display of the page (e.g., the theme colour, etc), nor should we consider this page to be able to trigger installation.

Currently, Chrome [does allow installation](https://crbug.com/961525) for this page, resulting in a weird situation where as soon as you install the app, you are shown an out-of-scope address bar since the page you were on is not part of the app scope.

I would like to add to the [installation process](https://www.w3.org/TR/appmanifest/#installation-process) text along these lines:

"If obtaining the manifest succeeds, if the current document URL is not [within scope](https://www.w3.org/TR/appmanifest/#dfn-within-scope) of the manifest, fail."

Also the installation process could generally be cleaned up, since it suggests that the install takes place in parallel to obtaining the manifest, which doesn't make sense because you need to obtain the manifest first before you can even know whether it can be installed.

See https://crbug.com/961525 for some Chrome-team discussion about what should happen. Another possibility is that we explicitly allow this, basically saying that a page can offer a manifest on the same origin but in a different scope for installation. My feeling is that this almost certainly represents an accidentally-too-small-scope so we shouldn't allow it.

We should also change [authority of the manifest's metadata](https://www.w3.org/TR/appmanifest/#authority-of-the-manifest-s-metadata).

Note: Original test case was https://harmonious-cosmonaut.glitch.me which has no "scope" (using the implicit scope of the manifest directory); I forked it to https://stormy-pincushion.glitch.me which has an explicit scope to more directly demonstrate the issue.

mgiuca:
From an offline discussion with @dominickng : I'm pseudo-convinced that we should _allow_ the installation from an off-scope page of a manifest that is on the **same origin**.

We identified two separate use cases for why you might want to do this as a site owner:

[...]

Issue #788, opened 2019-09-12
Refactor the manifest install flow


mgiuca:
The [**steps to install the web application**](https://www.w3.org/TR/appmanifest/#dfn-steps-to-install-the-web-application) is a bit of a mess. (In addition to #787.)

Firstly, it's confusing having both [**installation process**](https://www.w3.org/TR/appmanifest/#dfn-installation-process) (the unspecified user-agent-specific part of installation) as well as **steps to install the web application** (with specific steps). The latter calls the former as its first step, but the order of operations is quite confusing here: first the install process is initiated (does that mean wait until it's finished, or just start it off in a separate thread of execution, in which case this step should have its own "in parallel"?), then we do things like fetching the manifest which would seem necessary to the install process. But we can't move the install process later on in the steps, because prompting the user is part of the install process.

Secondly, prompting the user can't be part of the steps to install the web application, since some flows (those that use [**request to present an install prompt**](https://www.w3.org/TR/appmanifest/#dfn-request-to-present-an-install-prompt)) have already prompted the user by that point (although, those [don't actually trigger these steps](https://github.com/w3c/manifest/issues/786)).

Determining whether the manifest is valid and the site is installable really can't be part of "steps to install the web application" since we need that information before we decide to even show a prompt to the user (e.g., so the user agent can determine whether to show an install button in its UI).

I propose that we:
1. Split apart the determination of whether a site is installable out of the steps to install, making it a separate process, "steps to determine installability of the document".
2. Split apart prompting the user out of the "installation process", making it a precursor step (i.e., "steps to install the web application" is only triggered *after* the user accepts the prompt).
3. Remove the concept of "installation process" that runs in parallel to "steps to install the web application"; instead adding an explicit step to perform unspecified actions to register the app with the host OS.

mgiuca:
Closed by #790

Issue #787, opened 2019-09-12
Steps to install the web application does not take into account whether the installation process was canceled


mgiuca:
The spec defines [**installation was canceled**](https://www.w3.org/TR/appmanifest/#dfn-installation-was-canceled) but this term is never referenced anywhere.

The [**steps to install the web application**](https://www.w3.org/TR/appmanifest/#dfn-steps-to-install-the-web-application) trigger the installation process but then proceeds without explicitly checking whether the install was cancelled.

mgiuca:
Closed by #790

Issue #789, opened 2019-09-12
Service worker installation uses an algorithm (Start Register) that doesn't exist in SW spec


mgiuca:
> Invoke [Start Register](https://www.w3.org/TR/appmanifest/#dfn-start-register) with scope and src members of the registration, a new promise, client, manifest URL, plus the type and update_via_cache members of the registration, in which case the state of the settled promise determines whether the installation succeeded or not.

There is no "Start Register" in the SW spec. As far as I can tell, the nearest thing would be to simply invoke the [`register()`](https://www.w3.org/TR/service-workers-1/#navigator-service-worker-register) method.

I'm also not sure why this method would (ever) have taken the manifest URL. If we can avoid passing the manifest URL, then we don't need to store it in this algorithm.

Also note that this feature doesn't seem to have been implemented by any user agent, so it is considered "at risk" (though I still believe it would be useful which is why I haven't pushed to deleted it yet).

PR #790, opened 2019-09-12
Rewrite installation process and install prompting logic

### LINKED ISSUE #786: BeforeInstallPromptEvent.prompt() does not actually perform installation
There is actually no logic in the spec that calls "steps to install the web application" or "installation process". Maybe it's implied that the user agent does this in response to a user request to do so, but at least this needs to be explicit in the BeforeInstallPrompt.prompt() flow, which currently (according to the spec) does this:

1. [`prompt()`](https://www.w3.org/TR/appmanifest/#prompt-method) calls **request to present an install prompt**.
2. [**Request to present an install prompt**](https://www.w3.org/TR/appmanifest/#dfn-request-to-present-an-install-prompt) calls **present an install prompt**.
3. [**Present an install prompt**](https://www.w3.org/TR/appmanifest/#dfn-present-an-install-prompt) has no specific logic associated, but returns the user's choice.
4. **Request to present an install prompt** resolves the `prompt()` promise with the user's choice.

Nowhere in this flow does the app install process actually get triggered!

mgiuca:
This change (choose one):

Commit message:

Rewrite installation process and install prompting logic per #788.

- Introduces a new concept of the document being "installable", a stateful
  property of the document based on validity of the manifest.
- Adds a new process, "steps to determine installability of the document", split
  out from "steps to install the web application".
- Removed the concept of "installation process", "installation succeeded",
  "installation canceled" (which was unused; see #787) and "installation
  failed". These concepts are now part of the "steps to install the web
  application".
- "Steps to install the web application" now assumes that the document is
  installable, and that user has already accepted the install prompt, where
  previously the determination of installability and prompting were all part of
  that same process.
- Processes that show an install prompt now explicitly call "steps to install
  the web application" after the user accepts the prompt (see #786).

Normative changes are to fix bugs in the spec which made it non-implementable; I
believe these simply bring the spec in line with all known implementations, so
no implementation change is required due to this spec change.

Closes #786, #787, #788.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/790.html" title="Last updated on Sep 26, 2019, 6:42 AM UTC (7b0b5d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/790/6851b18...mgiuca:7b0b5d7.html" title="Last updated on Sep 26, 2019, 6:42 AM UTC (7b0b5d7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -307,7 +307,7 @@
             user-agent-specific way (e.g., setting
             <var>manifest</var>.<a data-link-for="WebAppManifest">`name`</a> to
             the document <a data-cite="HTML#the-title-element">`title`</a>) and
-            considering the document <a>installable</a>, or
+            considering the document <a>installable</a>.
             </li>
             <li>Consider the document non-<a>installable</a>.
             </li>
--- a/index.html
+++ b/index.html
@@ -421,8 +421,8 @@
         <ul>
           <li>An end-user can <dfn data-lt="manual installation">manually</dfn>
           trigger the installation process through the user agent's
-          <abbr>UI</abbr>, directly invoking the steps to <a>present an install
-          prompt</a>.
+            <abbr title="User Interface">UI</abbr>, directly invoking the steps
+            to <a>present an install prompt</a>.
           </li>
           <li>The installation process can occur through an <dfn>automated
           install prompt</dfn>: that is, a <abbr>UI</abbr> that the user agent
--- a/index.html
+++ b/index.html
@@ -465,11 +465,13 @@
           "AppBannerPromptOutcome">accepted</a> or <a data-link-for=
           "AppBannerPromptOutcome">dismissed</a>.
           </li>
-          <li>If <var>result</var> is <a data-link-for=
-          "AppBannerPromptOutcome">accepted</a>, then in parallel, run the <a>
-            steps to install the web application</a>.
-          </li>
-          <li>Return <var>result</var>.
+          <li>Return <var>result</var>, and <a>in parallel</a>:
+            <ol>
+              <li>If <var>result</var> is <a data-link-for=
+              "AppBannerPromptOutcome">accepted</a>, run the <a>steps to
+              install the web application</a>.
+              </li>
+            </ol>
           </li>
         </ol>
         <p>
--- a/index.html
+++ b/index.html
@@ -463,14 +463,11 @@
           "AppBannerPromptOutcome">accepted</a> or <a data-link-for=
           "AppBannerPromptOutcome">dismissed</a>.
           </li>
-          <li>In parallel:
-            <ol>
-              <li>If the user choose to proceed, run the <a>steps to install
-              the web application</a>.
-              </li>
-              <li>Return <var>result</var>.
-              </li>
-            </ol>
+          <li>If <var>result</var> is <a data-link-for=
+          "AppBannerPromptOutcome">accepted</a>, then in parallel, run the <a>
+            steps to install the web application</a>.
+          </li>
+          <li>Return <var>result</var>.
           </li>
         </ol>
         <p>
--- a/index.html
+++ b/index.html
@@ -454,8 +454,9 @@
           showing the prompt.
         </p>
         <p>
-          To <dfn data-lt="presenting an install prompt">present an install
-          prompt</dfn>:
+          To <dfn data-lt=
+          "presenting an install prompt|presentation of the install prompt">present
+          an install prompt</dfn>:
         </p>
         <ol>
           <li>Show some user-agent-specific UI, asking the user whether to
@@ -512,15 +513,15 @@
           Privacy and security considerations
         </h3>
         <p>
-          During the <a data-lt="present an install prompt">presentation of the
-          install prompt</a>, it is RECOMMENDED that the user agent allow the
-          end-user to inspect the icon, name, <a>start URL</a>, origin, etc.
-          pertaining to a web application. This is to give an end-user an
-          opportunity to make a conscious decision to approve, and possibly
-          modify, the information pertaining to the web application before
-          installing it. This also gives the end-user an opportunity to discern
-          if the web application is spoofing another web application, by, for
-          example, using an unexpected icon or name.
+          During the <a>presentation of the install prompt</a>, it is
+          RECOMMENDED that the user agent allow the end-user to inspect the
+          icon, name, <a>start URL</a>, origin, etc. pertaining to a web
+          application. This is to give an end-user an opportunity to make a
+          conscious decision to approve, and possibly modify, the information
+          pertaining to the web application before installing it. This also
+          gives the end-user an opportunity to discern if the web application
+          is spoofing another web application, by, for example, using an
+          unexpected icon or name.
         </p>
         <p>
           It is RECOMMENDED that user agents prevent other applications from
@@ -633,14 +634,13 @@
         </p>
         <div class="note">
           If the <a>BeforeInstallPromptEvent</a> is <em>not</em> cancelled, the
-          user agent is allowed to <a data-lt=
-          "present an install prompt">present an automated install prompt</a>
-          to the end-user. Canceling the default action (via <a data-cite=
+          user agent is allowed to <a>present an install prompt</a>
+          (specifically, an <a>automated install prompt</a>) to the end-user.
+          Canceling the default action (via <a data-cite=
           "DOM#dom-event-preventdefault">preventDefault</a>) prevents the user
-          agent from <a data-lt="present an install prompt">presenting an
-          automated install prompt</a>. The user agent is free to run <a>steps
-          to notify that an install prompt is available</a> again at a later
-          time.
+          agent from <a>presenting an install prompt</a>. The user agent is
+          free to run <a>steps to notify that an install prompt is
+          available</a> again at a later time.
         </div>
         <p data-dfn-for="PromptResponseObject">
           The <dfn>PromptResponseObject</dfn> contains the result of calling
@@ -754,8 +754,7 @@
           </h4>
           <p>
             The <dfn>AppBannerPromptOutcome</dfn> enum's values represent the
-            outcomes from <a data-lt="present an install prompt">presenting the
-            end-user with an install prompt</a>.
+            outcomes from <a>presenting an install prompt</a>.
           </p>
           <dl data-dfn-for="AppBannerPromptOutcome">
             <dt>
--- a/index.html
+++ b/index.html
@@ -305,9 +305,9 @@
             <li>Fall back to using the <a>top-level browsing context</a>
             {{Document}}'s metadata to to populate <var>manifest</var> in a
             user-agent-specific way (e.g., setting
-            |manifest|.{{WebAppManifest/name}} to
-            the document <a data-cite="HTML#the-title-element">`title`</a>) and
-            considering the document <a>installable</a>.
+            |manifest|.{{WebAppManifest/name}} to the document <a data-cite=
+            "HTML#the-title-element">`title`</a>) and considering the document
+            <a>installable</a>.
             </li>
             <li>Or, consider the document not <a>installable</a>.
             </li>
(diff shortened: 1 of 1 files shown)


mgiuca:
@dominickng to also have a look.

This PR will likely be a prerequisite to fix #784 and #789 (I started work on #784 and found the algorithm to be non-sensical, hence this PR).

index.html
```diff
@@ -414,41 +418,59 @@ <h2>
...
+          recommendations relating to this UI. The <var>result</var> of this
+          choice is either <a data-link-for=
+          "AppBannerPromptOutcome">accepted</a> or <a data-link-for=
+          "AppBannerPromptOutcome">dismissed</a>.
+          </li>
+          <li>If <var>result</var> is <a data-link-for=
+          "AppBannerPromptOutcome">accepted</a>, then in parallel, run the <a>
+            steps to install the web application</a>.
+          </li>
+          <li>Return <var>result</var>.
```
marcoscaceres: This is ambiguous: is `result` returned before "in parallel" runs, or after? As it reads (to me), it's returning before, which I think is incorrect, right?

index.html
```diff
@@ -457,9 +479,10 @@ <h2>
           <li>Wait until the {{Document}} of the <a>top-level browsing
           context</a> is <a>completely loaded</a>.
           </li>
-          <li>If there is already an <a>installation process</a> being
-          <a data-lt="present an install prompt">presented</a>, terminate this
-          algorithm.
+          <li>If there is already an <a data-lt=
```
marcoscaceres: We probably need need some kind of internal slot to track this, which can be reset as part of task once installation completes (and can hook into manual installation)... we don't need  to do it as part of this PR tho.

index.html
```diff
@@ -287,6 +287,35 @@ <h2>
         and again as an example, the user agent could <a>install</a> the web
         application into a list of bookmarks within the user agent itself.
       </p>
+      <p>
+        A {{Document}} may either be <dfn>installable</dfn> or not. The initial
+        state of a document is non-<a>installable</a>.
```
```suggestion
        state of a document is not <a>installable</a>.
```

index.html
```diff
@@ -287,6 +287,35 @@ <h2>
...
+        agent MAY either:
+          <ol>
+            <li>Fall back to using the <a>top-level browsing context</a>
+            {{Document}}'s metadata to to populate <var>manifest</var> in a
+            user-agent-specific way (e.g., setting
+            <var>manifest</var>.<a data-link-for="WebAppManifest">`name`</a> to
+            the document <a data-cite="HTML#the-title-element">`title`</a>) and
+            considering the document <a>installable</a>, or
+            </li>
+            <li>Consider the document non-<a>installable</a>.
```
```suggestion
            <li>Or, Consider the document not <a>installable</a>.
```

index.html
```diff
@@ -287,6 +287,35 @@ <h2>
...
+        </li>
+        <li>If <a>obtaining the manifest</a> results in an error, the user
+        agent MAY either:
+          <ol>
+            <li>Fall back to using the <a>top-level browsing context</a>
+            {{Document}}'s metadata to to populate <var>manifest</var> in a
+            user-agent-specific way (e.g., setting
+            <var>manifest</var>.<a data-link-for="WebAppManifest">`name`</a> to
+            the document <a data-cite="HTML#the-title-element">`title`</a>) and
+            considering the document <a>installable</a>, or
```
marcoscaceres: Too easy to miss the "or" right at the end there... 
```suggestion
            considering the document <a>installable</a>.
```

index.html
```diff
@@ -287,6 +287,35 @@ <h2>
...
+        <li>Let <var>manifest</var> and <var>manifest URL</var> be the result
+        of <a>obtaining the manifest</a>.
+        </li>
+        <li>If <a>obtaining the manifest</a> results in an error, the user
+        agent MAY either:
+          <ol>
+            <li>Fall back to using the <a>top-level browsing context</a>
+            {{Document}}'s metadata to to populate <var>manifest</var> in a
+            user-agent-specific way (e.g., setting
+            <var>manifest</var>.<a data-link-for="WebAppManifest">`name`</a> to
```
```suggestion
            |manifest|.{{WebAppManifest/name}} to
```

index.html
```diff
@@ -334,77 +363,52 @@ <h3>
...
-                </ol>
-              </li>
-              <li>If the <a>installation succeeded</a>, <a>queue a task</a> on
-              the <a>application life-cycle task source</a> to <a>fire an
-              event</a> named <code>appinstalled</code> at the
-              <var>window</var> object.
+              <li>Let <var>client</var> be the <a>top-level browsing
+              context</a> {{Document}}'s <a>relevant settings object</a>, or
+              <code>null</code> if unavailable.
+              </li><!-- Start Register no longer exists in the Service Worker
```
marcoscaceres: Do we have some other alternative if Start Register is gone?

index.html
```diff
@@ -414,41 +418,59 @@ <h2>
...
-          sufficient <a>installability signals</a> to warrant
-          <a>installation</a> of the web application.
-          </li>
-          <li>The <a>installation process</a> can occur through a
-          <dfn>site-triggered install prompt</dfn>: the site can
-          programmatically request that the user agent present an install
-          prompt to the user. The user agent MAY restrict the availability of
-          this feature to cases where, for instance, there are sufficient
+          trigger the installation process through the user agent's
+          <abbr>UI</abbr>, directly invoking the steps to <a>present an install
```
marcoscaceres: Just a note... `<abbr title="User Interface">UI</abbr>` just needs to be defined once in the document and ReSpec will find all instances of UI and add the right things for you.

index.html
```diff
@@ -457,9 +479,10 @@ <h2>
           <li>Wait until the {{Document}} of the <a>top-level browsing
           context</a> is <a>completely loaded</a>.
           </li>
-          <li>If there is already an <a>installation process</a> being
-          <a data-lt="present an install prompt">presented</a>, terminate this
-          algorithm.
+          <li>If there is already an <a data-lt=
+          "present an install prompt">install prompt being presented</a> or if
```
marcoscaceres: note: you shouldn't need the `data-lt` here.

index.html
```diff
@@ -487,14 +510,15 @@ <h3 id="installation-sec">
...
         <p>
-          During the <a>installation process</a>, it is RECOMMENDED that the
-          user agent allow the end-user to inspect the icon, name, <a>start
-          URL</a>, origin, etc. pertaining to a web application. This is to
-          give an end-user an opportunity to make a conscious decision to
-          approve, and possibly modify, the information pertaining to the web
-          application before installing it. This also gives the end-user an
-          opportunity to discern if the web application is spoofing another web
-          application, by, for example, using an unexpected icon or name.
+          During the <a data-lt="present an install prompt">presentation of the
```
marcoscaceres: Note, you can put the `data-lt="presentation of the install prompt"` on the `dfn` instead, then you don't need the `data-lt` on the  `a` below.

index.html
mgiuca: It's intended to be before (i.e., your reading is correct). I don't think it's ambiguous: that's precisely the meaning of "in parallel" (otherwise I wouldn't have said "in parallel").

The result of "present an install prompt" should be the user's choice, not the success of the installation. The result of this promise reflects the user's choice, whereas the `appinstalled` event reflects whether it was actually installed.

At least, that was my intention. Now I'm not so sure. Looking at Chrome's [app_banner_manager](https://cs.chromium.org/chromium/src/chrome/browser/banners/app_banner_manager_desktop.cc?l=197), the code that sends BannerAccepted (which results in the promise resolving with `"accepted"`) has a comment:

```C++
  // Catch only kSuccessNewInstall and kUserInstallDeclined. Report nothing on
  // all other errors.
```

implying that maybe we resolve with `"accepted"` if the user accepts AND installation is success, and perhaps we never resole if the user accepts and installation fails??? @dominickng Can you confirm what Chrome is doing here?

We should decide what the (currently unspecified) behaviour should be if the user accepts the banner but the installation fails: if we want to resolve with `"accepted"` then the proposed text is correct. If not, we should delete "in parallel" and figure out what to resolve with (perhaps we should introduce a new `"accepted-but-failed"` result?

index.html
mgiuca: Yep. You're tracking this at #761 so I thought I'd just leave it for now.

Similarly, we should probably add an internal slot for "is installable" since that's a document-global property that I'm adding, but I didn't want to do it in this PR.

index.html
mgiuca: Yeah but I'll look at this as part of #789.

index.html
mgiuca: This seems to be the one place we do need it (since "install prompt being presented" isn't defined as an alias, and it seems too specific to do so).

index.html
mgiuca: OK I've gone through all references to this in the document and normalized them (in some cases slightly rewording) to avoid data-lts.

mgiuca:
Thanks for the quick review, @marcoscaceres. I've accepted or made most of the changes you proposed. I will follow up with @dominickng about the behavioural question.

index.html
mgiuca: Discussed with @dominickng. Chrome does:

- On Android, it resolves with `"accepted"` as soon as the user chooses to install, regardless of whether the installation actually succeeds (matching the proposed "in parallel" text).
- On Windows / Chrome OS / Linux / macOS, if the user accepts and the install "fails", the promise won't resolve. But the failures are all very much edge cases (like something has gone horribly wrong, or the web contents is destroyed, etc), not expected to happen in normal situations. We would consider these edge cases Chrome bugs if they happened. For all intents and purposes, the proposed "in parallel" text matches our behaviour.

So I would prefer to keep this as-is.

index.html
```diff
@@ -334,77 +365,52 @@ <h3>
...
           The <dfn>steps to install the web application</dfn> are given by the
           following algorithm:
         </p>
         <ol>
-          <li>Let <var>window</var> be the {{Window}} object of the
-          <a>top-level browsing context</a> for which the user agent will
-          attempt installation.
+          <li>Let <var>manifest</var> and <var>manifest URL</var> be the values
+          that were created during <a>steps to determine installability of the
+          document</a>.
```
dominickng: whether the document is installable ?

index.html
```diff
@@ -287,6 +287,37 @@ <h2>
...
+            |manifest|.{{WebAppManifest/name}} to the document <a data-cite=
+            "HTML#the-title-element">`title`</a>) and considering the document
+            <a>installable</a>.
+            </li>
+            <li>Or, consider the document not <a>installable</a>.
+            </li>
+          </ol>
+        </li>
+        <li>Otherwise, the {{Document}} MAY be considered <a>installable</a>
+        (at the user agent's discretion; see <a href="#installability-signals">
```
marcoscaceres: This should auto-expand to the right section. 
```suggestion
        (at the user agent's discretion; see [[[#installability-signals]]].
```

index.html
```diff
@@ -287,6 +287,37 @@ <h2>
...
+          Installability signals</a>).
```
```suggestion
```

index.html
```diff
@@ -414,41 +420,60 @@ <h2>
...
-          <dfn>site-triggered install prompt</dfn>: the site can
-          programmatically request that the user agent present an install
-          prompt to the user. The user agent MAY restrict the availability of
-          this feature to cases where, for instance, there are sufficient
+          trigger the installation process through the user agent's
+            <abbr title="User Interface">UI</abbr>, directly invoking the steps
+            to <a>present an install prompt</a>.
+          </li>
+          <li>The installation process can occur through an <dfn>automated
+          install prompt</dfn>: that is, a <abbr>UI</abbr> that the user agent
```
marcoscaceres: ReSpec will add abbr automagically. 

```suggestion
          install prompt</dfn>: that is, a UI that the user agent
```

index.html
marcoscaceres: Ok, alternative:

```
<li>Return |result| and in parallel: .... 
```
 
That way it's clear that in parallel goes off to do other things independently of |result|.

mgiuca: OK that works.

PR #791, opened 2019-09-12
Require document to be in scope to be styled by the manifest

### LINKED ISSUE #784: Clarify behaviour of a page linking to a manifest that is not within scope of that manifest

mgiuca:
This change (choose one):

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

Require document to be in scope to be styled by the manifest.

This introduces a small new normative requirement that the manifest NOT be
applied to a document that it outside of its scope. A note explains that this
only applies to the authority of the metadata, not the installation.

Adds a non-normative note that the app can be installed from a document that is
outside the scope of that app.

Closes #784.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/791.html" title="Last updated on Sep 26, 2019, 7:00 AM UTC (5c2f0a4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/791/f8f227e...mgiuca:5c2f0a4.html" title="Last updated on Sep 26, 2019, 7:00 AM UTC (5c2f0a4)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/791.html" title="Last updated on Oct 31, 2024, 10:19 PM UTC (6fe490b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/791/f8f227e...mgiuca:6fe490b.html" title="Last updated on Oct 31, 2024, 10:19 PM UTC (6fe490b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -336,7 +336,7 @@
           the installable web application) to show installation UI.
         </p>
         <p>
-          Implementors should be aware of this possibility. One consequence is
+          Implementors need to be aware of this possibility. One consequence is
           that if, post install, the user agent opens the document URL in an
           <a>application context</a>, it would immediately start showing the
           "out of scope" warning UI (that is recommended in
--- a/index.html
+++ b/index.html
@@ -269,7 +269,7 @@
       </section>
     </section>
     <section>
-      <h2>
+      <h2 id="installable-web-applications">
         Installable web applications
       </h2>
       <p>
@@ -331,13 +331,25 @@
           Authority of the manifest's metadata
         </h3>
         <p>
-          When a <a>manifest</a> is linked from a {{Document}}, it indicates to
-          the user agent that the metadata is <dfn>authoritative</dfn>: that
-          is, the user agent SHOULD use the metadata of the manifest instead of
-          the one in the {{Document}}. However, in cases where metadata is
-          missing, or in error, a user agent MAY fallback to the {{Document}}
-          to find suitable replacements for missing manifest members (e.g.,
-          using `application-name` in place of `short_name`).
+          When a <a>manifest</a> is linked from a {{Document}}, and the
+          {{Document}} is <a data-lt="within-scope-manifest">within scope</a>
+          of the manifest, it indicates to the user agent that the metadata is
+          <dfn>authoritative</dfn>: that is, the user agent SHOULD use the
+          metadata of the manifest instead of the one in the {{Document}}.
+          However, in cases where metadata is missing, or in error, a user
+          agent MAY fallback to the {{Document}} to find suitable replacements
+          for missing manifest members (e.g., using `application-name` in place
+          of `short_name`). The user agent MUST NOT apply the metadata of a
+          manifest to a {{Document}} that is not <a data-lt=
+          "within-scope-manifest">within scope</a> of that manifest.
+        </p>
+        <p class="note">
+          A {{Document}} that is not <a data-lt="within-scope-manifest">within
+          scope</a> of a linked manifest can still be used to install the web
+          application associated with the manifest, it just can't have the
+          manifest's metadata applied directly to the document. See the note in
+          <a href="#installable-web-applications">Installable web
+          applications</a>.
         </p>
       </section>
       <section data-link-for="WebAppManifest">
--- a/index.html
+++ b/index.html
@@ -326,6 +326,25 @@
         (at the user agent's discretion; see [[[#installability-signals]]]).
         </li>
       </ol>
+      <div class="note" data-link-for="WebAppManifest">
+        <p>
+          The {{Document}} might not be <a data-lt=
+          "within-scope-manifest">within scope</a> of the linked manifest, by
+          design. For example, a page at <code>https://example.com</code> might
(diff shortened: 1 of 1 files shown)


mgiuca:
@dominickng @marcoscaceres Could you look over this language before circulating this more widely.

I think we should seek approval from several vendors on both of the changes included here (both the normative change to ban a manifest from applying to an out-of-scope document, and the explicit decision _not_ to make normative changes to ban manifests from being installable from an out-of-scope document).

Note that this PR is based on #790 so the diff is confusing (come on GitHub please support upstream branches!!!!). Just look at the last two commits.

dominickng:
Does it make sense to surface a warning when processing the manifest that the current document URL is out of scope? Otherwise this seems good to me and I think Chrome can align with this behaviour.

mgiuca:
I think if you want to consider this a legitimate use case (as you have argued to me), then we shouldn't show a warning.

If we consider it "bad enough" to show a warning, then I think we should just ban it.

dominickng:
This is referring to the change to ban manifests from *applying* to out of scope documents. The ability to *install* from an out of scope document to an in-scope document on the same origin wouldn't be what the warning is targeting. But I'm fine with not surfacing a warning in any case.

marcoscaceres:
@mgiuca, what should we do about this PR?

mgiuca:
Hmm, I forgot about this. It might be superseded by #880. I'm waiting for your other PR to land before I tackle #880, so let me get back to this once those land.

marcoscaceres:
sounds like a plan.

index.html
```diff
@@ -326,18 +326,48 @@ <h2>
...
+        <p>
+          The {{Document}} might not be <a data-lt=
+          "within-scope-manifest">within scope</a> of the linked manifest, by
+          design. For example, a page at <code>https://example.com</code> might
+          link a manifest with <a>scope</a> <code>"/installable/"</code>,
+          allowing the top-level page of the site (which itself is not part of
+          the installable web application) to show installation UI.
+        </p>
+        <p>
+          Implementors should be aware of this possibility. One consequence is
```
marcoscaceres, line 339:
```suggestion
          Implementors need to be aware of this possibility. One consequence is
```

index.html
```diff
@@ -326,18 +326,48 @@ <h2>
...
+          link a manifest with <a>scope</a> <code>"/installable/"</code>,
+          allowing the top-level page of the site (which itself is not part of
+          the installable web application) to show installation UI.
+        </p>
+        <p>
+          Implementors need to be aware of this possibility. One consequence is
+          that if, post install, the user agent opens the document URL in an
+          <a>application context</a>, it would immediately start showing the
+          "out of scope" warning UI (that is recommended in
+          [[[#navigation-scope]]]). Hence, implementors might wish to avoid
```
marcoscaceres, line 343:
```suggestion
          <a>application context</a>, it would start exhibiting the
          "out of scope" behavior (that is recommended in
          [[[#navigation-scope]]]). Hence, implementors might wish to avoid
```

marcoscaceres:
Let's close this one as #1151 landed, and this is covered by that.

PR #792, opened 2019-09-13
Make Service Worker registration work properly (don't use outdated steps)

### LINKED ISSUE #789: Service worker installation uses an algorithm (Start Register) that doesn't exist in SW spec

mgiuca:
This change (choose one):

Commit message:

Make Service Worker registration work properly (don't use outdated steps).

Closes #789.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/792.html" title="Last updated on Nov 12, 2019, 8:21 AM UTC (771b8c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/792/e46df46...mgiuca:771b8c5.html" title="Last updated on Nov 12, 2019, 8:21 AM UTC (771b8c5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4041,11 +4041,6 @@
                 </li>
                 <li>
                   <a data-cite=
-                  "service-workers-1#dom-registrationoptions-type"><dfn for=
-                  "RegistrationOptions">type</dfn></a>
-                </li>
-                <li>
-                  <a data-cite=
                   "service-workers-1#dom-registrationoptions-updateviacache"><dfn for="RegistrationOptions">
                   updateViaCache</dfn></a>
                 </li>
--- a/index.html
+++ b/index.html
@@ -378,26 +378,14 @@
           following algorithm:
         </p>
         <ol>
-          <li>Let <var>manifest</var> and <var>manifest URL</var> be the values
-          that were created during <a>steps to determine installability of the
-          document</a>.
+          <li>Let <var>manifest</var> be the manifest value created during <a>
+            steps to determine installability of the document</a>.
           </li>
-          <li>If <var>manifest URL</var> exists, and the result of running <a>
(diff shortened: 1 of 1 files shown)


mgiuca:
The upstream PR has now merged, so this PR's delta is correct.

kenchris:
Well Jeffrey (talked at TPAC) says to please not do that as he is going to use that innWeb Packaging which uses web app manifest

kenchris:
@jyasskin could you comment?

marcoscaceres:
Sure, but the multi-implementer problem remains. Not sure at this point we want to support this in Gecko, but need to check if WebKit supports it.

jyasskin:
We might use the `serviceworker` field to have navigation to a web package automatically install its service worker before the start page loads, but I think we can re-add the field if/when we do that. It won't hurt us to remove it until then.

marcoscaceres:
Ok, let's proceed with the plan to remove it... we can always bring it back.

mgiuca:
Oops, thanks for closing this out.

PR #797, opened 2019-09-19
Make BeforeInstallPrompt optional

marcoscaceres:
Closes #???

This change (choose one):

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

difference will be.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/797.html" title="Last updated on Sep 19, 2019, 5:39 AM UTC (9577457)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/797/ad4e579...9577457.html" title="Last updated on Sep 19, 2019, 5:39 AM UTC (9577457)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -562,10 +562,14 @@
         </p>
       </section>
     </section>
-    <section>
+    <section class="atrisk">
       <h2>
         Installation Events
       </h2>
+      <p>
+        Installation events and supporting the {{BeforeInstallPrompt}} is
+        OPTIONAL.
+      </p>
       <p>
         DOM events <a>fired</a> by this specification use the <dfn>application
         life-cycle task source</dfn>.


index.html
```diff
@@ -562,10 +562,14 @@ <h3>
         </p>
       </section>
     </section>
-    <section>
+    <section class="atrisk">
       <h2>
         Installation Events
       </h2>
+      <p>
+        Installation events and supporting the {{BeforeInstallPrompt}} is
```
mgiuca: I'm not sure why "installation events" (which includes appinstalled) is optional. I don't think there's anything particularly controversial about appinstalled is there? We should get that supported everywhere.

I'm happy to mark BIP as optional, since it's effectively optional already, as it's up to the user agent to decide when and if to fire it (so any user agent that doesn't implement it could claim conformance already). Marking it as optional just makes that clear (and also makes it legal for window.BeforeInstallPromptEvent to not exist).

index.html
```diff
@@ -562,10 +562,14 @@ <h3>
         </p>
       </section>
     </section>
-    <section>
+    <section class="atrisk">
```
mgiuca: Has there been any more discussion about the "at risk" status of this feature? I don't think we can just deprecate it at this point; too many sites rely on it (though as I say below, I think it's fine for it to be perpetually optional).

aarongustafson: I don‘t have a strong feeling on "at risk". Keeping it optional also leaves the door open for browsers to decide to never do anything with it or to remove support is they have it an are unhappy with it. If we get to the point that no UA is interested in supporting it anymore, we can label it "at risk" and begin to let folks know they may want to consider dropping it from their code.

index.html
aarongustafson: Is there any privacy risk with `appinstalled`?

mgiuca: We (me, @dominickng and @marcoscaceres ) couldn't think of any. You're going to get a CSS media query telling you that the window is standalone as soon as you open the window (which in Chrome Desktop is immediately upon install) anyway, so this doesn't give you any additional information.

aarongustafson: I’m good then. Merge away!

aarongustafson:
@marcoscaceres @mgiuca Any reasons not to merge this?

mgiuca:
LGTM

PR #811, opened 2019-10-11
Various fixes to ImageResource processing algorithms

mgiuca:
This change (choose one):

Commit message:

Normative:

- Removed passing the name of the member to the processing algorithm. It
  just directly takes the array of image resources. This fixes a
  mismatch between the callers (which were already passing the array)
  and the definition of the algorithm.
- Removed definition of steps for processing "src". This was doing
  almost nothing; just inline the parsing of the URL.
- Copy "platform" into the output array.

Formatting:

- Fixed links to "sizes" definition.
- Remove redundant list in processing algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/811.html" title="Last updated on Oct 11, 2019, 4:47 AM UTC (e09a660)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/811/708f4b1...mgiuca:e09a660.html" title="Last updated on Oct 11, 2019, 4:47 AM UTC (e09a660)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1542,13 +1542,12 @@
           </li>
           <li>Set <var>manifest</var>["<a>icons</a>"] to the result of running
           <a>processing ImageResource members</a> given
-          <var>manifest</var>["<a>icons</a>"], <var>manifest URL</var>, and
-          <code>"icons"</code>.
+          <var>manifest</var>["<a>icons</a>"] and <var>manifest URL</var>.
           </li>
           <li>Set <var>manifest</var>["<a>screenshots</a>"] to the result of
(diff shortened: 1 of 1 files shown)


mgiuca:
@aarongustafson Please have a look at this.

I noticed so many issues while reviewing w3c/image-resource#6 which were copied from here, that I think we should fix them upstream first (since they won't be deleted from Manifest for some time).

If you wouldn't mind, can you copy the changes into w3c/image-resource#6 as well? (The commit description summarizes all the changes which you could make by hand.)

PR #832, opened 2019-12-06
Rewrite processing shortcuts algorithm to be more precise

### LINKED ISSUE #831: Steps for processing shortcut member are very vague
I'm reviewing https://chromium-review.googlesource.com/c/chromium/src/+/1941051, which implements parsing of the shortcuts member in Chromium. I'm finding the spec very deficient in describing the parsing. For instance:

- should shortcut item URLs be restricted to be within the scope of the manifest? The spec seems to imply this ("The shortcuts member is an array of ShortcutItems that provide access to key tasks within a web application."), but the processing algorithm doesn't prescribe any checks here. It seems weird to allow arbitrary URLs.
- when should developer warnings be issued when parsing the member?
- when should items be ignored (presumably when `name` or `url` are empty?)

We should clarify these questions in the spec. @aarongustafson @mgiuca @rayankans

mgiuca:
This change (choose one):

To my knowledge, no vendor has implemented this yet, so it is safe to change without getting commitment. This is a very small change to add some new error cases which result in the shortcut items being ignored.

Commit message:

Normative changes:

* shortcuts that meet certain failure conditions are ignored.
  * Checks that name and url are present, URL is valid and within scope of the manifest.

Other changes:

* Clarify that the ShortcutItem url member needs to be within scope.
* Rewrite processing shortcuts algorithm as an ordered list.
* Creates a new array and returns it, rather than discarding the result.
* Explicitly says where developer warnings should be issued.

Closes #831

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/832.html" title="Last updated on Dec 9, 2019, 3:55 AM UTC (36daa47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/832/a9d989d...mgiuca:36daa47.html" title="Last updated on Dec 9, 2019, 3:55 AM UTC (36daa47)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1531,7 +1531,7 @@
           </li>
           <li>Set <var>manifest</var>["<a>shortcuts</a>"] to the result of
           running <a>processing the <code>shortcuts</code> member</a> given
-          <var>manifest URL</var>, and <code>"shortcuts"</code>.
+          <var>manifest</var>["<a>shortcuts</a>"] and <var>manifest URL</var>.
           </li>
           <li>
             <a>Extension point</a>: process any proprietary and/or other
@@ -2345,8 +2345,8 @@
           are given by the following algorithm. The algorithm takes a
           <a data-cite=
           "WEBIDL#sequence-type">sequence</a>&lt;<a>ShortcutItem</a>&gt;
-          <var>shortcuts</var> as an argument. This algorithm returns a
-          <a data-cite=
+          <var>shortcuts</var> and a <a>URL</a> <var>manifest URL</var>. This
+          algorithm returns a <a data-cite=
           "WEBIDL#sequence-type">sequence</a>&lt;<a>ShortcutItem</a>&gt;.
         </p>
         <ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2345,17 +2345,43 @@ <h3>
...
+        <ol>
+          <li>Let <var>processedShortcuts</var> be a new Array object created
+          as if by the expression [].
+          </li>
+          <li>For each <var>shortcut</var> (<a>ShortcutItem</a>) in the
+          sequence:
+            <ol>
+              <li>If <var>shortcut</var>["name"] or <var>shortcut</var>["url"]
+              are undefined, <a>issue a developer warning</a> and
+              <a>continue</a>.
```
dominickng: Undefined or the empty string perhaps?

mgiuca: I thought about this, but for consistency we don't seem to check for empty string anywhere else in the file (e.g., for the `name` field at the top level). Do you think it's worth inconsistent treatment here?

dominickng: We do check for empty strings where it makes sense, like the [Web Share Target spec](https://wicg.github.io/web-share-target/#share_target-member). I think it makes sense here.

We don't do it elsewhere because manifest properties are interpretable without `name`, so we don't want to invalidate the whole manifest when its `name` is empty. In this case, it's seems to be implied that if the `name` of a `ShortcutItem` is the empty string, it's invalid.

aarongustafson: Agreed on warnings for missing and empty. We were actually talking about this internally as we build out the implementation in Windows.

mgiuca: OK. Added "or if name is the empty string". Web Share Target is a reasonable precedent (even though it's not in the manifest spec yet).

mgiuca:
Taking @aarongustafson 's LGTM as still valid since the only change I made he was in agreement.

dominickng:
LGTM, thanks!

marcoscaceres:
Gentle reminder to please avoid using the (mostly) deprecated `data-cite="spec#whatever"` feature. ReSpec's `xref` feature now handles that for us automagically and it avoids bugs. 

If you need to look up a term, you can do directly from ReSpec's Pill's "Search Definitions" option. Or, alternatively, head over to https://respec.org/xref/ to search for terms and "how to cite" them :)

PR #833, opened 2019-12-06
BREAKING CHANGE: Replace "badge" with "monochrome"

### LINKED ISSUE #830: Rename `purpose: badge` to `purpose: monochrome` and expand on it
I'd like to experiment with letting PWAs specify the icon that appears in the "Site controls" notification (the silent notification where you can copy the URL). Since Android expects a grayscale icon for notifications, I'd like to have a purpose defined for this case.

Site controls notifications in Chrome and Firefox

We currently have the `"badge"` purpose. In the spec its defined as:
> A user agent can present this icon where space constraints and/or color requirements differ from those of the application icon. 

I don't feel like this is clear enough for browsers or developers, and no browser currently supports this purpose type. 

---

There's been some discussion on this in multiple issues that I'll try to summarize here.

- The badge type was first proposed in #480, and named after the `badge` field in the [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notification/badge).
- When Shortcuts were added in #768, we had some discussion about how different platforms have different shortcut styles and @mgiuca proposed a `"monochrome"` purpose.
- #795 was spun off to discuss this more, specifically for Shortcuts. 
- I've outlined some potential future uses of a `"monochrome"` purpose in #808, but I just want to focus on Site controls for now.

---

I'd like to propose that we specify monochrome icons as images where all the r,g,b color information is removed and only the alpha is kept. Browsers can then apply whatever color is needed by the OS, or let the OS handle the tinting. This has the added benefit of letting developers potentially reuse their transparent PWA icons by specifying `purpose: "monochrome any"` if the silhouette is distinct. 

If no monochrome icon is available, the browser may choose to create a fallback. For Safari's pinned tabs, [icons are converted automatically](https://github.com/w3c/manifest/issues/480). In this example, Facebook has not specified any special icon.

Safari pinned tabs

In the case of Site controls, it also makes sense for the browser to show a default icon like its logo or the PWA logo, as Chrome and Firefox currently do.

To best handle different use cases, the browser may also add padding to the icon depending on where it is used. In the notification, no padding is needed. If the icon is used in a shortcut on Android, the browser may add padding so that it lines up with platform conventions.

NotWoods:
Closes #830

This change (choose one):

* [ ] Safari (https://bugs.webkit.org/show_bug.cgi?id=204963)
* [x] Chrome (https://bugs.chromium.org/p/chromium/issues/detail?id=1031696)
* [ ] Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1602126)
* [ ] Edge (no public signal)

Commit message:

Following discussion in #830, renaming `purpose: "badge"` to `purpose: "monochrome"` and expanding on the definition.

***
<a href="https://pr-preview.s3.amazonaws.com/NotWoods/manifest/pull/833.html" title="Last updated on Jun 11, 2020, 8:13 AM UTC (b0786b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/833/632df40...NotWoods:b0786b1.html" title="Last updated on Jun 11, 2020, 8:13 AM UTC (b0786b1)">Diff</a>


NotWoods:
@mgiuca ready for review

aarongustafson:
Should this be filed against [`ImageResource`](https://github.com/w3c/image-resource/) instead?

NotWoods:
@aarongustafson I believe that right now, all the "purpose" entries are only in the manifest spec.

aarongustafson:
> @aarongustafson I believe that right now, all the "purpose" entries are only in the manifest spec.

You’re right, I forgot about that. I just got back from a week off, brain is still booting up :-)

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+        <p>
+          If no monochrome icon is defined when an icon with a solid icon must
+          be presented, the user agent MAY show a fallback icon.
+        </p>
+        <p class="note">
+          Designers of <a>monochrome</a> icons are recommended to set all
+          pixels to black and use transparency to create a silhouette of their
+          icon.
+        </p>
+        <p>
```
aarongustafson: Do we want to allow for inversion as well? Windows currently does this when switching between light & dark mode.

NotWoods: The intention is that the spec allows for the icon to be displayed in any color (including white). This is just a recommendation that developers save the asset as an all-black image. The user agent can tint it afterwards.

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
           </div>
         </section>
       </section>
+      <section id="solid-color-icons">
```
```suggestion
      <section>
```

index.html
```diff
@@ -2556,26 +2556,29 @@ <h3>
...
         <dl>
           <dt>
-            <dfn data-lt="badge purpose">badge</dfn>:
+            <dfn data-lt="monochrome purpose">monochrome</dfn>:
           </dt>
           <dd>
-            A user agent can present this icon where space constraints and/or
-            color requirements differ from those of the application icon.
+            A user agent can present this icon where an icon with a
+            <a href="#solid-color-icons">solid color</a> is needed. The
```
marcoscaceres: We should `dfn` solid color somewhere below. 
 
```suggestion
            <a>solid color</a> is needed. The
```

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
         </section>
       </section>
+      <section id="solid-color-icons">
+        <h2>
+          Solid color icons
+        </h2>
+        <p>
+          Some platforms enforce that icons be displayed with a single color
+          or gradient, where only the transparency of the icon can be
+          controlled. As web applications should work across multiple
```
```suggestion
          controlled. As web applications need to work across multiple
```

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+          platforms, it it possible to indicate that an icon can have a
```
```suggestion
          platforms,  it is possible to indicate that an icon can have a
```

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+          user-agent-specified color applied by adding the <a>monochrome</a>
+          purpose. This allows the platform to ensure that the icon looks well
+          integrated with the platform, and even apply different colors and
+          padding in different places throughout the platform.
+        </p>
+        <p>
+          When processing a <a>monochrome</a> icon, the user agent MUST NOT use
```
marcoscaceres: Doesn't the OS do the color conversion defined here? Or do we need an algorithm to convert images before we hand them to the OS... that's gonna be tricky.

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+          the red component, green component, or blue component of a pixel. If
+          it has alpha equal to zero, the user agent SHOULD NOT display it.
+          If it the alpha component is greater than zero, the user agent SHOULD
+          display it with any tint.
```
marcoscaceres: Does the OS or the UA display these?

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+        </p>
+        <p>
+          If no monochrome icon is defined when an icon with a solid icon must
```
```suggestion
          If no <a>monochrome</a> icon is defined when an icon with a solid icon is expected to
```

marcoscaceres:
Before I forget, make sure you also update the ECHIDNA file with the new image:
https://github.com/w3c/manifest/blob/gh-pages/ECHIDNA

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+        </p>
+        <section>
+          <h3>
+            Example usage of solid color icons
+          </h3>
+          <h2 class="icon-title">
+            Usage examples
+          </h2>
+          <div class="icons">
+            <div>
```
marcoscaceres: should this be a `<figure>`?

index.html
```diff
@@ -2810,6 +2813,84 @@ <h2 class="icon-title">
...
+              <img src="images/monochrome-icon-plain.svg" alt=
+              "A black icon over a checkerboard background">
+              <div class="icon-title">
+                Image
+              </div>
+              <div>
+                The base image with no color.
+              </div>
+            </div>
+            <div>
```
marcoscaceres: As above...

index.html
NotWoods: I'll need to investigate what's done here. On Android, it's done automatically for notifications but not shortcuts.

index.html
NotWoods: Here I'm trying to indicate that r,g,b should be ignored and only alpha should be used. Alpha 0 is transparent so it shouldn't show up. I expect the processing to strip out/replace rgb and leave alpha alone, then the OS will display the icon after.

index.html
NotWoods: I'm not sure, I based this on the maskable icon examples present in the spec.

index.html
aarongustafson: Github assigned the comment to a line, which made this slightly confusing. I was not saying this was for authors, but rather for implementors to give them the flexibility to adapt the icon as necessary.

index.html
aarongustafson: Windows converts monochrome icons automatically in the jump list.

aarongustafson:
> Also, just bike-shedding, "monochrome" seems a bit weird as a "purpose". Is that normal nomenclature across different OS? Do we know what Windows uses or iOS?

Agreed it does feel odd. Tintable? Silhouette? Shape? Simple? Mark? Imprint?

index.html
NotWoods: By inversion, are you referring to making the solid areas transparent and the transparent areas solid? If so I agree that would be good to add in.

If you mean replace black with white, then I try to express this in the spec by indicating that the r,g,b values can be replaced by the user agent. Do you have any suggestions for making this more clear?

NotWoods:
Some other ideas for the purpose:
- `monochrome` (used by [Android](https://material.io/design/platform-guidance/android-notifications.html#anatomy-of-a-notification))
- `monochromatic`
- `single_color` (used by [Safari](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html))
- `shady`

`shape` seems like it could be confused with `maskable`. \
`simple` and `mark` aren't clear terms, so I think they have the same issues as badge.

I lean towards some variation of "single color" (monochrome/monochromatic/single_color) since that has the most clarity.

marcoscaceres:
I'm kinda like "single-color"... it doesn't imply that it will be monochrome... just a single color.

NotWoods:
Technically monochrome means single color and not greyscale 😉 Though `single_color` is probably more intuitive.

marcoscaceres:
TIL (just wikipedia-educated myself). I've always thought of it as greyscale. I checked what Google images returns for monochrome, and it seems the zeitgeist appears to be "grayscale".

index.html
```diff
@@ -2582,7 +2582,7 @@ <h3>
           <a>purpose</a>.
         </p>
         <p class="note">
-          For example, an icon with purpose "<a>monochrome</a>" could be used
+          For example, an icon with purpose "<a>single_color</a>" could be used
```
marcoscaceres: Because we like to confuse everyone, I think we want a “-“ instead of “_” (see also orientation).

index.html
mgiuca: I think both of these are already implied by the spec. If a user agent wanted to, say, display the icon as pink-on-green, it would make the pixels pink, and then display the icon on a green background. If it wanted to invert that color scheme, it would make the pixels green and then display the icon on a pink background.

The only thing that wouldn't be supported would be if the UA, for some reason, wanted to make the "transparent" part of the image a solid colour (say, pink), and then have the "opaque" part of the image show through some background underneath. I would say we explicitly do not want to support that, since then we're kinda messing with the brand logos in a way they don't necessarily want.

index.html
```diff
@@ -2836,6 +2839,84 @@ <h2 class="icon-title">
...
+        <p>
+          When processing a <a>single-color</a> icon, the user agent MUST NOT use
+          the red component, green component, or blue component of a pixel. If
+          it has alpha equal to zero, the user agent SHOULD NOT display it.
+          If it the alpha component is greater than zero, the user agent SHOULD
+          display it with any tint.
+        </p>
+        <p>
+          If no single color icon is defined when an icon with a solid color is
+          expected to be presented, the user agent MAY show a fallback icon.
```
mgiuca: It's unclear what this means. Does it mean it MAY fall back to using the standard icon? Or it MAY supply its own default icon?

I'm not sure we need to say this at all. User agents can show any icon they want wherever they want, and don't need permission of the spec. We should just focus on how user agents are expected to treat these monochrome icons, if supplied, and not start dictating their UI more broadly.

NotWoods: MAY supply its own default icon. For example, in the PWA site controls notification image I give in the issue, the browser logo (Chrome) or PWA logo (Firefox) could continue to be used if the manifest has not supplied an icon with purpose: single-color

NotWoods:
My intention is to allow for anti-aliasing, so I'll make adjustments using your suggestions. 

I don't want to use greyscale, because that implies that grey is treated as 50% opacity when the colour has no effect on the displayed image. I'm happy with monochrome but I can also see that developers will read it as "black and white" rather than all one color (pink, blue, etc).

index.html
NotWoods: Since maskable doesn't have any similar fallback text I'll remove this section.

NotWoods:
I was able to find [monochrome monitors](https://en.wikipedia.org/wiki/Monochrome_monitor), such as green text on black backgrounds. (Confusingly, monochrome monitors are specified as two colours - foreground and background.) 

I don't think that using single-color means that we are dumbing down the term. 
Many, many examples of monochrome are greyscale and black and white. "Single colour" is a bit more specific, as the displayed image can be displayed in any single colour.

NotWoods:
Throwing another name idea onto the pile: `stamp`. The website provides the stamp and the user agent provides the "ink pad" used to color it. No black-and-white association.

index.html
NotWoods: Using `<figure>` messes with the styling since extra properties are injected, so I'll leave this as `<div>` for now

leolux:
How about setting a color value for the notification color:?
image

NotWoods:
The color value is already set by theme_color in Firefox. I don't think the spec can dictate that UI, so it's up to the other browser teams to do something similar. We don't need a new color field for the notification.

index.html
```diff
@@ -2625,9 +2628,9 @@ <h3>
           If an icon contains multiple purposes, it could be used for any of
           those purposes. If none of the stated purposes are recognized, the
           icon is totally ignored. For example, if an icon has purpose
-          <code>"badge fizzbuzz"</code>, then it could be used as a badge, but
-          if an icon has just the purpose <code>"fizzbuzz"</code>, then it will
-          be ignored.
+          <code>"single-color fizzbuzz"</code>, then it could be used as a
+          single color icon, but if an icon has just the purpose
```
kenchris, line 2632: the invalid purpose?

index.html
```diff
@@ -2587,26 +2587,29 @@ <h3>
...
-            <dfn data-lt="badge purpose">badge</dfn>:
+            <dfn data-lt="single-color purpose">single-color</dfn>:
           </dt>
           <dd>
-            A user agent can present this icon where space constraints and/or
-            color requirements differ from those of the application icon.
+            A user agent can present this icon where an icon with a
+            <a>single color</a> is needed. The
+            color information in the icon is discarded and only the alpha data
+            is used.
```
kenchris, line 2609: So it is basically being used as a mask that can be used with any color by the user agent/platform. I think it might make sense to spell that out

index.html
```diff
@@ -2855,6 +2858,87 @@ <h2 class="icon-title">
...
+      <section>
+        <h2>
+          Single color icons
+        </h2>
+        <p>
+          Some platforms enforce that icons be displayed with a <dfn>single
+          color</dfn>, where only the transparency of the icon can be
+          controlled. As web applications need to across multiple
+          platforms, it is possible to indicate that an icon can have a
+          user-agent-specified color applied by adding the <a>single-color</a>
```
kenchris, line 2870: User Agents can use different colors depending on where they display it - they could even use a gradient - it is really about masking that allows them to do so

kenchris: For this reason the name "single-color" might be wrong.

index.html
```diff
@@ -2855,6 +2858,87 @@ <h2 class="icon-title">
...
+        <p>
+          When processing a <a>single-color</a> icon, the user agent MUST NOT
+          use the red component, green component, or blue component of a pixel.
+          The user agent SHOULD display each pixel with its original alpha
+          value, but with a red, green, and blue value of the user agent's
+          choosing. It is RECOMMENDED that the user agent use the same color
+          value for all pixels.
+        </p>
+        <p class="note">
+          Designers of <a>single-color</a> icons are recommended to set all
```
kenchris, line 2884: This recommendation doesn't help in any way - why not use "could" or so instead of "recommended"

index.html
NotWoods: Do you have any preferences on the names previously suggested? (`monochrome`, `greyscale`, `stamp`)

kenchris: The **purpose** of this icon is that it will be use as **monochrome** - sounds wrong.

Maybe silhouette?

kenchris: Apart from this I like monochrome more than single-color

index.html
```diff
@@ -2741,7 +2745,7 @@ <h2>
         </p>
         <p>
           If the icon contains transparent pixels, the user agent MUST
-          composite the icon onto a solid color (eg. white) of the user agent's
+          composite the icon onto a single color (eg. white) of the user agent's
```
kenchris, line 2748: but it doesn't have to be a single color :)

index.html
kenchris: Maybe "mask" or "image-mask" might be the most descriptive

kenchris: cut-out

1. a shape cut out of board or another material.
    "a life-size cut-out"
2. a hole cut in something for decoration or to allow the insertion of something else.
     "the mains output socket is fitted into a suitable cut-out on the rear panel"

NotWoods: I think that mask, image-mask, and cut-out could be confused with maskable. Image masking is like cutting out parts of the image that aren't inside the mask. I'd like to steer away from mask/cut terminology.

Similarily, I think you could argue "the **purpose** of this icon is that it will be used as **maskable**" sounds just as odd. I think you could rephrase it as:
- The purpose of this icon is that it is maskable
- The purpose of this icon is that it is monochrome

I like silhouette but it doesn't fit that dichotomy. I'm going to switch back to monochrome as that has the most consensus.

NotWoods:
I've rebased and switched back to the "monochrome" name. Let me know if there are any more changes that need to be made.

index.html
```diff
@@ -2203,26 +2203,30 @@ <h3>
...
+            <dfn data-lt="monochrome purpose">monochrome</dfn>:
           </dt>
           <dd>
-            A user agent can present this icon where space constraints and/or
-            color requirements differ from those of the application icon.
+            A user agent can present this icon where a
+            <a href="#monochrome-icons-and-solid-fills">monochrome icon with a
+            solid fill</a> is needed. The color information in the icon is
+            discarded and only the alpha data is used. The icon can then be
+            used user agent like a mask over any colored background.
```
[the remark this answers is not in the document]
NotWoods: I've tried to limit the usage of the term "mask", but the actual image processing steps is still masking, just using the icon as the mask instead of the image to mask. Because of this, I think rephrasing would hurt readability.

index.html
```diff
@@ -2439,6 +2444,87 @@ <h2 class="icon-title">
...
+        </p>
+        <p class="note">
+          Designers of <a>monochrome</a> icons could set all pixels to black
+          and only use transparency to create a silhouette of their icon.
+        </p>
+        <p>
+          The user agent MAY enlarge the icon by adding additional padding.
+        </p>
+        <p>
+          The user agent MAY add a background of any color behind transparent
```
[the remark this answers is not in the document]
marcoscaceres: Switched it, as SHOULD needs to apply to a use agent.

index.html
```diff
@@ -2439,6 +2444,87 @@ <h2 class="icon-title">
...
+          indicate that an icon can have an user-agent-specified color applied
+          by adding the <a>monochrome</a> purpose. This allows the platform
+          to ensure that the icon looks well integrated with the platform, and
+          even apply different colors and padding in different places
+          throughout the platform.
+        </p>
+        <p>
+          When processing a <a>monochrome</a> icon, the user agent MUST NOT
+          use the red component, green component, or blue component of a pixel.
+          The user agent SHOULD display each pixel with its original alpha
```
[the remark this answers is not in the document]
marcoscaceres: Ok cool. thanks for clarifying!

marcoscaceres:
@dominickng @mgiuca, or other potential implementers, we are planning to support this in Firefox, but we'd like a second implementer before merging into the spec. 

Is this something you'd be interested in supporting at some point (and thus something you would also liked merged into the spec)?

mgiuca:
I'm not sure I can say for sure that we're going to use this, but I'd like us to wherever it makes sense (e.g., in notifications).

We currently parse the badge purpose, but we don't seem to use it anywhere (it might be that we use it on our server). I need to confirm that it isn't going to break any sites if we stop parsing "badge".

If I can confirm that, then I would be happy to change our parser to parse "monochrome" instead. I can't commit to actually using it, but I don't think that should stop us landing this change (it'd be going from one unused-in-Chrome field to another, so the "two implementor" rule isn't being any more violated than it is now). Perhaps as a separate issue we would look at offloading it from the CR version of the spec if there aren't two implementations at that time, but I don't think it should stop this landing.

marcoscaceres:
Ok, sounds like a plan. And yeah, we can mark it "at risk" during CR if need be.

index.html
```diff
@@ -2401,6 +2405,88 @@ <h2 class="icon-title">
           </div>
         </section>
       </section>
+      <section>
+        <h2>
+          Monochrome icons and solid fills
+        </h2>
+        <p>
```
marcoscaceres, line 2412:
```suggestion
        <p class="issue atrisk" data-number="905"></p>
        <p>
```

marcoscaceres:
@NotWoods, just need you to rejoin the working group: https://www.w3.org/2004/01/pp-impl/114929/join

mgiuca:
Confirmed that Chrome does **not** use purpose: badge anywhere, including on the server-side infrastructure. From our perspective, we are good to rename it and change its semantics. LGTM.

Psoffie009:
> Closes #830
> 
> This change (choose one):
> 
> * [x]  Breaks existing normative behavior (please add label "breaking")
> * [ ]  Adds new normative requirements
> * [x]  Adds new normative recommendations or optional items
> * [ ]  Makes only editorial changes (only changes informative sections, or
>   changes normative sections without changing behavior)
> * [ ]  Is a "chore" (metadata, formatting, fixing warnings, etc).
> 
> Implementation commitment (delete if not making normative changes):
> 
> * [ ]  Safari (https://bugs.webkit.org/show_bug.cgi?id=204963)
> * [x]  Chrome (https://bugs.chromium.org/p/chromium/issues/detail?id=1031696)
> * [ ]  Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1602126)
> * [ ]  Edge (no public signal)
> 
> Commit message:
> 
> Following discussion in #830, renaming `purpose: "badge"` to `purpose: "monochrome"` and expanding on the definition.
> 
> [Preview](https://pr-preview.s3.amazonaws.com/NotWoods/manifest/pull/833.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/manifest/833/632df40...NotWoods:b0786b1.html)

Psoffie009:
> > Closes #830
> > This change (choose one):
> > 
> > * [x]   Breaks existing normative behavior (please add label "breaking")
> > * [x]   Adds new normative requirements
> > * [x]   Adds new normative recommendations or optional items
> > * [x]   Makes only editorial changes (only changes informative sections, or
> >   changes normative sections without changing behavior)
> > * [x]   Is a "chore" (metadata, formatting, fixing warnings, etc).
> > 
> > Implementation commitment (delete if not making normative changes):
> > 
> > * [x]   Safari (https://bugs.webkit.org/show_bug.cgi?id=204963)
> > * [x]   Chrome (https://bugs.chromium.org/p/chromium/issues/detail?id=1031696)
> > * [x]   Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1602126)
> > * [x]   Edge (no public signal)
> > 
> > Commit message:
> > Following discussion in #830, renaming `purpose: "badge"` to `purpose: "monochrome"` and expanding on the definition.
> > [Preview](https://pr-preview.s3.amazonaws.com/NotWoods/manifest/pull/833.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/manifest/833/632df40...NotWoods:b0786b1.html)

Psoffie009:
> Before I forget, make sure you also update the ECHIDNA file with the new image:
> https://github.com/w3c/manifest/blob/gh-pages/ECHIDNA

PR #692, opened 2018-06-14
Separate BeforeInstallPrompt from automatic prompt

mgiuca:
***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/692.html" title="Last updated on Jul 9, 2018, 6:59 AM GMT (144c2f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/692/cc53c75...mgiuca:144c2f5.html" title="Last updated on Jul 9, 2018, 6:59 AM GMT (144c2f5)">Diff</a>

PR #836, opened 2019-12-13
Remove beforeinstallprompt and appinstalled events.

### LINKED ISSUE #578: Provide a way to find out the result of the install prompt without prompt()
There are currently two ways to find out whether the user installed your app or not:

- The `appinstalled` event. (Only gives you a positive answer, doesn't tell you if they cancelled the install prompt.)
- The `beforeinstallprompt` event: call `prompt()`. The result of `prompt` tells you whether the user accepted.

Chrome's implementation has an additional `userChoice` attribute (on the `BeforeInstallPromptEvent`) that lets you passively detect prompt acceptance/denial without calling `prompt()`.

There's one problem with using `prompt()` instead of a passive `userChoice`: it may produce slightly different prompting behaviour.

As the spec currently stands, calling `prompt()` in the `beforeinstallprompt` event handler is a no-op; it doesn't change the behaviour at all (which defaults to prompting), and therefore, `prompt()` in the event handler is equivalent to `userChoice`. However, #576 calls for a change to allow user agents to not prompt by default, but show a prompt if `prompt()` is called. This means developers using `prompt()` as a substitute for `userChoice` would suddenly get non-default behaviour. There would be no way for a site to passively detect prompt acceptance/denial without inadvertently forcing the prompt to be shown.

Some possible approaches here (presented without opinion):
- "We aren't going to action #576, and therefore `prompt()` is just as good as `userChoice`".
- "This use case is not important enough; if you care enough about collecting data, then you can decide for yourself when the prompt should be shown and not rely on the default behaviour. Or use `appinstalled` to collect data."
- "We do want a way to let developers passively collect this data. Let's add Chrome's `userChoice`."
- "We do want a way to let developers passively collect this data. But `userChoice` (being an attribute promise) is weird; let's design something else."

Thoughts?

### LINKED ISSUE #627: Allowing only site-triggered install prompt
Hi folks, do you really like this idea, that a user-agent will "analyze" a website somehow and decide, if it is appropriate to show an install prompt? It can be disturbing, if the "analysis" is wrong. I also hate the fact, that a browser makes such decissions (browser manufacturers can show the prompt more often for websites, which pay them, nobody will ever find out). I think that the browser should be as "invisible" as possible, and each action should come from the user, or from the website with a users permit.

I am using the following code to be able to trigger an install prompt.
    
    window.addEventListener("beforeinstallprompt", function(e){  e.preventDefault();  myEvent=e;  });
    if ("serviceWorker" in navigator)  navigator.serviceWorker.register("sw.js", { scope: "./" });  
    //  I call myEvent.prompt();  later, when user clicks Install

The problem I have is, that if both the website and the user want the site to be installed, the user-agent itself becomes an obstacle (by not dispatching "beforeinstallprompt"). What about replacing it all with this:

    navigator.install();

It can throw an exception, when installing is not appropriate. If you think there is a risk of "attack" by calling it too often, we can deal with it the same way we deal with alert(...). Imagine if alert(...); worked only on Mondays and only between 12th and 37th minute of each hour. That is how I see the current draft now.

### LINKED ISSUE #691: Clarify circumstances in which BeforeInstallPromptEvent.prompt() may be called
Currently a few issues:

1. Chrome requires prompt() to be called on user gesture. This isn't in the spec. Is this something we can just require in the spec, or can we say that the user agent MAY require a user gesture?
2. Neither Chrome nor the spec requires preventDefault() to be called before prompt(). Consider this code (assuming the #install_app button is disabled on load):

```js
addEventListener('beforeinstallprompt', e => {
  document.querySelector('#install_app').disabled = false;
  stashed_bip = e;
});

document.querySelector('#install_app').addEventListener('click', async () => {
  console.log(await stashed_bip.prompt());
});
```

It doesn't call `preventDefault`, which means if you have a standards-compliant browser that shows an automated prompt, the above code works this way:

1. The UA fires the `beforeinstallprompt` event.
2. The `beforeinstallprompt` event handler enables the "Install" button.
3. The UA shows an automated prompt, and the user accepts or dismisses it.
4. The user clicks the still-active "Install" button.
5. `prompt()` immediately resolves with the result of the automated install banner, without showing a second prompt.

This is a bad user experience because they are left with an enabled, but ineffective Install button.

However, the above code runs perfectly well in a browser that does not use automated install prompts, such as Chrome.

Suggested changes:

1. Require a user gesture for `prompt()`. (Are there any UAs that implement `BeforeInstallPromptEvent.prompt()` that don't require a user gesture?)
2. I think we need to disassociate the automated install prompt with BeforeInstallPromptEvent. UAs are still allowed to show an automated install prompt, and may fire BIP at the same time as showing the automated prompt. But calling `prompt()` from inside BIP would require the UA to show a second prompt, even if an automated one was already shown. `prompt()` wouldn't just return the result of the previously-shown automated prompt. You would still be able to use `preventDefault` to cancel *some* automated prompts (see below). There would be no need to use `preventDefault` before calling `prompt`, but sites may still wish to do this.

This gives UAs a couple of options:

1. Don't ever show an automated prompt; dispatch the BeforeInstallPromptEvent, requiring the site to opt in to showing a prompt.
2. Show an automated prompt, and do not implement BeforeInstallPromptEvent. The prompt is considered part of the user agent's user interface, not under the control of the developer (no different to any other affordance offered by a user agent). **Firefox today**
3. Show an automated prompt, and independently, dispatch the BeforeInstallPromptEvent (perhaps at different times or under different circumstances). This is just a combination of the previous two options. The automated prompt might look quite different to the site-triggered prompt (e.g., being smaller or more subtle). In this approach, using preventDefault on the BeforeInstallPromptEvent would have no effect, since it is independent of the automated install prompt. **Chrome today**
4. Dispatch the BeforeInstallPromptEvent, and if it is not cancelled, show an automated prompt. In this approach, using preventDefault would stop the automatic prompt from appearing. User agents that adopt this approach would need to show a second prompt if the event's prompt() method is called (otherwise, the site might show an "Install" button that does nothing when clicked). **Chrome from 2017**

**Edit**: Updated the descriptions above to be clearer (based on text in the spec draft).

The above suggestion, I believe, will be compatible with Chrome and Firefox today, and avoid the awkward compat issue with any future browsers (or old versions of Chrome) highlighted above.

@jungkees does Samsung Browser behave in a way compatible with the described changes?

@dominickng @marcoscaceres @kenchris 

(I can rework the described changes in the form of a pull request if it's unclear.)

### LINKED ISSUE #835: beforeinstallprompt : Prompting user makes it to hard to discern whether the user truly wanted to "install" a web app
Prompting the user to "install" a website (via `beforeinstallprompt`) makes it hard to tell whether the user actually wanted to save the website for later use, or the user just wanted to pass through the prompts encountered before using / while the web app.

If we're trying to gate any feature only to installed apps, then it goes directly against the goal of differentiating websites that the user trusts; by the virtue of eagerly asking the user to install an app, we've decreased the likelihood that the user intentionally did so.

marcoscaceres:
Closes #835 
Closes #843
Closes #692

These should be relocated: 
Closes #627 
Closes #578 
Closes #691 

This change (choose one):

* [X] Safari - never implemented.
* [ ] Chrome (link to issue)
* [X] Firefox - never shipped BIP, remove [onappinstalled](https://bugzilla.mozilla.org/show_bug.cgi?id=1625384). 
* [ ] Edge (public signal)

Web platform tests:
*  [X] [mark tests as tentative](https://github.com/web-platform-tests/wpt/pull/22521)

Commit message:

As discussed in #835, the proposal is to remove "beforeinstallprompt" and leave it to browsers to provide better UI to surface installability of web applications. It also removes the `window.appinstalled` event handler. This also removes install prompting and the installation algorithm itself.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 1, 2020, 12:52 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fmanifest%2Fd41440489d02ab7b64fff9b4923bf4072aab9811%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 30000 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/manifest%23836.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3545,10 +3545,6 @@
               <a data-cite="HTML#the-element's-base-url"><dfn>the element's
               base URL</dfn></a>.
             </li>
-            <li>
-              <a data-cite="HTML#queue-a-post-load-task"><dfn>queue a post-load
-              task</dfn></a>
-            </li>
           </ul>
         </li>
         <li>[[INFRA]] defines the following terms:
--- a/index.html
+++ b/index.html
@@ -3253,6 +3253,17 @@
           "https://wicg.io"><abbr title=
           "Web Incubator Community Group">WICG</abbr></a>.
         </dd>
+        <dt>
+          `share_target` member
+        </dt>
+        <dd>
+          The `share_target` member registers a web application as "target" for
+          share actions (e.g., for sharing a text, a URL, or a file). The
+          `share_target` member is part of the <a href=
+          "https://wicg.github.io/web-share-target/">Web Share Target</a>
+          specification, being incubated at the <a href=
+          "https://wicg.io">WICG</a>.
+        </dd>
       </dl>
     </section>
     <section class="appendix">
--- a/index.html
+++ b/index.html
@@ -329,8 +329,8 @@
             </li>
           </ul>
         </li>
-        <li>Otherwise, the {{Document}} is considered <a>installable</a> (at
-        the user agent's discretion; see [[[#installability-signals]]]).
+        <li>Otherwise, at its discretion, the user agent MAY consider the
+        {{Document}} <a>installable</a> (see [[[#installability-signals]]]).
         </li>
       </ol>
       <section>
@@ -382,7 +382,7 @@
         </h3>
         <p>
           It is RECOMMENDED that UI that affords the end user the ability to
-          <a>install</a> a web application also allow inspecting the icon,
+          <a>install</a> a web application also allows inspecting the icon,
           name, <a>start URL</a>, origin, etc. pertaining to a web application.
           This is to give an end-user an opportunity to make a conscious
           decision to approve, and possibly modify, the information pertaining
@@ -3462,18 +3462,6 @@
             <li>
               <a data-cite=
               "ECMA-402#sec-canonicalizelanguagetag"><dfn>CanonicalizeLanguageTag</dfn></a>
-            </li>
-          </ul>
-        </li>
-        <li>[[DOM]] defines the following terms:
-          <ul>
-            <li>
-              <a data-cite="DOM#event"><dfn><code>Event</code>
-              interface</dfn></a>
-            </li>
-            <li>
-              <a data-cite="DOM#concept-event-fire"><dfn data-lt=
-              "fire|fired|firing">Fire an event</dfn></a>
             </li>
           </ul>
         </li>
--- a/index.html
+++ b/index.html
@@ -3246,7 +3246,7 @@
           The `BeforeInstallPrompt` event and `window.onappinstalled` event
           were originally part of this specification. However, they were
           <a href="https://github.com/w3c/manifest/pull/836">removed from the
-          specification</a> because they didn't not have support from two or
+          specification</a> because they did not have support from two or
           more implementers. You can now find them in the <a href=
           "https://github.com/WICG/beforeinstallprompt">`BeforeInstallPrompt`
           event and `window.onappinstalled` repository</a> at the <a href=
--- a/index.html
+++ b/index.html
@@ -398,120 +398,19 @@
         </ol>
       </section>
       <section>
-        <!-- TODO(mgiuca): Move this section up above Installation process. (In
-             a separate PR; otherwise it would be too hard to review.) -->
-        <h2>
-          Install prompts
-        </h2>
-        <p>
-          There are multiple ways that the installation process can be
-          triggered:
-        </p>
-        <ul>
-          <li>An end-user can <dfn data-lt="manual installation">manually</dfn>
-          trigger the installation process through the user agent's
-            <abbr title="User Interface">UI</abbr>, directly invoking the steps
-            to <a>present an install prompt</a>.
-          </li>
-          <li>The installation process can occur through an <dfn>automated
-          install prompt</dfn>: that is, a UI that the user agent presents to
-          the user when, for instance, there are sufficient <a>installability
-          signals</a> to warrant <a>installation</a> of the web application.
-          </li>
-          <li>The installation process can occur through a <dfn>site-triggered
-          install prompt</dfn>: the site can programmatically request that the
-          user agent present an install prompt to the user. The user agent MAY
-          restrict the availability of this feature to cases where, for
-          instance, there are sufficient <a>installability signals</a> to
-          warrant <a>installation</a> of the web application.
-          </li>
-        </ul>
-        <p>
-          In any case, the user agent MUST NOT <a>present an install prompt</a>
-          if the document is not <a>installable</a>.
-        </p>
-        <p>
-          Prior to presenting an <a>automated install prompt</a>, a user agent
-          MUST run the <a>steps to notify that an install prompt is
-          available</a>, to give the site the opportunity to prevent the
-          default action (which is to install the application). Alternatively,
-          the user agent MAY, at any time (only if the document is
-          <a>installable</a>), run the <a>steps to notify that an install
-          prompt is available</a> at any time, giving the site the opportunity
-          to show a <a>site-triggered install prompt</a> without automatically
-          showing the prompt.
-        </p>
-        <p>
-          To <dfn data-lt=
-          "presenting an install prompt|presentation of the install prompt">present
-          an install prompt</dfn>:
-        </p>
-        <ol>
-          <li>Show some user-agent-specific UI, asking the user whether to
-          proceed with installing the app. See <a href=
-          "#installation-sec">privacy and security considerations</a> for
-          recommendations relating to this UI. The <var>result</var> of this
-          choice is either <a data-link-for=
-          "AppBannerPromptOutcome">accepted</a> or <a data-link-for=
-          "AppBannerPromptOutcome">dismissed</a>.
-          </li>
-          <li>Return <var>result</var>, and <a>in parallel</a>:
-            <ol>
-              <li>If <var>result</var> is <a data-link-for=
-              "AppBannerPromptOutcome">accepted</a>, run the <a>steps to
-              install the web application</a>.
-              </li>
-            </ol>
-          </li>
-        </ol>
-        <p>
-          The <dfn>steps to notify that an install prompt is available</dfn>
-          are given by the following algorithm:
-        </p>
-        <ol>
-          <li>Wait until the {{Document}} of the <a>top-level browsing
-          context</a> is <a>completely loaded</a>.
-          </li>
-          <li>If there is already an <a data-lt=
-          "present an install prompt">install prompt being presented</a> or if
-          the <a>steps to install the web application</a> are currently being
-          executed, then abort this step.
-          </li>
-          <li>
-            <a>Queue a task</a> on the <a>application life-cycle task
-            source</a> to do the following:
-            <ol>
-              <li>Let <var>event</var> be a newly constructed
-              <a>BeforeInstallPromptEvent</a> named
-              <code>beforeinstallprompt</code>, with its
-              <code>cancelable</code> attribute initialized to true.
-              </li>
-              <li>Let <var>mayShowPrompt</var> be the result of <a>firing</a>
-              <var>event</var> at the {{Window}} object of the <a>top-level
-              browsing context</a>.
-              </li>
-              <li>If <var>mayShowPrompt</var> is true, then the user agent MAY,
-              <a>in parallel</a>, <a>request to present an install prompt</a>
-              with <var>event</var>.
-              </li>
-            </ol>
-          </li>
-        </ol>
-      </section>
-      <section>
         <h3 id="installation-sec">
           Privacy and security considerations
         </h3>
         <p>
-          During the <a>presentation of the install prompt</a>, it is
-          RECOMMENDED that the user agent allow the end-user to inspect the
-          icon, name, <a>start URL</a>, origin, etc. pertaining to a web
-          application. This is to give an end-user an opportunity to make a
-          conscious decision to approve, and possibly modify, the information
-          pertaining to the web application before installing it. This also
-          gives the end-user an opportunity to discern if the web application
-          is spoofing another web application, by, for example, using an
-          unexpected icon or name.
+          It is RECOMMENDED that UI that affords the end user the ability to
+          <a>install</a> a web application also allow inspecting the icon,
+          name, <a>start URL</a>, origin, etc. pertaining to a web application.
+          This is to give an end-user an opportunity to make a conscious
+          decision to approve, and possibly modify, the information pertaining
+          to the web application before installing it. This also gives the
+          end-user an opportunity to discern if the web application is spoofing
+          another web application, by, for example, using an unexpected icon or
+          name.
         </p>
         <p>
           It is RECOMMENDED that user agents prevent other applications from
@@ -584,203 +483,21 @@
         Installation Events
       </h2>
       <p>
-        Installation events and supporting the {{BeforeInstallPrompt}} is
-        OPTIONAL.
-      </p>
-      <p>
         DOM events <a>fired</a> by this specification use the <dfn>application
         life-cycle task source</dfn>.
       </p>
-      <section data-dfn-for="BeforeInstallPromptEvent" data-link-for=
-      "BeforeInstallPromptEvent">
-        <h3>
-          <dfn>BeforeInstallPromptEvent</dfn> Interface
-        </h3>
-        <div class="note">
-          The <a>beforeinstallprompt</a> event is somewhat misnamed, as it does
-          not necessarily signal that an <a>automated install prompt</a> will
-          follow (depending on the user agent, it might just be giving the site
-          the ability to trigger an install prompt). It is so named for
-          historical reasons.
-        </div>
-        <pre class="idl" data-cite="DOM">
-          [Exposed=Window]
-          interface BeforeInstallPromptEvent : Event {
-            constructor(DOMString type, optional EventInit eventInitDict = {});
-            Promise&lt;PromptResponseObject&gt; prompt();
-          };
-
-          dictionary PromptResponseObject {
-            AppBannerPromptOutcome userChoice;
-          };
-
-          enum AppBannerPromptOutcome {
-            "accepted",
-            "dismissed"
-          };
-        </pre>
-        <p>
-          The <a>BeforeInstallPromptEvent</a> is dispatched when the site is
-          allowed to present a <a>site-triggered install prompt</a>, or prior
-          to the user agent presenting an <a>automated install prompt</a>. It
-          allows the site to cancel the <a>automated install prompt</a>, as
-          well as manually present the <a>site-triggered install prompt</a>.
-        </p>
-        <div class="note">
-          If the <a>BeforeInstallPromptEvent</a> is <em>not</em> cancelled, the
-          user agent is allowed to <a>present an install prompt</a>
-          (specifically, an <a>automated install prompt</a>) to the end-user.
-          Canceling the default action (via <a data-cite=
-          "DOM#dom-event-preventdefault">preventDefault</a>) prevents the user
-          agent from <a>presenting an install prompt</a>. The user agent is
-          free to run <a>steps to notify that an install prompt is
-          available</a> again at a later time.
-        </div>
-        <p data-dfn-for="PromptResponseObject">
-          The <dfn>PromptResponseObject</dfn> contains the result of calling
-          <a data-lt="BeforeInstallPromptEvent.prompt()">prompt()</a>. It
-          contains one member, <dfn data-link-for=
-          "PromptResponseObject">userChoice</dfn>, which states the user's
-          chosen outcome.
-        </p>
-        <p>
-          An instance of a <a>BeforeInstallPromptEvent</a> has the following
-          internal slots:
-        </p>
-        <dl>
-          <dt>
-            <dfn>[[\didPrompt]]</dfn>
-          </dt>
-          <dd>
-            A boolean, initially <code>false</code>. Represents whether this
-            event was used to <a>present an install prompt</a> to the end-user.
-          </dd>
-          <dt>
-            <dfn>[[\userResponsePromise]]</dfn>
-          </dt>
-          <dd>
-            A promise that represents the outcome of <a>presenting an install
-            prompt</a>.
-          </dd>
-        </dl>
-        <section>
-          <h4>
-            <code>prompt()</code> method
-          </h4>
-          <p>
-            The <dfn>prompt</dfn> method, when called, runs the following
-            steps:
-          </p>
-          <ol>
-            <li>If <var>this</var>.<a>[[\userResponsePromise]]</a> is pending:
-              <ol>
-                <li>If this event's <a data-cite=
-                "DOM#dom-event-istrusted"><code>isTrusted</code></a> attribute
-                is <code>false</code>, reject
-                <var>this</var>.<a>[[\userResponsePromise]]</a> with
-                {{"NotAllowedError"}}, optionally informing the developer that
-                untrusted events can't call <code>prompt()</code>.
-                </li>
-                <li>Else if <var>this</var>.<a>[[\didPrompt]]</a> is
-                <code>false</code>, set <var>this</var>.<a>[[\didPrompt]]</a>
-                to <code>true</code>, then <a>in parallel</a>, <a>request to
-                present an install prompt</a> with this event. Wait, possibly
-                indefinitely, for the end-user to make a choice.
-                </li>
-              </ol>
-            </li>
-            <li>Return <var>this</var>.<a>[[\userResponsePromise]]</a>.
-            </li>
-          </ol>
-          <p>
-            To <dfn data-noexport="">request to present an install prompt</dfn>
-            with <a>BeforeInstallPromptEvent</a> <var>event</var>:
-          </p>
-          <ol>
-            <li>
-              <a>Present an install prompt</a> and let <var>outcome</var> be
-              the result.
-            </li>
-            <li>Resolve <var>event</var>.<a>[[\userResponsePromise]]</a> with a
-            newly created <a>PromptResponseObject</a> whose <a data-link-for=
-            "PromptResponseObject">userChoice</a> member is the value of <var>
-              outcome</var>.
-            </li>
-          </ol>
-        </section>
-        <section class="informative">
-          <h4>
-            Usage example
-          </h4>
-          <p>
-            This example shows how one might prevent an automated install
-            prompt from showing until the user clicks a button to show a
-            <a>site-triggered install prompt</a>. In this way, the site can
-            leave installation at the user's discretion (rather than prompting
-            at an arbitrary time), whilst still providing a prominent UI to do
-            so.
-          </p>
-          <pre class="example" title=
-          "Using beforeinstallprompt to present an install button">
-            window.addEventListener("beforeinstallprompt", event =&gt; {
-              // Suppress automatic prompting.
-              event.preventDefault();
-
-              // Show the (disabled-by-default) install button. This button
-              // resolves the installButtonClicked promise when clicked.
-              installButton.disabled = false;
-
-              // Wait for the user to click the button.
-              installButton.addEventListener("click", async e =&gt; {
-                // The prompt() method can only be used once.
-                installButton.disabled = true;
-
-                // Show the prompt.
-                const { userChoice } = await event.prompt();
-                console.info(`user choice was: ${userChoice}`);
-              });
-            });
-          </pre>
-        </section>
-        <section data-dfn-for="AppBannerPromptOutcome">
-          <h4>
-            <code>AppBannerPromptOutcome</code> enum
-          </h4>
-          <p>
-            The <dfn>AppBannerPromptOutcome</dfn> enum's values represent the
-            outcomes from <a>presenting an install prompt</a>.
-          </p>
-          <dl data-dfn-for="AppBannerPromptOutcome">
-            <dt>
-              <dfn>accepted</dfn>:
-            </dt>
-            <dd>
-              The end-user indicated that they would like the user agent to
-              <a>install</a> the web application.
-            </dd>
-            <dt>
-              <dfn>dismissed</dfn>:
-            </dt>
-            <dd>
-              The end-user dismissed the install prompt.
-            </dd>
-          </dl>
-        </section>
-      </section>
       <section>
         <h3>
-          Extensions to the <code>Window</code> object
+          Extensions to the `Window` object
         </h3>
         <p>
-          The following extensions to the <code><dfn data-cite=
-          "HTML/window-object.html#window">Window</dfn></code> object specify
-          the <a>event handler idl attribute</a> on which events relating to
-          the <a>installation</a> of a web application are <a>fired</a>.
+          The following extensions to the {{Window}} object specify the
+          <a>event handler idl attribute</a> on which events relating to the
+          <a>installation</a> of a web application are <a>fired</a>.
         </p>
         <pre class="idl" data-cite="HTML">
           partial interface Window {
             attribute EventHandler onappinstalled;
-            attribute EventHandler onbeforeinstallprompt;
           };
         </pre>
         <pre class="example js" title=
@@ -807,18 +524,6 @@
             interface</a> [[DOM]]. This event is dispatched as a result of a
             successful installation (see the <a>steps to install the web
             application</a>).
-          </p>
-        </section>
-        <section data-dfn-for="Window">
-          <h4>
-            <code>onbeforeinstallprompt</code> attribute
-          </h4>
-          <p>
-            The <dfn>onbeforeinstallprompt</dfn> is an <a>event handler IDL
-            attribute</a> for the "<dfn>beforeinstallprompt</dfn>" event type.
-            The interface used for these events is the
-            <a>BeforeInstallPromptEvent</a> interface (see the <a>steps to
-            notify that an install prompt is available</a>).
           </p>
         </section>
       </section>
--- a/index.html
+++ b/index.html
@@ -315,9 +315,9 @@
         </li>
         <li>If <a>obtaining the manifest</a> results in an error, the user
         agent MAY either:
-          <ol>
+          <ul>
             <li>Fall back to using the <a>top-level browsing context</a>
-            {{Document}}'s metadata to to populate <var>manifest</var> in a
+            {{Document}}'s metadata to populate <var>manifest</var> in a
             user-agent-specific way (e.g., setting
             |manifest|.{{WebAppManifest/name}} to the document <a data-cite=
             "HTML#the-title-element">`title`</a>) and considering the document
@@ -325,9 +325,9 @@
             </li>
             <li>Or, consider the document not <a>installable</a>.
             </li>
-          </ol>
+          </ul>
         </li>
-        <li>Otherwise, the {{Document}} MAY be considered <a>installable</a>
+        <li>Otherwise, the {{Document}} is considered <a>installable</a>
         (at the user agent's discretion; see [[[#installability-signals]]]).
         </li>
       </ol>
@@ -373,34 +373,6 @@
           suited for the space available (e.g., the <a>short_name</a> member
           might be better suited for the space available underneath an icon).
         </p>
-      </section>
-      <section>
-        <h3>
-          Installation process
-        </h3>
-        <p>
-          The <dfn>steps to install the web application</dfn> are given by the
-          following algorithm:
-        </p>
-        <ol>
-          <li>Let <var>manifest</var> be the manifest value created during <a>
-            steps to determine installability of the document</a>.
-          </li>
-          <li>Perform an unspecified sequence of actions to attempt to register
-          the web application in the user's operating system (e.g., create
-          shortcuts that launch the web application, register the application
-          in the system uninstall menu, etc.). If the installation fails (which
-          can be for any reason, for example, the OS denying permission to the
-          user agent to add an icon to the home screen of the device), abort
-          these steps.
-          </li>
-          <li>
-            <a>Queue a task</a> on the <a>application life-cycle task
-            source</a> to <a>fire an event</a> named <code>appinstalled</code>
-            at the the {{Window}} object of the <a>top-level browsing
-            context</a> for which the installation took place.
-          </li>
-        </ol>
       </section>
       <section>
         <h3 id="installation-sec">
@@ -472,8 +444,8 @@
           Uninstallation
         </h3>
         <p>
-          User agents SHOULD provide a mechanism for the user to remove the
-          installed application.
+          User agents SHOULD provide a mechanism for the user to remove an
+          <a>installed</a> web application.
         </p>
         <p>
           It is RECOMMENDED that at the time of removal, the user agent also
@@ -481,56 +453,6 @@
           and settings associated with the application, such as permissions and
           persistent storage.
         </p>
-      </section>
-    </section>
-    <section class="atrisk">
-      <h2>
-        Installation Events
-      </h2>
-      <p>
-        DOM events <a>fired</a> by this specification use the <dfn>application
-        life-cycle task source</dfn>.
-      </p>
-      <section>
-        <h3>
-          Extensions to the `Window` object
-        </h3>
-        <p>
-          The following extensions to the {{Window}} object specify the
-          <a>event handler idl attribute</a> on which events relating to the
-          <a>installation</a> of a web application are <a>fired</a>.
-        </p>
-        <pre class="idl" data-cite="HTML">
-          partial interface Window {
-            attribute EventHandler onappinstalled;
-          };
-        </pre>
-        <pre class="example js" title=
-        "Two ways of handling the 'appinstalled' event">
-          function handleInstalled(ev) {
-            const date = new Date(ev.timeStamp / 1000);
-            console.log(`Yay! Our app got installed at ${date.toTimeString()}.`);
-          }
-
-          // Using the event handler IDL attribute
-          window.onappinstalled = handleInstalled;
-
-          // Using .addEventListener()
-          window.addEventListener("appinstalled", handleInstalled);
-        </pre>
-        <section data-dfn-for="Window">
-          <h4>
-            <code>onappinstalled</code> attribute
-          </h4>
-          <p>
-            The <dfn>onappinstalled</dfn> is an <a>event handler IDL
-            attribute</a> for the "<dfn>appinstalled</dfn>" event type. The
-            interface used for these events is the <a><code>Event</code>
-            interface</a> [[DOM]]. This event is dispatched as a result of a
-            successful installation (see the <a>steps to install the web
-            application</a>).
-          </p>
-        </section>
       </section>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -327,8 +327,8 @@
             </li>
           </ul>
         </li>
-        <li>Otherwise, the {{Document}} is considered <a>installable</a>
-        (at the user agent's discretion; see [[[#installability-signals]]]).
+        <li>Otherwise, the {{Document}} is considered <a>installable</a> (at
+        the user agent's discretion; see [[[#installability-signals]]]).
         </li>
       </ol>
       <section>
@@ -3225,6 +3225,33 @@
           </pre>
         </section>
       </section>
+    </section>
+    <section class="appendix">
+      <h2>
+        Incubations
+      </h2>
+      <p>
+        Non-standard extensions to this specification are being "incubated" in
+        parallel by the Web Community. If two or more implementers decide to
+        support an incubated feature, then those features will become part of
+        this specification in the future:
+      </p>
+      <dl>
+        <dt>
+          `BeforeInstallPrompt` and `window.onappinstalled` event
+        </dt>
+        <dd>
+          The `BeforeInstallPrompt` event and `window.onappinstalled` event
+          were originally part of this specification. However, they were
+          <a href="https://github.com/w3c/manifest/pull/836">removed from the
+          specification</a> because they didn't not have support from two or
+          more implementers. You can now find them in the <a href=
+          "https://github.com/WICG/beforeinstallprompt">`BeforeInstallPrompt`
+          event and `window.onappinstalled` repository</a> at the <a href=
+          "https://wicg.io"><abbr title=
+          "Web Incubator Community Group">WICG</abbr></a>.
+        </dd>
+      </dl>
     </section>
     <section class="appendix">
       <h2>
--- a/index.html
+++ b/index.html
@@ -3510,10 +3510,6 @@
         <li>[[HTML]] defines the following terms:
           <ul>
             <li>
-              <dfn data-cite="HTML/parsing.html#completely-loaded">completely
-              loaded</dfn>
-            </li>
-            <li>
               <a data-cite=
               "HTML#unordered-set-of-unique-space-separated-tokens"><dfn>unordered
               set of unique space-separated tokens</dfn></a>
--- a/index.html
+++ b/index.html
@@ -3233,10 +3233,11 @@
         Incubations
       </h2>
       <p>
-        Non-standard extensions to this specification are being "incubated" in
-        parallel by the Web Community. If two or more implementers decide to
-        support an incubated feature, then those features will become part of
-        this specification in the future:
+        Extensions to this specification are being incubated in parallel by the
+        Web Community, some of which are shipping in multiple browsers. If two
+        or more browser engines end up supporting an incubated feature, then
+        those features will become part of this specification in the future -
+        allowing them to become a standard the Web Platform:
       </p>
       <dl>
         <dt>
@@ -3246,8 +3247,8 @@
           The `BeforeInstallPrompt` event and `window.onappinstalled` event
           were originally part of this specification. However, they were
           <a href="https://github.com/w3c/manifest/pull/836">removed from the
-          specification</a> because they did not have support from two or
-          more implementers. You can now find them in the <a href=
+          specification</a> because they did not have support from two or more
+          implementers. You can now find them in the <a href=
           "https://github.com/WICG/beforeinstallprompt">`BeforeInstallPrompt`
           event and `window.onappinstalled` repository</a> at the <a href=
           "https://wicg.io"><abbr title=
--- a/index.html
+++ b/index.html
@@ -3506,18 +3506,6 @@
             </li>
           </ul>
         </li>
-        <li>[[DOM]] defines the following terms:
-          <ul>
-            <li>
-              <a data-cite="DOM#event"><dfn><code>Event</code>
-              interface</dfn></a>
-            </li>
-            <li>
-              <a data-cite="DOM#concept-event-fire"><dfn data-lt=
-              "fire|fired|firing">Fire an event</dfn></a>
-            </li>
-          </ul>
-        </li>
         <li>[[HTML]] defines the following terms:
           <ul>
             <li>


dominickng:
It would be good to let @mgiuca weigh in as well as the Edge folks before this lands.

mgiuca:
Sorry, I've had a hell of a week and haven't seen the thread #835 until just now. (I am sick now so I'm not going to try and figure out a rational position at this hour on a Friday night.)

Can we hold off on this until next week? Reading over the thread, there are reasonable arguments on both sides, so ultimately it should come down to whether there are enough browsers implementing the standard for it to remain in.

I'll make a couple of quick points:

- Unlike other APIs, where the existence or absence of a method causes potential compatibility issues between browsers that sites have to deal with, BIP is an event that is specified as firing at the UA's discretion, which could be never. Therefore, sites do not need to explicitly account for the possibility that a UA hasn't implemented BIP (e.g., by detecting the presence of a method). They can just hook an event listener and if it never fires, oh well.
- When we tried to modify BIP into a different API several years ago, we faced resistance from the fact that it wasn't just Chrome implementing. From memory, Samsung Internet and Opera and maybe other browsers had independently implemented this method. Now I suspect (but haven't confirmed) that Edge-on-Chromium will also include this feature. This may or may not be considered sufficient implementor interest, but it should give us pause before removing something from the standard. This isn't just a Chrome-proprietary feature.

marcoscaceres:
>  am sick now so I'm not going to try and figure out a rational position at this hour on a Friday night.

Absolutely, the proposal is for discussion and to reach consensus. No rush! Get well and we pick this up when you are back and feeling better!

othermaciej:
Code to implement this event exists in Chromium, with a significant part, i.e. the event itself, lying in Blink:

https://cs.chromium.org/search/?q=beforeinstallprompt&sq=package:chromium&type=cs

Is there evidence that implementations in non-Chrome but Chromium-based browsers is or would be independent? 

(In my view, independently implemented prompts or install actions that nonetheless use this shared code would not count as independent implementations, since the specification only deals with the event and its dispatch, not with whatever native prompt may be related to it.)

dominickng:
Edge is an interesting case here, and I'd love for Microsoft folks (@aarongustafson) to weigh in here and make sure my understanding is correct.

Edge has a separate downstream repository that is kept up to date with Chromium, but my understanding is that Edge will ship features from the web platform layer upwards that Chromium does not have (where we diverge in opinions or the ability to accept implementations in Chromium at the time Microsoft implements them). Similarly, Edge would disable features that Chromium introduces for similar reasons, or adjust the implementation to suit their views. 

Even if the actual code implementing the feature is shared inside Blink from Chromium, it still seems like Edge is making conscious choices about the web platform it chooses to ship, and not simply replicating what Chromium is doing. It seems a little harsh to say that they aren't independent when they do actually exhibit control and opinion down to the web platform layer.

dominickng:
Also, I know one could argue that pretty much all Chromium derivatives have their own repositories and theoretically could diverge their web platform implementations, but Edge is possibly one of the few that are committed to doing so.

rniwa:
> Even if the actual code implementing the feature is shared inside Blink from Chromium, it still seems like Edge is making conscious choices about the web platform it chooses to ship, and not simply replicating what Chromium is doing. It seems a little harsh to say that they aren't independent when they do actually exhibit control and opinion down to the web platform layer.

The part where you say "actual code implementing the feature is shared inside Blink from Chromium" is the thing that says Edge does not constitute an independent implementation.

Consider this. Almost every port of WebKit ships a different set of features. In fact, Apple's macOS port and iOS port of WebKit ship a different set of features, and some features are even implemented in completely different code (e.g. anything related to touch or pointer events). Yet, I don't think we would consider each port of WebKit as an independent implementation of whatever Web feature.

Similarly, there are GTK+, WPE, WinCairo ports that are maintained by other companies, and each port ships their own browser (e.g. to PlayStation). All of these ports don't even share the same networking or graphics layer, and they also have a completely different backend for things like WebGL and storage. Yet we still wouldn't consider those ports as independent implementations either.

othermaciej:
Even if Microsoft was exercising some sort of editorial control, code from the same upstream repo implementing the same feature would still not be an independent implementation. The point of this general requirement is to demonstrate that the feature can be interoperably implemented from the spec, without having to copy specific code.

But even so, I'm not aware of Edge removing a significant number of (or really any) web-facing features that ship in Chrome.

marcoscaceres:
I agree. AFAIK, Microsoft folks have conceded that they are not an independent implementation for the purpose of standardization. However, @aarongustafson or @travisleithead can confirm. 

Regardless, we wholeheartedly appreciate and value Microsoft’s input - as well as the input of any WebAppsWG member, which includes Samsung, and the developer community at large. However, we do have a (W3C) Process to follow and features that are not going to be interoperable due to lack of implementation must be removed from the spec.

mgiuca:
Hi Marcos,

As I wrote in #835, I'll leave the discussion of why folks are opposed to this API over there, and assume they Mozilla and Apple will not be implementing this.

> Regardless, we wholeheartedly appreciate and value Microsoft’s input - as well as the input of any WebAppsWG member, which includes Samsung, and the developer community at large. However, we do have a (W3C) Process to follow and features that are not going to be interoperable due to lack of implementation must be removed from the spec.

Are you able to link me to the text of that process? Forgive my ignorance, I've never been very clear on what the exact rules are and it would be good to be speaking on the same page.

I'd like to clarify what Microsoft's role is now that they are using Chromium code. It seems unfortunate to me if two of the main browsers agree to implement something that we can't spec it on the grounds that those two browsers happen to share code.

Regardless, if this is taken out of the spec, I think it's unlikely it'll be removed from Chromium. There is too much developer demand for this feature, and we think we've found a good balance between developer demand and being respectful of the user. We would likely take the text of this API and move it into its own spec (maybe in WICG).

rniwa:
See the following links I’ve provided in the other issue:
W3C’s technical document requirement: https://www.w3.org/2019/Process-20190301/#implementation-experience
Web Apps WG’s working mode: https://www.w3.org/2019/05/webapps-charter.html#working-mode

Also please refer to my earlier comment about different ports of WebKit.

marcoscaceres:
See also the removal of features marked “at risk” as part of the CR phases. We are officially not in CR (for administrative reasons, because it would prevent us from iterating) - but for all intents and purposes we’ve considered the API implementable for a while. 

https://www.w3.org/2019/Process-20190301/#candidate-rec

https://www.w3.org/2019/Process-20190301/#revised-cr

rniwa:
> Regardless, if this is taken out of the spec, I think it's unlikely it'll be removed from Chromium. There is too much developer demand for this feature, and we think we've found a good balance between developer demand and being respectful of the user. We would likely take the text of this API and move it into its own spec (maybe in WICG).

FWIW, I don’t think moving to WICG is appropriate. WICG is a place for people to *incubate* ideas, not a place to hoard all the ideas that have failed / rejected (for the lack of better words) elsewhere. 

In the case of a single vendor browser feature like this, producing a technical note or adding an obsolete section in the spec is what we’d usually do.

mgiuca:
Thanks for the links folks.

I'd like to get Microsoft's feedback on this (btw I checked, Edge Canary does support this feature) and continue discussing with Google web people. Since people are starting to go away on holidays (I will be out after this week ends) I would suggest we hold off on the removal until January.

WICG might not be the right place for it. We're essentially talking about maintaining a feature in Chromium long-term that is not a web platform standard (not something we usually try to do) so it's not clear what we'd do with the spec text. But I'd like for it to be hosted somewhere rather than this simply being an "undocumented" feature of Chromium-derived browsers.

marcoscaceres:
@rniwa wrote:
> FWIW, I don’t think moving to WICG is appropriate. WICG is a place for people to incubate ideas, not a place to hoard all the ideas that have failed or have been rejected (for the lack of better words) elsewhere.

Essentially, what @rniwa writes this is correct. It would run into the same issue at the WICG: the WICG generally give incubations around year to mature and make progress, otherwise we [archive them](https://docs.google.com/spreadsheets/d/1K2EtkvKXMDk_h2goR34uMIWmw1LfPhIi-QAJHcQBP_4/edit#gid=1759632076).

> In the case of a single vendor browser feature like this, producing a technical note or adding an obsolete section in the spec is what we’d usually do.

My preference is keep them documented on MDN as Chrome-only feature clearly marked as non-standard. W3C Notes or otherwise on TR confuses people... alternatively, we host it as clearly marked "unofficial", similar to [execCommand](https://w3c.github.io/editing/execCommand.html).

marcoscaceres:
@mgiuca, waiting to next year from all stakeholders is fine. I'd also really like to hear from other folks. I'm also away all of from the 22nd of Dec until ~20th of January.

aarongustafson:
Catching up…

> But even so, I'm not aware of Edge removing a significant number of (or really any) web-facing features that ship in Chrome.

Nor am I.

I am in this week, but many of my colleagues are out through the 6th (or so) of January. I think we (Microsoft) need to have a chat about our overall position on BIP. I personally think it’s (potentially) useful, but I’m not sure it needs to be explicitly part of this spec. That said, I don’t have an objection to including a note in the spec about UAs being free to implement some form of prompting event, perhaps even with a specific callout of the existing (in this event "non-standard") beforeinstallprompt event name. I don’t know that any of us can claim to know how useful this particular event will be going into the future. Perhaps it will eventually claim multiple interoperable implementations and become a standard (like XHR did), but maybe not as part of this spec.

That said, this is purely my personal opinion. Let me see if there are conflicting views from others in Microsoft. I’ll circle the wagons and respond in the new year.

Of course none of this changes the reality that we are not an independent rendering engine anymore… but that’s another conversation for another day (and probably needs to happen in another venue).

mgiuca:
@aarongustafson 
> I personally think it’s (potentially) useful, but I’m not sure it needs to be explicitly part of this spec.

It doesn't *need* to be part of any spec in particular (it could, for instance, be its own spec document, which is where I'll put it if it gets removed from Manifest, though I'm not sure where to host it). But being part of Manifest seems like the most appropriate place for it, if it's going to be standard.

> That said, I don’t have an objection to including a note in the spec about UAs being free to implement some form of prompting event, perhaps even with a specific callout of the existing (in this event "non-standard") beforeinstallprompt event name.

That doesn't sound good. That would mean different UAs could implement different, incompatible events, with different names, or with the same name but different methods, or with the same `prompt()` method but different semantics. That's precisely why we have standards, to establish a contract between developers and the web platform, so they don't have to write code specific to each browser.

If there is only one implementation (in Chromium), then it doesn't need to be (/ isn't allowed to be) in a W3C standard, because there aren't going to be several divergent implementations. But if ever a second independent implementation happens (say Mozilla changes their mind, or Microsoft rips out our code and writes their own implementation of `beforeinstallprompt`), then at that point, we will want an agreed standard. It doesn't make sense for the spec to non-normatively suggest `beforeinstallprompt`. It is either part of the standard (developers code against the standard), or it is not (developers can code against Chromium's implementation).

briantbutton:
"and leave it to browsers to provide better UI to surface installability of web applications"

As a developer, it shocked me to read that coming from Mozilla.  Back to the wild west, eh?  I may have misunderstood the Mozilla mission.

May I ask what the Firefox browser plans to provide?  On Firefox for Android, I have found nothing. 
 Is there a counter-proposal from Mozilla?  Perhaps, rather than junking beforeinstallprompt, identify if there are specific deficiencies that can be repaired to make it "OK".

AFAIK, the "App Install" conveys no new privileges; so I don't see a security problem.  Currently, users must click "yes" twice in succession to get an install on Chrome.  So, the notion of an "accidental install" seems a bit spurious.

The nature of our App is that users will use it intensively and frequently for a few weeks; and when they use it, the "standalone" formatting is critical to a good experience.  So we need the Web App installation.

Currently we deal with two different install mechanisms.  One more won't hurt . . . much.  But it really surprised me to see us making a policy to drive that way.

mgiuca:
Hi Marcos,

I've discussed the issue internally with various colleagues on Chrome.

@slightlyoff suggested as a compromise position that BeforeInstallPrompt be removed from the normative part of the spec, and placed into a "non-normative annex" at the bottom. He cited ECMAScript which has done this (e.g. https://www.ecma-international.org/ecma-262/5.1/#sec-B).

This annex would basically say "this is not part of the standard, and you do not need to implement it to comply with the standard, **but**, if you do implement it, you might want to implement this API rather than designing your own."

That would allow browsers that do implement it to remain in agreement (without having to go off and find a side document). Basically what @aarongustafson suggested earlier but it wouldn't be a vague note, it would be a complete non-normative specification. What do you think of that counter-proposal?

mgiuca:
I have put up a counter-proposal PR at #843. The normative effect of which should be largely the same as this PR, but it preserves the text of BeforeInstallPrompt non-normatively.

Also, I think this PR was over-zealous with its removal; it not only removes BeforeInstallPrompt but all of the normative prose surrounding install prompts, which I think is worth keeping. My PR keeps the non-BIP-related part of install prompting in the normative part of the document.

erikmchut:
From a user privacy perspective, it is always better, (and frequently a legal requirement), to inform the user in clear, understandable terms what they are agreeing to. The beforeinstallprompt is a key component in user privacy for PWAs. beforeinstallprompt allows a website to clearly and concisely explain to the user what installing a specific app means and what the user benefit is for doing so.

If the beforeinstallprompt event is removed, the user is surprised by a prompt with little context due to the generic messaging in the default prompt. Frequently this default prompt comes at the wrong time in the user lifecycle and leads to user confusion. I've read through the above thread and lined issue, and it feels misguided -- the working group added beforeinstallprompt as an essential privacy-preserving feature.

david0178418:
As yet another web developer commenting on this, I greatly oppose this PR and favor @mgiuca's compromise in #843 (Honestly, I'm shocked that this functionality is being put on the chopping block at all).

As mentioned several times, the ability to badge an app as installable has immense value to both the users and developers.  We need to begin to educate users that they don't need to go to the app store to get our apps.  They are looking at it **now** and can install it to their device like any other app.  Burying the option in the browser UI (`especially browsers that conveniently further bury the option below the fold...`) is essentially killing the functionality.  Yes, every site can be installed.  But giving this feature the same level of visibility in the context of a personal blog as you would a shopping list application or game that works entirely offline is just turning it into white noise that that the user will ignore.

Is the automatic prompt associated with the "beforeinstallprompt" event appropriate?  I don't know.  I feel like it shouldn't be automatic.  After reviewing thread, the original issue raised by @rniwa (#835 **"beforeinstallprompt : Prompting user makes it to hard to discern whether the user truly wanted to "install" a web app"**) does not seem to be the heart of the matter.  This could be addressed with a counter proposal that would require user interaction before prompting the native WebAPK install/Add to Homescreen functionality. but as stated here: https://github.com/w3c/manifest/issues/835#issuecomment-565123787

>> The real issue is the automatic "mini info bar" which shouldn't be automatic (or shouldn't exist at all) but let the developer decide when he can inform the user that the application is installable (and allow him to install it with the same experience despite browsers differences) is a must have.
>
> I understand you and some other developers feel that way, but this is precisely what we're objecting. We don't want to give away that control to web apps / sites.

A lot of time could have been saved if the original issue had been stated as "beforeinstallprompt: Sites should not be allowed to prompt installability".

With that appearing to be the heart of the matter, like others, I'm having a very difficult time following the reasoning behind opposing this control.  And judging from the other comments from web developers, I'm not the only one.

marcoscaceres:
With my WebApps WG Chair hat on, the [charter](https://www.w3.org/2019/05/webapps-charter.html#working-mode) we all agreed to operate under states that:

> Additionally, support from two or more web platform implementers is required before a substantive change can be made to a specification. 

The above was explicitly put into the charter to prevent single implementer features being added into WebAppsWG specifications. I understand that we are in a very difficult situation, in that we added this feature with the hope that it would be adopted by multiple user agents. In hindsight, it was a mistake to do that - and this was added under the old charter, in which we allowed "aspirational features" to be added to the spec in the hope that other implementers would follow what was in the spec. But despite years and multiple implementations (Gecko + WebKit), BIP was not adopted. 

Additionally, the W3C Process requires us that things in a specifications have multiple implementations or otherwise will be removed before the specification enters the Proposed Recommendation phase - i.e., this will be "at risk". 

Even without any malice intended, and will only the best intentions, making this as a feature optional circumvents the W3C Process would set a bad precedence for the reasons @rniwa and myself have already outlined.  

Thus, we have the following W3C Process options: 

 1. A Working Group member can Formally Object - in which case, we will can escalate to the Director to weigh the evidence/proposal and make a decision to break the deadlock.
 1. We ask dissenting voices to concur, and respect the spirit of the Charter - and find a different place to put BIP (e.g., move it to the WICG for now).

david0178418:
> But despite years and multiple implementations (Gecko + WebKit), BIP was not adopted.

Looking at the thread, it's still not clear as to _why_ it was not adopted.  The original objection was around the automatic notification.  The reason for the consistent protest from web developers is that no counter proposal is being offered to address the very real need for awareness of whether or not the app has been installed.  As mentioned, this is vital in educating customers that they are looking at our apps that they can install and don't need to go to the app store.

Opposition to this proposal is not specifically being objected to.  Taking away this ability enabled by this specific implementation is.  Before mozilla and apple decide to kill this off, can we first get a counter proposal that addresses the prompting/user confusion concerns around BIP while maintaining our ability to educate our customers with badging?

marcoscaceres:
> Looking at the thread, it's still not clear as to why it was not adopted.

On the Firefox side, we don't think we need BIP. We are happy to do "ambient badging" and feel it's sufficient for our users at this point in time. This may change in the future, hence me saying we should incubate this as we gather more user and developer feedback at scale.  

> Before mozilla and apple decide to kill this off, can we first get a counter proposal that addresses the prompting/user confusion concerns around BIP while maintaining our ability to educate our customers with badging?

The counter proposals are what is shipping in Safari and in Firefox Preview.

rniwa:
> Taking away this ability enabled by this specific implementation is.

Given that neither Gecko nor WebKit ever implemented this feature, we’re not taking anyway any ability. We’re simply restoring the specification to reflect the reality that this has been and will be Blink only feature for now.

frlinw:
It's not a big deal if it's out of the spec as long as there is a mention of "related APIs/incubated APIs" in the header of the spec.
I'll still fight for a better BIP because - for some specific features - engagement should come from the app content, not rely on the will of browser vendors.

marcoscaceres:
Hi All, I'd like to close this issue by this Friday so we can continue moving forward with the rest of the spec. @mgiuca, @dominickng, I'd like to hear back from you about the options I put forward in https://github.com/w3c/manifest/pull/836#issuecomment-597935593

mgiuca:
Hi, sorry, been slow to respond, significant changes to work environment, etc. ☹️

I'm not going to raise a formal objection. I think we established earlier that we're not going to fight the removal of BIP from the spec.

> Even without any malice intended, and will only the best intentions, making this as a feature optional circumvents the W3C Process would set a bad precedence

My proposal #843 was not intended to be marking the feature as "optional". Maybe this is just semantics, but I was explicitly framing it as "non-standard functionality, but here is the text if you wish to implement". In my mind, that's different to "optional". It means there's no agreement at the W3C to implement it. There is no requirement, not even a "MAY" or a "SHOULD" to implement it like this. There is a non-normative explanation that says "psst, if you do want something like this, here's how we've done it in Blink, so perhaps you would like to be compatible with us?". You might look at that and go, "that's just another way of saying optional". 🤷

Essentially, no matter what, a) Chromium will continue to expose this feature, and b) the specification of the feature will continue to exist somewhere. #843 was just a suggestion to keep the specification text in a readily available place, so that folks don't have to go searching for a separate WICG. If you don't want that, I'm not going to keep fighting. Let's take it out, and I will move it to WICG.

**But I do have one request**. Let's merge the patch in #843 with the non-normative annex removed, as opposed to this patch #836. I think #836 removes way too much text (the whole of the "install prompts" section, and all of the normative text around how to show an install prompt, which goes beyond the removal of the BIP interface). I did a careful job in my patch of separating out the BIP interface from the install prompts section. If you are unwilling to accept my non-normative annex, given the above clarification, please let me know, and I will update #843 to remove the non-normative annex, but keep the install prompts section. Alternatively, update this PR to do the same. At the same time, I will create a WICG repo for BeforeInstallPrompt and create a new spec, which contains the text of the non-normative annex from #843.

othermaciej:
It’s not typical for standards to reference external specs for removed features. Do you really want to formally object over lack of such a non-normative note?

(FWIW I think such a note might be ok depending on how it’s worded. “Spec formerly included this feature, now it’s back to incubation.”)

mgiuca:
> It’s not typical for standards to reference external specs for removed features. Do you really want to formally object over lack of such a non-normative note?

No, I don't. (Re-read my previous email: "I'm **not** going to raise a formal objection. I think we established earlier that we're not going to fight the removal of BIP from the spec.")

> (FWIW I think such a note might be ok depending on how it’s worded. “Spec formerly included this feature, now it’s back to incubation.”)

If we have to end up moving it to WICG, then I would appreciate a spec linking out non-normatively to that incubator repo, but as I said I'm not going to raise objections about it. While we're at it, I think it might be useful if we also linked non-normatively to other relevant incubations, such as [web-share-target](https://github.com/WICG/web-share-target/), just to say "heads up, here are some other Manifest features we're thinking about including".

marcoscaceres:
> If we have to end up moving it to WICG, then I would appreciate a spec linking out non-normatively to that incubator repo, but as I said I'm not going to raise objections about it. While we're at it, I think it might be useful if we also linked non-normatively to other relevant incubations, such as web-share-target, just to say "heads up, here are some other Manifest features we're thinking about including".

That seems reasonable. Let's spin up the spin up a BIP repo, add then note, then merge this. I'll draft up some the note text and ping you. I'll also take a look at the reverting some of the installation text as you suggested above (or see if #843 is more suitable).

othermaciej:
@mgiuca I’m sorry, I misread your previous comment! Read the “not” as “now”. Please disregard that portion of my comment. I’ll strike through it if necessary.

mgiuca:
Great thanks. Can you spin up a new BIP repo (WICG/beforeinstallprompt) and if you want, you can leave it empty. I'll paste my text in there.

Before we move stuff there, we should decide on whether we're going to:

a) Jettison all of the "Install prompts" section, or
b) Surgically remove BIP from the "install prompts" section as I have done in #843 (my preference).

This decision will affect how much text has to go into the BIP repo. I'll leave it up to @marcoscaceres to decide between the two.

@othermaciej no problem, "not" and "now" are annoyingly close and mean basically the opposite :)

marcoscaceres:
Coming back to this... so it's actually a little bit more, um, unfortunate and larger in scope than we had originally discussed. If BIP goes, then `.onappinstalled` also needs to go it seems. Safari and Firefox don't support `.onappinstalled` either. This would mean also removing the "Installation process" and all the "Installation Events" section.

mgiuca:
That's surprising... I thought Firefox was implementing `appinstalled`.

Do you have an objection to this event similar to BIP, or just nobody got around to implementing it?

marcoscaceres:
I think I implemented it a while back but I don't think it ever got exposed... but let me make absolutely sure.

marcoscaceres:
yeah, [I implemented it back in 2016](https://searchfox.org/mozilla-central/source/dom/webidl/Window.webidl#275) under the pref "dom.manifest.onappinstalled". However, it never got shipped in any product. I'll check with the Fenix team again if they are interested in it.

marcoscaceres:
I've got confirmation that the Fenix/GeckoView team doesn't intend to fire the event either. I'm going to remove it from Gecko to avoid further confusion.

mgiuca:
OK. I'm happy to tie `appinstalled` to `beforeinstallprompt`, or not. Just don't want to flip-flop around.

@othermaciej @rniwa does Safari implement the `appinstalled` event or do you have plans to?

othermaciej:
We don't currently implement it. I don't think we've considered whether we would eventually.

`appinstalled` doesn't seem obviously harmful, but it's not clear what the use case is. From what I can tell, neither the explainer nor the spec explain why it might be useful to listen to this event. If the use case is to indicate completion of an in-page install button, then it makes sense for it to be tied to `beforeinstallprompt`. If there are other use cases, it would be good to know.

mgiuca:
The use case is somewhat tied to beforeinstallprompt (when the app is installed, you can remove your in-page install button). But it's also useful without beforeinstallprompt, because developers will want to show some UI to tell users how to install the app, and if the user does install the app, they'll want to remove that UI.

dominickng:
Users can always use browser UI ("Add to Home screen" menu item) to install a web app outside of using `beforeinstallprompt`. This is where `appinstalled` becomes important in isolation.

To expand on Matt's point, we often saw developers create custom web content UI that pointed to our 3-dot menu saying "you can install in here - choose the menu item". I've seen this pattern on iOS too where sites create UI that points to the middle of the bottom of the screen (the share button), which is where "Add to Home screen" lives in Mobile Safari.

`beforeinstallprompt` alleviates this pattern - it lets developers prompt the user more respectfully without creating prompting UI of their own that is tied to browser-specific UI that may change and may be different between different browsers.

In browsers without `beforeinstallprompt`, `appinstalled` allows developers to suppress any prompting UI they create. It also allows them to measure install conversions.

marcoscaceres:
ok, removed `onappinstalled` from Gecko. 

The issue that @dominickng points to remains - so we will have to see how that plays out longterm. It may go away on its own, similarly to how websites used to ask every user to "bookmark this site!" back in the 90's and 2000s... as users become more savvy with using browsers, nagging users to bookmark stuff went away. Same might happen here: users become more savvy about adding to home screen the webapps they like without needing any prompting🤞.

briantbutton:
As a data point, here is where we stand:

Our App is associated with stays at resort hotels; it will be used frequently during the stay and never afterwards.  The hotel will guide them to install the App when they arrive as it provides major benefits for understanding attractions, getting service and sharing their experiences.

On Android, we guide users on Chrome to install the PWA.  We can also detect if the PWA was already installed and simply redirect them to it if they happen to open in the browser.  This is convenient, looks good and is pretty foolproof.

Users opening our website with Firefox in Android will be directed to close it and use Chrome.

Users with iPhone are be directed to use Safari only.  We give them little instruction with pictures of the buttons they should seek and press.  Different iPhones put these buttons in different places and sometimes THEY LOOK DIFFERENT but we do the best we can do.  65% of iPhone users succeed in following the breadcrumbs.

This is pretty adequate, especially considering the respective market shares of mobile OS's outside of the U.S.  It hardly seems like a shining example of standards bodies bringing uniformity.

mgiuca:
@marcoscaceres have you given thought to the (a) and (b) options in https://github.com/w3c/manifest/pull/836#issuecomment-600938135?

marcoscaceres:
> @marcoscaceres have you given thought to the (a) and (b) options in #836 (comment)?

Yeah, sorry. I was going with "Jettison all of the "Install prompts" section" (current PR). 

I'm about the add the section about incubation. Treat whatever I send up as "draft" and we can iterate on it.

marcoscaceres:
Ok, this is ready for another round of review.

aarongustafson:
This works for me. I like the shift to positive language (and perhaps hope?) in the section that covers this stuff.

index.html
```diff
@@ -315,20 +315,20 @@ <h2>
...
             <a>installable</a>.
             </li>
             <li>Or, consider the document not <a>installable</a>.
             </li>
-          </ol>
+          </ul>
         </li>
-        <li>Otherwise, the {{Document}} MAY be considered <a>installable</a>
-        (at the user agent's discretion; see [[[#installability-signals]]]).
+        <li>Otherwise, the {{Document}} is considered <a>installable</a> (at
```
mgiuca, line 330: I don't know why "MAY" changed to "is". It's still a may, because you still allow user agent discretion.

index.html
```diff
@@ -374,149 +374,20 @@ <h3>
...
-          RECOMMENDED that the user agent allow the end-user to inspect the
-          icon, name, <a>start URL</a>, origin, etc. pertaining to a web
-          application. This is to give an end-user an opportunity to make a
-          conscious decision to approve, and possibly modify, the information
-          pertaining to the web application before installing it. This also
-          gives the end-user an opportunity to discern if the web application
-          is spoofing another web application, by, for example, using an
-          unexpected icon or name.
+          It is RECOMMENDED that UI that affords the end user the ability to
+          <a>install</a> a web application also allow inspecting the icon,
```
mgiuca, line 383: s/allow/allows (tbh this is goes past my concrete understanding of English grammar, but "allows" as opposed to "allow" feels correct in this sentence).

othermaciej:
@dominickng maybe this is moot, since it seems like the right approach has been decided. However, the "appinstalled" event seems insufficient for suppressing an invitation to install, because it doesn't provide a way to know if the user has installed in the past, just if they did it right now. Seems like it would work for conversion, but not clear if that's a strong enough use case by itself.

index.html
marcoscaceres: ah, good point... I'll flip the sentence around so that the MAY is applied to the user agent, not to the document.

dominickng:
@othermaciej: `appinstalled` covers the suppression of installation UI in the current context. Suppressing such UI entirely because the app is already installed is not what `appinstalled` would be used for - that is the realm of [getInstalledRelatedApps](https://wicg.github.io/get-installed-related-apps/spec/).

Pauldb8:
So that is it ?
Mozilla, the "champion" of the web, simply removed one of the most important features to gain traction for PWAs, and yet make the case for open web ? And all this without due diligence, with high developers support, high user demande for ui friendliness.
Nothing stated against make sense or adds up.
It feels like relegating the install button to share button behind the second row of icons is what Apple can do best, yet I understand, their business models *is* apps, but coming from Mozilla ?
I want this feature back, now.  Implement it, nobody asked you what you thought, the web will prevail, and we will not be let down because of fake excuses. Apple will lose market share because of this, good. User interaction will be improved, and Mozilla should better start backing up the web again. Come one why do we have to say this ?

dmurph:
I have a pull request to add it to the manifest-incubations spec. Might be able to get that in next week depending on my workload
https://github.com/WICG/manifest-incubations/pull/9

bathos:
@dmurph I recently needed to consult the original spec text and it was pretty challenging to find the most recent version where it had still been specified, so it's greatly appreciated that you're giving it a proper home again.

briantbutton:
@Pauldb8, good to hear your comments.  The conclusion reached above seems unsupportive of an open Web with consistent features supported cross-browsers and independent of OS.  But we have a job to do . . . 

We are publishing a PWA that, with an in-house server that allows folks to do high performance sharing in non-urban environments.  (http://www.button.network/usage.mp4; forward to 3:20).  Initially marketed toward resort/retreats, we are also getting interest from the scads of schools worldwide that have sub-optimal Internet.

This decision simply made us decide to not support Firefox.  It's not like we had another choice.  That's all.  We direct folks to one of two browsers, depending on whether they are using IOS or not.

bathos:
We also publish an installable PWA. In Firefox it’s not installable, of course, and we document in our FAQ that the application isn’t installable from all browsers, but to date I’ve thought of this as falling somewhere in the progressive enhancement realm sorta. The application remains usable in Firefox, just as it remains usable in Chrome, etc, without installation. This seems pretty important given the user can choose not to install even in a browser where installation is supported, and likely wouldn’t want to install until first using it enough to decide that doing so would be worthwhile. (To me this seems like the rare case where “the gap” is reversed — PWAs have a fantastic capability in this regard that _native apps lack._)

So your comment made me curious @briantbutton: what does “not support Firefox” mean for your app in practice? Do you mean: 

- “we don’t sweat it that FF users can’t install”
- “because FF doesn’t have this feature, we officially don’t consider FF supported, but the app might work there anyway”
- “we prompt users to use a different browser (but they can dismiss the prompt)”
- “we outright block FF usage”
- something else

Again, just curious — I haven’t heard of doing anything but the first before. Hopefully OP will forgive a bit of topic drift — it’s good to learn more about how people are really responding to the state of things I think.

FF usage seems to be getting so low now that all of this might be theoretical in another year or two anyway, for better or worse ... :(

Pauldb8:
Incredibly, they are not even providing an alternative, that would prove that they at least cared. Does not seem like, so please: implement it (major support), propose an alternative, or just die already Firerox. If Firefox does not support open web, I'm uninstalling it. Tell me what should I do ? I would go to Ungoogled Chromium and I would also start recommending it to my users  as well, hope we don't get there.
Same could go for Safari Explorer (the new IE...)

briantbutton:
@bathos, "the middle one".  We just notify them this won't do what they need and suggest a working alternative depending on their OS, but take no other steps.  It happens as they open the App, as I recall.  Even though we use Firefox in some of our S/W test, we don't want users to get started with it if they cannot get the advantage of an "installable WebApp".

I have no rancor on this any more.  We have one browser solution for each OS (two for Windows).  That is enough.  Barely.

My personal opinion, as summarized in my post of January 19 2020 (above), still stands.  Won't repeat it here.

marcoscaceres:
@Pauldb8, I understand you are frustrated, and we welcome your participation - but I kindly remind you of our [code of conduct](https://www.w3.org/Consortium/cepc/) ... inflammatory statements like "Firefox does not support open web" and  "Safari Explorer (the new IE...)" are inappropriate. 

Regarding "beforeinstallprompt", as shown by various browsers, the ability to install a web application is not gated on an install prompt. A user can simply "add to homescreen" any web page they choose - so even if all browsers supported the BIP event, one might still need to handle a situation where BIP is cancelled, and then the user manually chooses to install a web application without BIP (ie.., progressive enhancement to Safari)... Similarly, a user may choose not to install the application at all (progressive enhancement to Firefox).

briantbutton:
@marcoscaceres, if what you are saying is that browsers will handle PWAs as they see fit, I agree.  I don't consider that a good outcome.

For example, on IOS there is an incantation (several incantations, depending on your precise arrangement) but **only** if your browser is Safari.  Each other OS has a very specific, and different, set of options.  We must maneuver through these, guiding our users as best we can.

"Installation" is more than just a matter of putting a button on your desktop.  Both Chrome and Safari respect the Manifest differently on an "installed Web App" versus a regular web page.  The uninstalled version tends to overflow the screen because of a bunch of browser control buttons and orientation (critical to our App) is uncontrolled.  On a phone, with its tiny real estate, the difference in user experience is huge.

marcoscaceres:
> The uninstalled version tends to overflow the screen because of a bunch of browser control buttons and orientation (critical to our App) is uncontrolled. On a phone, with its tiny real estate, the difference in user experience is huge.

Understood, but that seems unrelated to "installation". If I'm reading what you are saying correctly, this appears to the an issues with detecting the display mode or if a set of UI controls are available (which *coincidently* are affected by the application being installed and the manifest being applied.)

Put differently, "applying a manifest" (having the manifest affect the behavior or presentation) is not inherently tied to installation or "add to home screen". This is also why I'm confused about how all this relates to "beforeinstallprompt", which is yet something totally unrelated. 

I think solving for the problem you describe:

>  "overflow the screen because of a bunch of browser control buttons and orientation (critical to our App) is uncontrolled"

Is certainly something we should address - but we should tackle this as unrelated to this whole "installation" thing.

briantbutton:
@marcoscaceres, that is a very useful idea.  Thank you.

I don't know what is connected, separable (or "coincidental") from the standards POV.  The star of the show for our users is the improved improves formatting and orientation-control that comes when the Manifest is applied.  It may sound trivial but there are scores of artifacts that make the whole thing a badly-behaved mess without it.  (Presumably that is why somebody invented Manifests.)

Aside from that, what is "installation"?  The appearance of an icon on the desktop?  (Certainly no new security privileges are received.)  This is handy for users, and gives a bit of panache, but it is not as critical as the user experience impact.

Jaifroid:
I realize W3C has rules on the number of browsers required for a feature to be standardized, and that's fine, but from what I can tell we have: Chrome, Edge, Opera, the mobile versions of these and Samsung Internet all implementing these install events. It seems to leave a minority of Firefox and Safari not having implemented.

Safari, I understand (even if I don't like it) because PWAs undermine Apple's investment in its App Store. I have to say I find the Firefox stance on these install events and on PWAs more broadly to be self-defeating. It could so easily have been at the forefront of these developments which promote an open, non-commercial model for Web applications. I don't really know what Mozilla stands for any more... _Tant pis_.

frlinw:
@Jaifroid All chromium-based browsers = 1. 

The recent activity here led me to think about [this](https://github.com/w3c/manifest/pull/836#issue-352744419)
> leave it to browsers to provide better UI to surface installability of web applications
 The question I ask: 
- Should a standard rely on something unreliable (browser's UI) for feature discovery?
- Should a standard conform to vendors interets or commons interests?
- Is it possible to conform to commons interests if vendors write the standard?

Jaifroid:
@frlinw This raises the broader issue of how W3C can decide sensibly on specifications in a market overwhelmingly populated by Chromium-based browsers (which is not a return to the old domination by Internet Explorer, because the Chromium-based browsers are all forking off their own implementations and doing something slighlty different). Something is awry when I look around and see a diversity of independent browser implementations but they are all counted as a single browser. How much does a codebase have to be forked for it to become a separate browser? I'm sure W3C has thought carefully about these things, but current policy does seem to be leading to counterintuitive decisions.

(FWIW, Personally I don't think the mobile versions should be counted as the same version. They often implement things differently, only complete part of a specification, or else add specific features that are different from the parent browser.)

This is quite related to the quesitons you ask, I think. There is a balance to be struck between vendor implementations and an overseeing body that can set aspirations for a common standard (or a standard for the good of all / the commons).

marcoscaceres:
> Should a standard rely on something unreliable (browser's UI) for feature discovery?

I don't know what this means, or who judges "unreliable"? However, the *aim* is for discoverability to be as good as implementers make it, and it's something browser vendors compete on.  

> Should a standard conform to vendors interests or commons interests?

We try to strike a balance. However, the balance is always tipped towards implementers - because if we put stuff into the spec that only one implementer supports, then the W3C Process forces us to remove the feature. Case in point: "beforeinstallprompt". 

> Is it possible to conform to commons interests if vendors write the standard?

That's another good question. We have a few checks and balances in place, including the W3C's need for "wide review". The working group Chair is supposed to operate from a vendor neutral perspective (I'm the Chair, I work for the W3C... though I used to work for Mozilla). I also edit the spec, and don't represent the vendor. Similarly, the other editors are a mix of folks, but we all contribute to different engines.

Further, we have a (newish) PULL_REQUEST_TEMPLATE that forces us to check that there is consensus (and no objections!) amongst all implementers before we put stuff into the standard. 

> How much does a codebase have to be forked for it to become a separate browser? I'm sure W3C has thought carefully about these things, but current policy does seem to be leading to counterintuitive decisions.

We don't consider "chromiums" as independent implementations: we consider WebKit, Chromium, and Gecko as separate implementations.... I should really update PULL_REQUEST_TEMPLATE to make this more clear.

Jaifroid:
> We don't consider "chromiums" as independent implementations: we consider WebKit, Chromium, and Gecko as separate implementations.... I should really update PULL_REQUEST_TEMPLATE to make this more clear.

I understand that, but the market has become distorted with so many different Blink implementations that it seems some other way of working out at what point a previously approved standard gets dropped is needed. Is it really enough that one engine-developer decides it doesn't have the resources to complete an implementation, or that it's not a priority, and pulls it (especially when we know the other competitor will never implement the spec due to commercial interest)? I understand that this is difficult and that the browser landscape is shifting, but it seems to me that dropping a standard should have a higher bar than that, or else you indeed risk one vendor dominating the landscape with proprietary but widely used features. The genuine consternation caused by the decision behind this PR ought to be food for thought IMHO.

marcoscaceres:
> I understand that, but the market has become distorted with so many different Blink implementations that it seems some other way of working out at what point a previously approved standard gets dropped is needed. 

Yes. Those discussions are happening at every level of the W3C: from the Advisory Board, to all over the community. Unfortunately, I can't find a link to the repo where that discussion is happening... I'll drop a link if I find it!  

> Is it really enough that one engine-developer decides it doesn't have the resources to complete an implementation, or that it's not a priority, and pulls it (especially when we know the other competitor will never implement the spec due to commercial interest)? 

Today, yes. It gives us a base set of interoperability. 

> I understand that this is difficult and that the browser landscape is shifting, but it seems to me that dropping a standard should have a higher bar than that, or else you indeed risk one vendor dominating the landscape with proprietary but widely used features. 

We can't really stop any vendor from shipping proprietary things. They all do it. Not just one vendor, but every browser vendor are equally as guilty... just some do it more than others, and not for nefarious reasons. Often time, the proprietary things get picked up by other browser vendors: for example, the proprietary Web Share API and the new Contacts API from Google got implemented by Apple, so they got moved or are getting moved to the W3C Standards track. But similarly, `<meta vieport>` was an Apple thing, and so was `theme-color`, if I recall correctly... ApplePayJS paved the way for the W3C Payment Request API, but ApplePayJS still ships in Safari. 

The point is that sometimes some ideas won't stick. That's ok... Google is not about to yank "beforeinstallprompt", but neither should we expect other browsers to support how Chrome does installation if they don't want, or need, to. 

> The genuine consternation caused by the decision behind this PR ought to be food for thought IMHO.

Believe me. This, or any other decision as to what goes into this this spec, wasn't taken lightly. I get a little nauseous thinking about the number of emails, bugs, and meetings (physical and virtual), we've had about this feature across various organizations over the last 5 years or so.

msimic:
I've read this whole thread. It is a hell lot of reading.
If there is one thing I've understood, it's that the way something is paved for a standard is wrong.

Taking and/or keeping in a standard a feature that is supported by a wide range of browsers is wrong.
Wide adoption should have absolutely no impact on what goes into a standard.

A widely implemented feature does not mean that the feature is good for feature the web, it just means it is good for the company developing the browser.

A feature that could be good for the future of the web could also nefariously be not implemented on purpose by one browser, since there are currently three browser engines in total (WebKit, Chromium, Gecko) - and could be blocking features for the standard. Democracy does not work when you have three voters.

The W3C need to have their own stance about a feature. Is the feature good for the web or not? If yes have it in the standard, if not remove it.
anybody who has affiliation with one of the companies developing the engines should not have a word in this.

mgiuca:
@msimic Mostly agree (disclosure: I work for Google, one of the big browser developers you say should have less power :))

> Democracy does not work when you have three voters

I agree. It's hard to get useful things done when it only takes 2 "votes" to oppose. An important point though about the intention of this system: this is not technically a democracy. Features don't get accepted or rejected based on the majority vote. Rather, features get accepted if there are at least 2 implementations. (e.g. if there were 100 browser vendors, we wouldn't need 51 implementations to get it into the standard, we would just need 2).

The reason for this rule is sound: it keeps features from becoming standards that are too locked to one specific vendor, or too tied to the details of one engine. Once you have two implementations, you can figure out what are the core concepts that need to be standardized and what can be implementation-specific. With just one, the standard would just be writing down what one browser does. That's also why all the Chromium-based browsers count as one "vote" -- because they only prove a single implementation.

(Whether this rule is right or wrong, I am not offering an opinion, just explaining why it is this way.)

The problem is that that idea was made when there were 4 or 5 engines actively being developed. Now there are just 3, the "rule of two" has turned into a very weak democracy, as you say, with three voters, and any two voters can veto a standard.

On the other hand, the fact that this feature isn't written down in the W3C standard doesn't mean it is going away (as Marcos said). It just means we [wrote it down somewhere else](https://wicg.github.io/manifest-incubations/index.html#installation-events), and it works in Chromium-based browsers. Even if it were in the standard, that wouldn't force Apple and Mozilla to implement it. So from the user/developer standpoint, the result is the same. That brings me to the next misconception:

> Somebody from the W3C can look at some browser, see something they like. If the W3C agrees, mandate that all other browsers implement it by making it a standard.

The W3C does not mandate that browsers implement anything. They simply recommend (that's why W3C standards are called "Recommendations"). When someone writes a web browser (anyone technically can, that's why all the documents are public), they do not have to sign a contract that says you implement everything in the spec. They just voluntarily implement the W3C recs for the sake of compatibility with other browsers. That means if you're a browser vendor, it's generally a good idea to implement everything you can (otherwise you risk falling behind), but if feel strongly about anything, you don't have to implement it.

By necessity, browser vendors have the power to decide what goes into the browser, just as anybody who builds anything has the power to decide what features they have. Consumers have the power to choose which product they use based on what features it has.

That's my standpoint (as the guy who wrote a lot of this standard and argued for its inclusion): I disagree with the decision to remove it, but I understand the reason for this process and ultimately which document it's written in won't affect which browsers support it, which means it doesn't change things for users or developers.

msimic:
> The W3C does not mandate that browsers implement anything. They simply recommend (that's why W3C standards are called "Recommendations"). When someone writes a web browser (anyone technically can, that's why all the documents are public), they do not have to sign a contract that says you implement everything in the spec. They just voluntarily implement the W3C recs for the sake of compatibility with other browsers. That means if you're a browser vendor, it's generally a good idea to implement everything you can (otherwise you risk falling behind), but if feel strongly about anything, you don't have to implement it.

Perhaps the word "mandate" I used is not the right one. But if you don't follow the standard you are asking for trouble, so it becomes indirectly a "mandation".
I for example can make a card that slots into a computer PCI express bus even though I'm not adhering to the standard.
It could work or it could not. But when it does not work it will be my problem. The person that will experience the problem will complain to me. So I'm better of adhering to it.

fabricedesre:
> That's my standpoint (as the guy who wrote a lot of this standard and argued for its inclusion): I disagree with the decision to remove it, but I understand the reason for this process and ultimately which document it's written in won't affect which browsers support it, which means it doesn't change things for users or developers.

So you're saying all these discussions and standards are kind of useless, because in the end browsers do what they want? :)

msimic:
> > That's my standpoint (as the guy who wrote a lot of this standard and argued for its inclusion): I disagree with the decision to remove it, but I understand the reason for this process and ultimately which document it's written in won't affect which browsers support it, which means it doesn't change things for users or developers.
> 
> So you're saying all these discussions and standards are kind of useless, because in the end browsers do what they want? :)

Let them do what they want. That's not the issue here. It is only about how standards are being created.

mgiuca:
@msimic I don't know much about hardware standards. Some standards you need to license and a condition of the license is that you follow them precisely -- W3C standards don't work that way since they are public standards.

If the PCI standard needs to be licensed, then it's a different situation since card manufacturers need to comply in order to keep their license. If not, then it's basically the same here: you _could_ make a non-compliant card but then it wouldn't work as expected and customers would complain.

That's basically how web standards work: you could make a browser that doesn't support `<b>` but then pages would look funny in your browser and users would complain and switch browsers. The more fringe APIs, like this one, the less important that becomes, and at some point, it's fine for browsers to pick and choose which APIs they support, especially if they don't impact the core experience of the site.

Then it just becomes about whether it's worth having in the standards document. Arguably, if only one engine supports an API, there's no point having it in the document, because there's nothing for it to be compatible with. But if another vendor does want to support it in the future, well we've written it down, they can implement that, and then we would make a case that it can go into the standard now.

> So you're saying all these discussions and standards are kind of useless, because in the end browsers do what they want? :)

In some sense, yes. If two or more engines are trying to implement something, then the standards discussions are highly relevant, because we need to agree on how it works. If only one engine is implementing, it doesn't really matter where we write it down, it's a tree falling in the forest :)

msimic:
> @msimic I don't know much about hardware standards. Some standards you need to license and a condition of the license is that you follow them precisely -- W3C standards don't work that way since they are public standards.

You don't have to license any standard, hardware of software.
You have to do it if you want to have a certification/badge to show off.
If you want to manufacture a product that slots into something that is standardized, and you are not breaking any laws while doing it you are free to do it.

For example if BIP was part of the standard then Firefox and Safari would not get a W3C certification, but would still be able to follow parts of the standard and develop without any issue, except for those that their users would file because they are not following it 100%.

marcoscaceres:
Can I please respectfully ask we keep this on topic (BIP)? 🙏 I'm worried we will spin off into the nuances of standardization and add more walls of text to what's already a very lengthy topic. 

As @mgiuca eluded to, the fact that it's not in the main spec doesn't mean that it's not specified. As @mgiuca said, that it doesn't have an independent implementation means (**today**) that it's not in the spec. If **tomorrow** some other engine decides "oh! we need that!", then we would just revert this change. 

@msimic, please understand that a lot of thought went into both the creation (and ultimate removal) of this feature. I was also involved in helping specify it and still believe it has merit. However, the rules of both the W3C Process and those working group membership agrees by signing up based on our [Charter](https://www.w3.org/2022/04/webapps-wg-charter.html) justify its removal. To quote:

> In order to advance to [Candidate Recommendation](https://www.w3.org/Consortium/Process/#RecsCR) and to add features after reaching Candidate Recommendation, each feature is expected to be supported by at least two implementations, which may be judged by factors including existing implementations, expressions of interest, and lack of opposition.

Despite the Editors' best efforts (which includes me), we were not able to overcome the strong opposition and convince implemented this feature is critical. And as @mgiuca said, even if we had it in the spec, there are well articulated reasons why BIP is not a good fit (and doesn't make any sense) for certain user agents. 

BIP is inherently tied to a particular install flow (showing a banner or modal prompt), and not all browsers have that UI flow.

msimic:
>  BIP is inherently tied to a particular install flow (showing a banner or
modal prompt), and not all browsers have that UI flow.

This is not true, every browser has HTML which does this.

And there are repercussions to excluding it because it prevents web sites
that host their own PWAs inside their own dialogs from installing their
PWAs, since browsers use the manifest if present from the url where the
site is hosted itself and not from sub-url. That makes it so you can only
install the website itself and not a useful standalone module of it.

PR #858, opened 2020-03-31
Obtain manifests only using the JSON MIME type

### LINKED ISSUE #821: Why does obtaining not check a MIME type?
We should really enforce MIME types for new formats.

marcoscaceres:
Closes #821

This change (choose one):

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1626158)
* [ ] Edge (public signal)

Commit message:

When obtaining a manifest, this now checks the mime type is a [JSON mime type](https://mimesniff.spec.whatwg.org/#json-mime-type) (anything that has a `text/json` or `application/json` or +json subtype in "essence"). 

We also now export the mime type from our spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/858.html" title="Last updated on May 25, 2020, 12:08 AM UTC (df0d9d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/858/f704809...df0d9d4.html" title="Last updated on May 25, 2020, 12:08 AM UTC (df0d9d4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -134,7 +134,7 @@
       }
     </style>
   </head>
-  <body data-cite="ENCODING">
+  <body data-cite="ENCODING MIMESNIFF">
     <section id='abstract'>
       <p>
         This specification defines a JSON-based manifest file that provides
@@ -265,11 +265,12 @@
         <div class="note" title="manifest.webmanifest or manifest.json?">
           The official file extension for the manifest is `.webmanifest`. Some
           web servers recognize this extension and transfer the file using the
-          standardized <a>MIME type for a manifest</a>
-          (`application/manifest+json`). Developers can also choose a different
+          standardized [=web app manifest MIME type=]:
+          [=application/manifest+json=]. Developers can also choose a different
           extension (e.g. `.json`) or none at all (e.g. `/api/GetManifest`),
           but are strongly encouraged to transfer the manifest using the
-          `application/manifest+json` <a data-cite="mimesniff">MIME type</a>.
+          [=application/manifest+json=] [=MIME type=] or, at least, a [=JSON
+          MIME type=].
         </div>
       </section>
     </section>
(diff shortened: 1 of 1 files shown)


marcoscaceres:
I've started working on a patch for Firefox. 

@mgiuca or @dominickng, will you be ok to do this in Chrome? Or will you want to check for potential breakage? 

@rniwa, same question re: WebKit?

marcoscaceres:
Thanks @kenchris for the speedy review!

index.html
```diff
@@ -1375,6 +1376,10 @@ <h3>
           <li>Await the result of performing a <a>fetch</a> with
           <var>request</var>, letting <var>response</var> be the result.
           </li>
+          <li>If [=header list/extract a MIME type=] from |response|'s
+          [=response/header list=] is failure or not a [=JSON MIME Type=], then
+          abort these steps.
```
dominickng, line 1381: This seems out of step with the earlier text, which "strongly encourages" ... "at least a" JSON MIME type. If we're going to abort the fetching of a manifest for a non-JSON MIME-type, should the earlier text be more explicit?

mgiuca: Also, do we know if this is going to break sites? I don't think implementations currently check that the MIME type is JSON. There are probably a lot of sites that don't serve their manifest with the appropriate MIME type which will break if we add this behaviour.

dominickng: I can confirm Chromium doesn't check the MIME type of the response - we just decode it as UTF8 and try parsing it and see if that succeeds or not.

marcoscaceres: Agree, and I'm also worried about potential breakage, although I don't know how bad it will be. 

We might need to reach out to someone to run a query for us on HTTP Archive or big table? Or see if there is some other way of getting this data. (e.g., via the https://github.com/webdevdata tools)

marcoscaceres:
Closing this, as we are moving all this to the HTML spec.... https://github.com/whatwg/html/pull/5581

PR #863, opened 2020-05-04
Add protocol handlers to the explainer and spec

### LINKED ISSUE #846: Add ability for PWAs to be registered as scheme/protocol handlers
Native applications often register themselves as protocol handlers to increase discoverability and usage. It would be great to have PWAs do the same.

While websites do currently have this ability via `registerProtocolHandler()` (which for the most part just registers the browser as the handler to the OS), it would be interesting to have PWAs be first-class citizens and be launched when a custom-scheme link is invoked. For instance, on Windows 10, one can choose default apps by protocol, and one would be able to point to an installed PWA to handle that protocol:

image

#### Explainer:
https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/master/URLProtocolHandler/explainer.md

#### Related issue:  
It looks like this proposal shares some similarities with #626 (Add a means to associate a PWA with a file extension), and we should probably discuss if they should be unified or kept separate.

fabiorocha:
Closes #846

This change (choose one):

* [ ] Safari ([link to issue](https://bugs.webkit.org/show_bug.cgi?id=211401))
* [x] Chrome ([link to issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1019239))
* [ ] Firefox ([link to issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1635272))
* [x] Edge ([public signal](https://github.com/MicrosoftEdge/MSEdgeExplainers/tree/master/URLProtocolHandler))

Commit message:
Add protocol handlers to the explainer and spec.

***
<a href="https://pr-preview.s3.amazonaws.com/fabiorocha/manifest/pull/863.html" title="Last updated on Feb 26, 2021, 12:27 PM UTC (7ee82ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/863/8c7fd7b...fabiorocha:7ee82ef.html" title="Last updated on Feb 26, 2021, 12:27 PM UTC (7ee82ef)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1171,6 +1171,11 @@
           running <a>processing the <code>shortcuts</code> member</a> given
           <var>manifest</var>["<a>shortcuts</a>"] and <var>manifest URL</var>.
           </li>
+          <li>Set <var>manifest</var>["<a>protocol_handlers</a>"] to the result
+          of running <a>processing the <code>protocol_handlers</code>
+          member</a> given <var>manifest</var>["<a>protocol_handlers</a>"] and
+          <var>manifest URL</var>.
+          </li>
           <li>
             <a>Extension point</a>: process any proprietary and/or other
             supported members at this point in the algorithm.
@@ -1247,6 +1252,7 @@
              sequence&lt;ExternalApplicationResource&gt; related_applications;
              boolean prefer_related_applications = "false";
              sequence&lt;ShortcutItem&gt; shortcuts;
+             sequence&lt;ProtocolHandlerItem&gt; protocol_handlers;
           };
       </pre>
       <p>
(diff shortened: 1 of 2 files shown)


fabiorocha:
@mgiuca / @marcoscaceres / @aarongustafson this is a first stab at having `protocol_handlers` added to the spec, as we discussed. Hopefully this can get some discussion around issue #846 going and we can settle on something soon, although I think most people we've talked to feel comfortable with the current proposal. Thanks for all your input so far.

reillyeon:
@dmurph who registered concerns with the overlap with `registerProtocolHandler()` on MicrosoftEdge/MSEdgeExplainers#286.

dmurph:
open issues: https://github.com/MicrosoftEdge/MSEdgeExplainers/labels/URLProtocolHandler

fabiorocha:
Thanks Daniel. There are at least a couple that can be moved forward with a PR against the explainer. I'll work on updating those.

aarongustafson:
Converted this to a draft while @diekus works on the updates.

marcoscaceres:
Just a heads up that the spec no longer uses WebIDL and I've changed slightly how processing of members happen. I also got rid of "issue a developer warning", and left that up to UAs.

index.html
```diff
@@ -2708,6 +2819,90 @@ <h3>
...
+        </p>
+      </section>
+      <section data-dfn-for="ProtocolHandlerItem" data-link-for=
+      "ProtocolHandlerItem">
+        <h3>
+          <code>url</code> member
+        </h3>
+        <p>
+          The <dfn>url</dfn> member of a <a>ProtocolHandlerItem</a> is the
+          <a>URL</a> <a data-lt="within scope of a manifest">within the
```
marcoscaceres, line 2872: Please use "manifest scope" here, otherwise it might get confusing as "application scope" is not a defined thing.

index.html
```diff
@@ -2708,6 +2819,90 @@ <h3>
...
+          application's scope</a> that opens when the associated protocol is
+          activated.
+        </p>
+        <p>
+          The <a>url</a> member of a <a>ProtocolHandlerItem</a> is equivalent
```
marcoscaceres, line 2877: as above... this is probably not ideal (as per the TAG comments about layering). We own our own definitions of these things, and then we make them work with the underlying model of HTML.

index.html
```diff
@@ -2708,6 +2819,90 @@ <h3>
...
+        <h3>
+          <dfn>Handling a protocol launch</dfn>
+        </h3>
+        <p>
+          When a <a>ProtocolHandlerItem</a> <var>protocol_handler</var> having
+          <a>WebAppManifest</a> <var>manifest</var> is invoked, run the
+          following steps:
+        </p>
+        <ol>
+          <li>Let <var>url</var> be <var>protocol_handler.url</var>.
```
marcoscaceres, line 2892: The algorithm should operate on the document's "processed manifest" instead.

index.html
```diff
@@ -2708,6 +2819,90 @@ <h3>
...
+          </li>
+          <li>Replace the first occurrence of the exact literal string "%s" in
```
marcoscaceres, line 2894: Ok, this bit probably needs really precise processing as a simple replace here might end up with security issues. If anything, we want to do whatever HTML does ... or maybe whatever Web Share Target does. 

cc @mgiuca

marcoscaceres (changes requested):
some issues, but a good base to build on! Thanks for putting this together @fabiorocha!

index.html
diekus: updated to "manifest scope".

christianliebel:
It looks like protocol handlers are already defined at the [Manifest Incubations](https://wicg.github.io/manifest-incubations/#protocol_handlers-member) spec. So, I would go ahead and close this four-year-old draft PR.

PR #879, opened 2020-05-27
Clarify off-scope theme color

### LINKED ISSUE #755: What theme color should be used when a PWA window is out of scope?
https://www.w3.org/TR/appmanifest/#theme_color-member says

> However, a document may override the default theme color through the inclusion of a valid [HTML] meta element whose name attribute is "theme-color".

I see this as a way for PWAs to dynamically enable dark mode for their site and have the window frame update as well. I don't think it makes sense for an out of scope page to override the app's choice of theme colour though.

What should the theme colour be when out of scope?
A: Manifest theme_color
B: Last used in-scope theme colour
C: Out of scope page theme colour

This question was first raised in: https://bugs.chromium.org/p/chromium/issues/detail?id=872121

christianliebel:
Closes #755

This change (choose one):

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

Currently, the specification does not define which theme color should be applied to an application context when the user navigates off the scope. This change attempts to clarify the behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/christianliebel/manifest/pull/879.html" title="Last updated on May 28, 2020, 12:22 AM UTC (e15b577)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/879/f1b640a...christianliebel:e15b577.html" title="Last updated on May 28, 2020, 12:22 AM UTC (e15b577)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -506,9 +506,10 @@
         [=Document/URL=] is <a>within scope</a>, in order to make it obvious
         that the user is navigating off scope. In addition, the user agent
         SHOULD reset the <a>application context</a>'s <a>default theme
-        color</a> to its default value. The off-scope document may override
-        that color through the inclusion of a valid [[HTML]] <code>meta</code>
-        element whose <code>name</code> attribute is "theme-color".
+        color</a> to a user agent-defined default value. The off-scope document
+        may override that color through the inclusion of a valid [[HTML]]
+        <code>meta</code> element whose <code>name</code> attribute is
+        "theme-color".
       </p>
       <div class="note">
         <p>
--- a/index.html
+++ b/index.html
@@ -504,7 +504,11 @@
         at least its <a>origin</a>, including whether it is served over a
         secure connection. This UI SHOULD differ from any UI used when the
         [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        that the user is navigating off scope. In addition, the user agent
+        SHOULD reset the <a>application context</a>'s <a>default theme
+        color</a> to its default value. The off-scope document may override
+        that color through the inclusion of a valid [[HTML]] <code>meta</code>
+        element whose <code>name</code> attribute is "theme-color".
       </p>
       <div class="note">
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -504,7 +504,11 @@ <h2>
         at least its <a>origin</a>, including whether it is served over a
         secure connection. This UI SHOULD differ from any UI used when the
         [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        that the user is navigating off scope. In addition, the user agent
+        SHOULD reset the <a>application context</a>'s <a>default theme
+        color</a> to its default value. The off-scope document may override
```
kenchris, line 509: Is its default value defined? or are you referring to "a User Agent defined default value"?

christianliebel: Updated. Thanks!

index.html
```diff
@@ -1672,11 +1677,11 @@ <h3>
...
           If the user agent honors the value of the <code>theme_color</code>
           member as the <a>default theme color</a>, then that color serves as
-          the <a>theme color</a> for all browsing contexts to which the
-          manifest is <a>applied</a>. However, a document may override the
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
-          <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          the <a>theme color</a> for all browsing contexts <a data-lt=
+          "within-scope-manifest">within scope</a> to which the manifest is
+          <a>applied</a>. However, a document may override the <a>default theme
```
kenchris, line 1682: Should that be MAY @marcoscaceres ?

index.html
```diff
@@ -504,7 +504,12 @@ <h2>
         at least its <a>origin</a>, including whether it is served over a
         secure connection. This UI SHOULD differ from any UI used when the
         [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        that the user is navigating off scope. In addition, the user agent
+        SHOULD reset the <a>application context</a>'s <a>default theme
```
mgiuca, line 508: I don't like the term "reset" or "default theme color", since it implies making a permanent change to the default theme color itself. Rather, could it just say "SHOULD ignore the application's default theme color, instead using the theme color of the current document". (This could be more formal, but that's the idea.)

index.html
```diff
@@ -504,7 +504,12 @@ <h2>
...
+        color</a> to a user agent-defined default value. The off-scope document
+        may override that color through the inclusion of a valid [[HTML]]
```
mgiuca, line 510: You can't use the word "may" without it being in all-caps, making it a normative requirement. And you can't impose a normative requirement on the document; all requirements are imposed on the user agent.

"The user agent MAY use the HTML meta element whose name attribute is "theme-color" of the off-scope document to theme the application context."

index.html
mgiuca: As I wrote on the other review comment: you can't use the word "may" without it being in all-caps, making it a normative requirement. And you can't impose a normative requirement on the document; all requirements are imposed on the user agent.

(I know this is a problem with the existing text as well. We should fix it.)

Issue #764, opened 2019-05-02
Add Manifest option for specifying browser app picker behavior


raymeskhoury:
**Problem:** When a user clicks a link in a web page, it may be able to be handled by many different apps installed on the system. For example, clicking a link to reddit may be able to handled by:
1. The default browser in which the link was clicked
2. A reddit PWA that's installed (via a [launch event handler](https://github.com/WICG/sw-launch/blob/master/explainer.md))
3. The reddit Android app that's installed
4. A native app that handles reddit links

Browsers/OS's can implement UI which allows users to choose which app they want to use to handle to link (an "app picker"). PWA's may want to give different _hints_ to the browser about whether or not they can handle different links into their app, e.g.
-Some PWAs may want the browser to always show the app picker for link clicks into their PWA, to give the user the option of which app to open the link with. 
-Some PWAs may want the browser to always directly open link clicks in their app without showing the app picker. This is analogous to the behavior of [Android App Links](https://developer.android.com/training/app-links).
-Some PWAs may never want the browser to open certain links in their app. For example, Facebook may have an app, but they are also an authentication provider and may not want redirects to login links to be opened in the app (e.g. https://www.facebook.com/v3.2/dialog/oauth)

The app picker needs to show up before the launch event is fired so apps need a way to define these behaviors up-front.

**Proposal:** Add a manifest entry to specify these _hints_. Note that the browser is always free to override the hint.
`link_capturing_preference: "never"` - tell the browser that this app never handles link clicks into itself and the app picker shouldn't list the app as an option. The link click should never launch the PWA or fire a launch event.
`link_capturing_preference: "ask_user"` - tell the browser that the app does handle link clicks, but the browser should ask the user whether to handle the link first. This would be the default setting if not specified.
`link_capturing_preference: "always"` - tell the browser that the app prefers to handle link clicks without the user being shown an app picker. Browsers are still free to override this and show a dialog if they choose.

By default these options apply to all link clicks in the scope of the app. However apps could also define  these options for more narrow scopes inside itself: 
```
link_capturing_preference: [
  { "/" : "always" },
  { "/v5.5/dialog/oauth" : "never" }]
```

/cc @benfredwells @dominickng @fallaciousreasoning

raymeskhoury:
@reillyeon

dominickng:
> -Some PWAs may want the browser to always show the app picker for link clicks into their PWA, to give the user the option of which app to open the link with.

Could you clarify a case where you think an app might want this behaviour?

[...]

PR #880, opened 2020-05-28
Clarify the definition of "navigation scope", "applied", and off-scope theming

### LINKED ISSUE #755: What theme color should be used when a PWA window is out of scope?

mgiuca:
Closes #755.

This change (choose one):

Commit message:

This further clarifies that a manifest is allowed to apply to URLs that
are not within its scope (that this was not clear was a left-over
mistake from #701), and that a document within scope is allowed to
override the theme color. Adds a new recommendation that an out of scope
document not be allowed to override the theme color.

This introduces new normative recommendations, only insofar as they were
already commonly understood, to my knowledge.

Navigation scope: No longer defined in terms of the URLs that a manifest
is allowed to apply to (since a manifest may apply to any URL). Rather,
it's just a set of URLs; the other normative requirements around
navigation scope define what it means.

Applied: Clarified that it is the user agent's discretion whether or not
to apply a manifest when a top-level browsing context is created, with
examples of when you would and would not apply it (previously, there was
simply no text around when a manifest should be applied).

Theme color: Fixes a "MAY" requirement imposed on the document
(requirements can only be imposed on user agents). Clarifies this
recommendation as only applying to in-scope documents, and adds a
recommendation against applying document theme for out-of-scope
documents.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/880.html" title="Last updated on May 29, 2020, 6:06 AM UTC (d07e2f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/880/f1b640a...mgiuca:d07e2f0.html" title="Last updated on May 29, 2020, 6:06 AM UTC (d07e2f0)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/880.html" title="Last updated on Oct 31, 2024, 10:34 PM UTC (0bcd1d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/880/f1b640a...mgiuca:0bcd1d2.html" title="Last updated on Oct 31, 2024, 10:34 PM UTC (0bcd1d2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1683,7 +1683,7 @@
           color=] if a document whose URL is [=within scope=] of the
           [=application context=]'s [=manifest=] includes a
           [^meta^] element whose [^meta/name^] attribute is
-          "theme-color". However, the user agent SHOULD NOT override the [=default theme
+        "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
           documents that are not [=within scope=], since the application has no control over
           these documents.
--- a/index.html
+++ b/index.html
@@ -1683,7 +1683,10 @@
           color=] if a document whose URL is [=within scope=] of the
           [=application context=]'s [=manifest=] includes a valid [[HTML]]
           <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color". The user agent SHOULD NOT override the [=default theme
+          "theme-color". However, the user agent SHOULD NOT override the [=default theme
+          color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
+          documents that are not [=within scope=], since the application has no control over
+          these documents.
           color=] for documents that are not [=within scope=], since the
           application has no control over these documents.
         </p>
--- a/index.html
+++ b/index.html
@@ -1683,7 +1683,7 @@
           color=] if a document whose URL is [=within scope=] of the
           [=application context=]'s [=manifest=] includes a
           [^meta^] element whose [^meta/name^] attribute is
-        "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
+          "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
           documents that are not [=within scope=], since the application has no control over
           these documents.
--- a/index.html
+++ b/index.html
@@ -1682,7 +1682,7 @@
           [=applied=]. However, the user agent MAY override the [=default theme
           color=] if a document whose URL is [=within scope=] of the
           [=application context=]'s [=manifest=] includes a valid [[HTML]]
-          <code>meta</code> element whose <code>name</code> attribute is
+          [^meta^] element whose [^meta/name^] attribute is
           "theme-color". However, the user agent SHOULD NOT override the [=default theme
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
           documents that are not [=within scope=], since the application has no control over
--- a/index.html
+++ b/index.html
@@ -439,10 +439,9 @@
       </h2>
       <p data-link-for="WebAppManifest">
         A <dfn data-export="">navigation scope</dfn> is a <a>URL</a> that
-        represents the set of URLs to which an <a>application context</a> can
-        be navigated while the manifest is <a>applied</a>. The <a>navigation
-        scope</a> of a manifest <var>manifest</var> is
-        <var>manifest</var>["<a>scope</a>"].
+        represents the set of URLs which are considered to be part of an
+        application. The <a>navigation scope</a> of a manifest
+        <var>manifest</var> is <var>manifest</var>["<a>scope</a>"].
       </p>
       <div class="note" data-link-for="WebAppManifest">
         <p>
@@ -1157,10 +1156,22 @@
           Applying the manifest
         </h3>
         <p>
-          A <a>manifest</a> is <dfn data-lt="apply|applying">applied</dfn> to a
-          <a>top-level browsing context</a>, meaning that the members of the
-          <a>manifest</a> are affecting the presentation or behavior of a
-          browsing context.
+          A <a>manifest</a> can be <dfn data-lt="apply|applying">applied</dfn>
+          to a <a>top-level browsing context</a>, meaning that the members of
+          the <a>manifest</a> are affecting the presentation or behavior of a
+          browsing context. Whenever a <a>top-level browsing context</a> is
+          created, the user agent MAY <a>apply</a> a manifest to it, before
+          [=navigate|navigation=] begins.
+        </p>
+        <p class="note">
+          Whether to [=apply=] a manifest, and which manifest to apply, is at
+          the discretion of the user agent, based on the user's actions. For
+          example, if the user launched an application from the system menu or
+          from a [=launching a shortcut|shortcut=], the user agent might create
+          a new [=top-level browsing context=] with that application's
+          [=manifest=] [=applied=], but it might not do so if the user simply
+          clicked a bookmark to a URL within the application's [=navigation
+          scope=].
         </p>
         <p>
           A <a>top-level browsing context</a> that has a manifest applied to it
@@ -1183,11 +1194,6 @@
             application was <a>installed</a>.
           </p>
         </div>
-        <p>
-          The appropriate time to <a>apply</a> a manifest is when the
-          <a>application context</a> is created and before
-          [=navigate|navigation=] to the <a>start URL</a> begins.
-        </p>
       </section>
       <section id="updating">
         <h3>
@@ -1671,12 +1677,15 @@
         </p>
         <p>
           If the user agent honors the value of the <code>theme_color</code>
-          member as the <a>default theme color</a>, then that color serves as
-          the <a>theme color</a> for all browsing contexts to which the
-          manifest is <a>applied</a>. However, a document may override the
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a valid [[HTML]]
           <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          "theme-color". The user agent SHOULD NOT override the [=default theme
+          color=] for documents that are not [=within scope=], since the
+          application has no control over these documents.
         </p>
         <p>
           The steps for <dfn>processing the <code>theme_color</code>
--- a/index.html
+++ b/index.html
@@ -1160,7 +1160,7 @@
           to a <a>top-level browsing context</a>, meaning that the members of
           the <a>manifest</a> are affecting the presentation or behavior of a
           browsing context. Whenever a <a>top-level browsing context</a> is
-          created, the user agent MAY <a>apply</a> a manifest to it, before
+          created, the user agent MAY <a>apply</a> a manifest to it before
           [=navigate|navigation=] begins.
         </p>
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -1687,7 +1687,6 @@
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
           documents that are not [=within scope=], since the application has no control over
           these documents.
-          application has no control over these documents.
         </p>
         <p>
           The steps for <dfn>processing the <code>theme_color</code>
--- a/index.html
+++ b/index.html
@@ -1681,7 +1681,7 @@
           [=theme color=] for all browsing contexts to which the manifest is
           [=applied=]. However, the user agent MAY override the [=default theme
           color=] if a document whose URL is [=within scope=] of the
-          [=application context=]'s [=manifest=] includes a valid [[HTML]]
+          [=application context=]'s [=manifest=] includes a
           [^meta^] element whose [^meta/name^] attribute is
           "theme-color". However, the user agent SHOULD NOT override the [=default theme
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
--- a/index.html
+++ b/index.html
@@ -1687,7 +1687,6 @@
           color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
           documents that are not [=within scope=], since the application has no control over
           these documents.
-          color=] for documents that are not [=within scope=], since the
           application has no control over these documents.
         </p>
         <p>


index.html
```diff
@@ -1157,10 +1156,22 @@ <h3 id="applying">
...
+        </p>
+        <p class="note">
+          Whether to [=apply=] a manifest, and which manifest to apply, is at
+          the discretion of the user agent, based on the user's actions. For
+          example, if the user launched an application from the system menu or
+          from a [=launching a shortcut|shortcut=], the user agent might create
+          a new [=top-level browsing context=] with that application's
+          [=manifest=] [=applied=], but it might not do so if the user simply
+          clicked a bookmark to a URL within the application's [=navigation
+          scope=].
```
kenchris, line 1174: so you mean that it might reuse an existing one?

mgiuca: This is meant to imply that it might either reuse an existing one (navigate the user's current tab) or open a new browser tab, as opposed to creating an app window.

The point is that the browser can decide based on the user's action whether to open an app window or a browser tab. (This is really just because the current spec says "a manifest is applied" but it doesn't say if it's always applied, or when exactly a user agent should apply a manifest. I'm just being explicit and saying "the user agent can do that at the appropriate time, but doesn't have to".)

marcoscaceres: ... or could even apply a user configured one...

index.html
```diff
@@ -439,10 +439,9 @@ <h2>
       </h2>
       <p data-link-for="WebAppManifest">
         A <dfn data-export="">navigation scope</dfn> is a <a>URL</a> that
-        represents the set of URLs to which an <a>application context</a> can
-        be navigated while the manifest is <a>applied</a>. The <a>navigation
-        scope</a> of a manifest <var>manifest</var> is
-        <var>manifest</var>["<a>scope</a>"].
+        represents the set of URLs which are considered to be part of an
```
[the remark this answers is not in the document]
mgiuca: > URLs outside the scope can still be part of the web application (e.g., an about.html page that opens in an overlay window), it's just that the manifest members are no longer in effect there.

But the manifest members _are_ in effect there. That's exactly why I'm making this change. When you're off-scope, all of the manifest members are still affecting the application context, because we want it to still feel like you're in the app.

Are you disagreeing with the _behaviour_ allowed by this change, or just the wording? i.e., do you think Chrome should not render about.html page in an overlay window with the full theming of the app manifest behind it (the theme colour, app name, icon, display mode, etc)? Or do you just disagree on the semantics?

mgiuca:
(why is there no way for me to reply to this comment inline?)

> ... or could even apply a user configured one...

Not sure what you're suggesting here or where this fits into the sentence.

index.html
mgiuca: Do you think it's worth pointing that out, Marcos? Seems like a kind of advanced use case that would be allowed, but not sure we need to call it out here.

marcoscaceres: sorry, no need to call that out... was just an example for @kenchris.

index.html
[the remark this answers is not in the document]
marcoscaceres: > Are you disagreeing with the behaviour allowed by this change, or just the wording?

Just the wording. Agree about the behavior. 

With the wording, I think it risks overstepping into defining "what is/is not a web application"... the original definition was just about the "manifest applies to set of URLs"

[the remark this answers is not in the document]
mgiuca: OK maybe we can workshop this. I don't want to return to the old definition, since it directly ties the scope to the set of URLs that the manifest "applies" to, where "applies" has a strict technical definition of having all of its members used by the top-level browsing context (and that can now be true for all URLs).

It wasn't my intention to try and define a web application here. What's really important here is three things:

1. A "navigation scope" is just a URL.
2. Really though, the URL represents a set of URLs, which is all the URLs that return true from the "within scope" algorithm.
3. The actual normative meaning of the navigation scope is defined by every other thing that uses the "navigation scope" or "within scope" concept.

So 2 and 3 are taken care of elsewhere. All that this definitional sentence really needs to do is say "it's a URL". I thought it was a bit empty, so I added "it represents the set of URLs which are considered to be part of an application". That feels a bit non-normative, so we can take it out, and just give the facts about "navigation scope" as if we were defining a typedef in a computer program, and not give its meaning in this statement. Simply: "A navigation scope is a URL."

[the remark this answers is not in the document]
marcoscaceres: > OK maybe we can workshop this. I don't want to return to the old definition, since it directly ties the scope to the set of URLs that the manifest "applies" to, where "applies" has a strict technical definition of having all of its members used by the top-level browsing context (and that can now be true for all URLs).

I'm confused why "that can now be true for all URLs"? 

> So 2 and 3 are taken care of elsewhere. All that this definitional sentence really needs to do is say "it's a URL".

> "A navigation scope is a URL.

Although the above is true (it's a URL indeed), that doesn't really define the role of the navigation scope as it relates to applying the manifest. That needs to be either here, or somewhere that says roughly "when you navigate, keep applying the manifest until you are no longer within the navigation scope".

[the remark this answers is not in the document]
mgiuca: (FYI I'm leaving this on hold until #882 lands, which will change the definition of "navigation scope", though it keeps the clause in question.)

> I'm confused why "that can now be true for all URLs"?

The current definition defines navigation scope as "the set of URLs to which an application context can be navigated while the manifest is applied." As of #701 (which landed nearly two years ago), the set of URLs to which an application context can be navigated while the manifest is applied is all URLs. If I navigate an app window on https://mysite.com to https://yoursite.com, the manifest is still _applied_.

So we can't have this definition, if the intention is to allow the manifest to be applied when you navigate out of scope. (This isn't directly contradicted by normative text, but the absence of any text that forbids out-of-scope navigation implies that you can navigate out of scope in an application context where the manifest is applied. A non-normative note directly contradicts this definition: "Nothing prevents an application context from navigating to a URL that is outside of the application's navigation scope , while still having the manifest applied to it.")

> Although the above is true (it's a URL indeed), that doesn't really define the role of the navigation scope as it relates to applying the manifest. That needs to be either here, or somewhere that says roughly "when you navigate, keep applying the manifest until you are no longer within the navigation scope".

Wait: "until you are no longer within the navigation scope" implies that you would not apply the manifest when navigated off-scope. Do we have a disagreement over the meaning of the term "apply"? I am considering a manifest to be applied permanently to the application context, regardless of whether it's navigated off-scope. When you navigate off-scope, it is RECOMMENDED that the application window shows some UI indicating that you are off scope, but the manifest is still applied. We would still show the theme colour, title and icon of the application from the manifest. So the spec needs to state that the manifest is always applied to both in-scope and off-scope documents.

I think we can add some text to "applying the manifest" that says "When a manifest is applied to a top-level browsing context, it is continually applied throughout the lifetime of the browsing context."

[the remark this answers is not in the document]
marcoscaceres: Discussed on call... manifest members MAY apply outside the navigation scope (e.g., theme color). We also discussed that a new/independent top-level browsing context is not created.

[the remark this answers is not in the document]
mgiuca: We just discussed this in our F2F. Agreed that the manifest _is_ applied to the out-of-scope page, based on testing Chrome, Firefox and Safari (the last of those we're not 100% sure but we suspect that's happening).

There isn't any normative text that says this, but it's implied by the fact that we do not make changes to the HTML navigation system, therefore navigation is expected to proceed as normal to an out-of-scope URL.

There is non-normative text that supports this:

> Unlike previous versions of this specification, user agents are no longer required or allowed to block off-scope navigations, or open them in a new top-level browsing context. This practice broke some sites that navigate to an off-scope URL (e.g., to perform third-party authentication). See Issue 646.

I proposed that since it's been 2 years, we collapse this text to not mention the history of it, just say that user agents are not allowed to block or transfer the navigation to another TLBC.

@marcoscaceres let me know if I misunderstood something.

benfrancis:
> I proposed that since it's been 2 years, we collapse this text to not mention the history of it, just say that user agents are not allowed to block or transfer the navigation to another TLBC.

Are you sure you want to forbid user agents from transferring *any* navigation to another top level browsing context? Is this not a product decision?

benfrancis:
I thought it might be useful to provide some examples of implementations I've worked on outside of Android and iOS, in order to not constrain the thinking to contemporary mobile operating systems and be more open minded about how this specification might be used in the future.

**Firefox OS 2.6**

In Firefox OS 2.6 the web browser was merged with the core system UI such that users could seamlessly transition between "pinned" web apps, and the rest of the web. This is in contrast with other mobile operating systems in which there are much firmer boundaries between installed applications.

When browsing the web, web applications could be "pinned" (installed) using their web app manifest. Whenever the user navigated to a URL within the scope of a pinned app, the manifest metadata was applied to the browsing context and whenever they navigated away from the scope of the app, the manifest would be un-applied.

When navigating to a URL within the scope of a pinned app with a standalone display mode, the URL bar would shrink down into the status bar and turn into a title bar to look more like an app. When navigating out of the scope of the pinned app, the URL bar would expand out into a full URL bar to look more like a browser. (The user could always access the full URL bar by tapping to expand it).

pin_app


There's a video [here](https://www.youtube.com/watch?v=CWUIvo0BzBU) which shows part of that behaviour.

**Krellian Kiosk**

I'm currently working on an [implementation](https://github.com/krellian/kiosk) of a web application runtime where a single web application may be installed to an interactive kiosk for use by the public. In this use case, the administrator wants to ensure that *only* the installed web application can be navigated to, and any attempted navigations outside of the scope of the web application are blocked. I believe Chrome OS has a kiosk mode which does something similar.

Another related use case is in deploying a single web application to a digital signage display, which has completely passive content and no user interaction on the display itself. Web app installation and navigation are controlled over the network from a separate controller application.

kiosk


The reason I'm providing these examples is to try not to constrain future innovation in user agents by simply describing the behaviour of existing smartphone web browsers in the specification. I suggest it would better to allow some flexibility in the specification for user agents to decide when to apply a manifest and how navigations outside of the manifest's scope should be handled.

How about defining scope as "the set of URLs to which the manifest applies." This could be qualified by saying that it doesn't mean that user agents can't also apply some aspects of the manifest's metadata to other URLs if they are navigated to from an application context, but that the metadata is primarily intended to apply to that set of URLs, not the whole web.

index.html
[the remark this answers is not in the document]
marcoscaceres: That sounds right to me. Thanks for the summary.

mgiuca:
@benfrancis it seems like your practical objection to this change comes from a desire to have additional restricted modes (such as kiosk modes) that do restrict PWAs to their scope or a restrictive set of URLs.

I would suggest that such a mode falls outside of the manifest spec, and into the realm of (legitimate) user agent interventions, by way of offering their users (administrators) additional control over what URLs can be browsed to. It's very similar to a parental control feature that a web browser may offer, to let parents restrict the set of browsable URLs to a specific allow-list. You don't need to the permission of the HTML spec to prevent navigation to a URL that's been banned by the parent, it's essentially a deviation from the spec requested by the user.

I would argue it makes sense to keep these two things (the scope of a PWA, and a kiosk-like prevention of navigating outside the PWA) separate. There's no reason kiosk mode needs to necessarily be limited to the scope of the PWA. For example, you might let the administrator configure an allow-list, so that you can navigate to certain other origins within the PWA, but not the whole internet.

> Are you sure you want to forbid user agents from transferring any navigation to another top level browsing context? Is this not a product decision?

It isn't a product decision, because it violates the basic assumption that sites make about how HTML navigation works. (As we discovered two years ago in #646.) This isn't obvious, to me anyway, it took months of analysis, discussion, and experimentation (as you can see by reading over #646, the corresponding Chrome bug https://crbug.com/861618, and the corresponding manifest PR #701), to come to the realisation that this wasn't a mere product decision; allowing UAs to open target=_self links in a new top-level browsing context was breaking websites because it violated the normal navigation flow of HTML. So, we changed the Manifest spec to explicitly _forbid_ UAs from doing that. This PR was not intended to re-open the debate; it's just fixing up a piece of text mistakenly left over from #701 .

As I said above, UAs are free to make a product decision to restrict the set of navigable URLs, at the request of the user, but this should be opted in by the user or administrator (e.g., by using a special kiosk mode). The normal operating mode of a compliant browser should not break websites.

> How about defining scope as "the set of URLs to which the manifest applies." This could be qualified by saying that it doesn't mean that user agents can't also apply some aspects of the manifest's metadata to other URLs if they are navigated to from an application context, but that the metadata is primarily intended to apply to that set of URLs, not the whole web.

That seems to be more or less what the spec (erroneously) says today: scope is "the set of URLs to which the manifest applies", but elsewhere in the spec it says that the manifest can apply to any URL, not just one in the navigation scope. Those are contradictory statements. I'm looking for clear normative text here that can be implemented in a web browser. "primarily intended to apply" is not a clear statement.

I think the change suggested in this PR is entirely compatible with your kiosk feature.

benfrancis:
> @benfrancis it seems like your practical objection to this change comes from a desire to have additional restricted modes (such as kiosk modes) that do restrict PWAs to their scope or a restrictive set of URLs.

Actually my main concern is that the specification does not prevent future user agents from coming up with better solutions to the off-scope navigation problem than the current in-app browser trend. The examples I gave were just to illustrate that the way Android and iOS currently work are not the only ways user agents can work.

> I would suggest that such a mode falls outside of the manifest spec, and into the realm of (legitimate) user agent interventions, by way of offering their users (administrators) additional control over what URLs can be browsed to.

As long as the specification makes clear that such user agent interventions are in fact legitimate, then I agree that's fine.

> It isn't a product decision, because it violates the basic assumption that sites make about how HTML navigation works. (As we discovered two years ago in #646.)

Thank you for the additional context, I hadn't read that thread before. Having read through all of the rationale I can see why current browsers have converged on option b1 from https://github.com/w3c/manifest/issues/646#issuecomment-402615266. In testing the latest version of Firefox Preview (Fenix) it appears that is the latest behaviour there too.

However, now that I better understand the rationale I have another concern. Does this change not break the concept of deep linking as described in [section 3.2](https://www.w3.org/TR/appmanifest/#deep-links) and the Abstract section?

If all off-scope navigations must stay inside the current application context (but with added chrome), it is no longer possible to deep link between installed web applications. At least not within the constraints of the Android and iOS application frameworks.

The Firefox OS 2.6 example I gave above actually most closely resembled option b1, but deep linking was possible because it essentially removed the distinction between an app window and browser window. It was therefore possible to seamlessly transition between the two states. A single browsing context could transition between having different manifests applied and the same window could effectively transition from one app to another. As far as I know that is not possible on Android or iOS.

Would you agree that with this change, deep linking could only ever happen if the source of a navigation explicitly requests a new window via `target=_blank` or `window.open()`? And even then some of the same technical constraints may cause problems. That seems like a real shame.

In the interests of getting to Candidate Recommendation I don't particularly object to changing the definition of scope to match what current browsers do, but I do want to highlight the potential implications for future user agents which may not be operating within the same constraints.

mgiuca:
> However, now that I better understand the rationale I have another concern. Does this change not break the concept of deep linking as described in [section 3.2](https://www.w3.org/TR/appmanifest/#deep-links) and the Abstract section?
> 
> If all off-scope navigations must stay inside the current application context (but with added chrome), it is no longer possible to deep link between installed web applications. At least not within the constraints of the Android and iOS application frameworks.
> 
> The Firefox OS 2.6 example I gave above actually most closely resembled option b1, but deep linking was possible because it essentially removed the distinction between an app window and browser window. It was therefore possible to seamlessly transition between the two states. A single browsing context could transition between having different manifests applied and the same window could effectively transition from one app to another. As far as I know that is not possible on Android or iOS.

I think I understand that concern. There's some degree of incompatibility between the following two designs:

1. Navigating _from_ an in-scope URL in an app context _to_ an out-of-scope URL remains in the app context, but with a navigation bar UI. [That's the behaviour I'm trying to capture.]
2. Navigating _from_ a normal browsing context _to_ an in-scope URL transitions into an app context. [That's what you're talking about above as deep links.]

The sort-of-incompatibility comes from the inconsistency that when moving _outwards_, the app context is fixed, but when moving _inwards_, the app context changes. To your point, when moving _from_ an in-scope URL in an app context _to_ a URL in another app's scope, there's direct conflict between the above two.

The way this resolves in my mind is that a browsing context never changes its application context. Number 2 above isn't true. If you ever navigate to a URL inside another app's context that is explicitly captured by that app (see #764), then it should not simply transition the current browsing context into an application context, it should create a new top-level browsing context in that new app context. That's how native apps typically work; you don't somehow "transform" one app window or activity into that of another application. You construct a new activity in that new application.

I think the whole model is simpler, and consistent, if we don't allow a top-level browsing context to change which manifest is applied to it (if any) once it has been created.

benfrancis:
Hi @mgiuca, thank you for responding.

> The way this resolves in my mind is that a browsing context never changes its application context. Number 2 above isn't true. If you ever navigate to a URL inside another app's context that is explicitly captured by that app (see #764), then it should not simply transition the current browsing context into an application context, it should create a new top-level browsing context in that new app context. 

OK, that makes sense. It was not clear to me from the specification and proposed text that app-to-app deep linking would be possible though. Do you think this could be made more explicit?

The discussion in #764 looks like a promising potential way forward for deep linking into installed web applications.

> That's how native apps typically work; you don't somehow "transform" one app window or activity into that of another application. You construct a new activity in that new application.

I agree that is how Android and iOS work. But I would argue this is the result of trying to shoehorn the web into the native app model of existing mobile operating systems, not necessarily the optimal design for web applications which would work to the strengths of the web.

I think we may have to agree to disagree that the optimal behaviour is for every installed web application to act as a mini browser that can be navigated to any URL on the web, and that by default a link only ever goes from browser to app or app to app, and never app to browser. But I concede that in order to move towards a Candidate Recommendation, describing the current behaviour of implementations is the most pragmatic and low-friction way forward.

mgiuca:
Hey @benfrancis 

> OK, that makes sense. It was not clear to me from the specification and proposed text that app-to-app deep linking would be possible though. Do you think this could be made more explicit?

Sure. Maybe we can tackle it as part of the #764 work. (Or, if you feel like it, feel free to put up a PR suggesting stronger wording for deep linking. FYI that section is very old and I don't think it's been touched since I've been an editor on the Manifest.)

> I agree that is how Android and iOS work. But I would argue this is the result of trying to shoehorn the web into the native app model of existing mobile operating systems, not necessarily the optimal design for web applications which would work to the strengths of the web.
> 
> I think we may have to agree to disagree that the optimal behaviour is for every installed web application to act as a mini browser that can be navigated to any URL on the web, and that by default a link only ever goes from browser to app or app to app, and never app to browser. But I concede that in order to move towards a Candidate Recommendation, describing the current behaviour of implementations is the most pragmatic and low-friction way forward.

OK. As I've said above, at a personal level, I agree with you wholeheartedly. I do not like it when apps keep you inside their frame when you're off browsing the web, and I turn it off for every app that I can. There's a reason we have full-featured web browser apps! However, I must note the reality that the web platform is in competition to native apps platforms, and that gives developers a certain amount of sway over us... they overwhelmingly want to keep users inside their app frames by default (so that the user can get back to their app with one tap), and so it behooves us as platform designers to accomodate that, in the way that best respects the user (i.e., giving the user controls to grab the URL and re-open the page in a web browser, for instance). If we don't do that, we'll find sites building their own browser UI in an iframe instead, which makes it much harder for us to intervene on the part of the user. So that's why I'm cautiously supportive of this "navigation off scope from within the app" behaviour. (And there are other, much more legitimate use cases, such as the authentication-on-another-origin case.)

index.html
```diff
@@ -1671,12 +1677,15 @@ <h3>
...
-          manifest is <a>applied</a>. However, a document may override the
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a valid [[HTML]]
           <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          "theme-color". The user agent SHOULD NOT override the [=default theme
```
dmurph, line 1686:
```suggestion
          "theme-color". However, the user agent SHOULD NOT override the [=default theme
          color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
          documents that are not [=within scope=], since the application has no control over
          these documents.
```

dmurph:
Changes look good other than that one sentence clarification.

index.html
```diff
@@ -1671,12 +1677,15 @@ <h3>
...
-          member as the <a>default theme color</a>, then that color serves as
-          the <a>theme color</a> for all browsing contexts to which the
-          manifest is <a>applied</a>. However, a document may override the
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a valid [[HTML]]
           <code>meta</code> element whose <code>name</code> attribute is
```
marcoscaceres, line 1685:
```suggestion
          [^meta^] element whose [^meta/name^] attribute is
```

index.html
```diff
@@ -1671,12 +1677,18 @@ <h3>
...
-          the <a>theme color</a> for all browsing contexts to which the
-          manifest is <a>applied</a>. However, a document may override the
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
-          <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a valid [[HTML]]
```
marcoscaceres, line 1684:
```suggestion
          [=application context=]'s [=manifest=] includes a valid
```

index.html
```diff
@@ -1671,12 +1677,18 @@ <h3>
...
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
-          <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a
+          [^meta^] element whose [^meta/name^] attribute is
+          "theme-color". However, the user agent SHOULD NOT override the [=default theme
```
marcoscaceres, line 1686:
```suggestion
        "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
```

index.html
```diff
@@ -1671,12 +1677,18 @@ <h3>
...
-          <a>default theme color</a> through the inclusion of a valid [[HTML]]
-          <code>meta</code> element whose <code>name</code> attribute is
-          "theme-color".
+          member as the [=default theme color=], then that color serves as the
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a
+          [^meta^] element whose [^meta/name^] attribute is
+        "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
```
marcoscaceres, line 1686:
```suggestion
          "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
```

index.html
```diff
@@ -1671,12 +1677,18 @@ <h3>
...
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a
+          [^meta^] element whose [^meta/name^] attribute is
+          "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
+          color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
+          documents that are not [=within scope=], since the application has no control over
+          these documents.
+          color=] for documents that are not [=within scope=], since the
```
dmurph, line 1690:
```suggestion
```

index.html
```diff
@@ -1671,12 +1677,18 @@ <h3>
...
+          application has no control over these documents.
```
marcoscaceres, line 1691:
```suggestion
          these documents.
```

index.html
```diff
@@ -1671,12 +1677,17 @@ <h3>
...
+          [=theme color=] for all browsing contexts to which the manifest is
+          [=applied=]. However, the user agent MAY override the [=default theme
+          color=] if a document whose URL is [=within scope=] of the
+          [=application context=]'s [=manifest=] includes a
+          [^meta^] element whose [^meta/name^] attribute is
+          "[^meta/name/theme-color^]". However, the user agent SHOULD NOT override the [=default theme
+          color=] via an [^meta^] element whose [^meta/name^] attribute is "theme-color" for
+          documents that are not [=within scope=], since the application has no control over
+          these documents.
+          application has no control over these documents.
```
dmurph, line 1690:
```suggestion
```

marcoscaceres (approved):

PR #882, opened 2020-05-29
chore: refactor and export scope terms

marcoscaceres:
This change (choose one):

Commit message:

Export "manifest/within scope" and "URL/within scope".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/882.html" title="Last updated on Jun 11, 2020, 7:48 AM UTC (3ed7528)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/882/632df40...3ed7528.html" title="Last updated on Jun 11, 2020, 7:48 AM UTC (3ed7528)">Diff</a>


index.html
```diff
@@ -522,13 +521,13 @@ <h2>
...
       <p>
         If the <a>application context</a>'s <a>active document</a>'s
-        [=Document/URL=] is not <a data-lt="within-scope-manifest">within
-        scope</a> of the <a>application context</a>'s manifest, the user agent
-        SHOULD show a prominent UI element indicating the [=Document/URL=] or
-        at least its <a>origin</a>, including whether it is served over a
-        secure connection. This UI SHOULD differ from any UI used when the
-        [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        [=Document/URL=] is not [=within manifest scope=] of the <a>application
```
mgiuca, line 524: How strongly do you feel about renaming this? I deliberately named this "within scope" (the same name as the other definition), designed to be a sort of C++ overloading of the name.

So that anywhere else, I can write "if _x_ is within scope of _app manifest_", without having to clarify that I am invoking this definition, or the other one. (And link to the correct one.) I would prefer to be able to keep doing that, rather than having to update all of the other usages of this to "within manifest scope of _app manifest_".

mgiuca: I see you updated some (maybe not all?) of the other references to distinguish "within the navigation scope of" and "within manifest scope of". Note that there are other references to this concept from other WICG documents (even if this is not properly exported), like Web Share and File Handling.

I would prefer that we don't have to add that verbosity to what I imagine are a huge amount of usages.

marcoscaceres: > How strongly do you feel about renaming this? I deliberately named this "within scope" (the same name as the other definition), designed to be a sort of C++ overloading of the name.

The rename is so it's more clear when used in other specs - want to avoid people accidentally using the wrong one ("_URL_ within scope of _other URL_" VS "within manifest scope"). When used locally, we can still do "within scope".

mgiuca: I'm glad you are :) But I'd still prefer for them not to have to be so verbose.

Context: I've written enough sentences like this one:

> If the application context's active document's URL is not within scope of the application context's manifest

That's already pretty hard to read. We would have to change it to:

> If the application context's active document's URL is not within **manifest** scope of the application context's manifest

I added this to allow us to be less verbose ("within scope of the manifest" as opposed to "within scope of the navigation scope of the manifest"). Having to say "within manifest scope of the manifest" is hardly better than just deleting this convenience wrapper and using the URL-based one each time.

I'm not too concerned about people accidentally using the wrong one, since if they do, it just means they'll be linking to a slightly wrong position in the page. The meaning is clear and unambiguous: you're either talking about being within scope of a URL or within scope of a manifest.

marcoscaceres: Happy to change them to simpler ones... um... how about:

```
[=manifest/within scope of=]
[=URL/within scope of=]
```

Then it takes the complexity out of reading the sentences, but puts the problem on the spec editor... they just get displayed as "within scope of" and links to the right thing.

mgiuca: That's great if you can pull it off. (Note: I'm not concerned about what the code looks like, just what the document looks like when rendered.) If we can preserve the overloading (at the text level), that would be awesome.

marcoscaceres: actually, there was a sentence where "within scope of" didn't work...  so "[=X/within scope=]" works all round... updating.

index.html
```diff
@@ -460,31 +460,30 @@ <h2>
...
+        following algorithm returns `true`:
       </p>
-      <ol>
-        <li>Let <var>scopePath</var> be the [=string/concatenation=] of
-        <var>scopes</var>'s <a data-cite="URL#concept-url-path">path</a>, using
-        U+002F (/).
+      <ol class="algorithm">
+        <li>If |target| and |scope| are not [=same origin=], return `false`.
+        </li>
+        <li>Let |scopePath:string| be the [=string/concatenation=] of
```
mgiuca, line 470: "using" is correct here. See the definition of [concatenation](https://infra.spec.whatwg.org/#string-concatenate). This is supposed to place "/" in between each piece of the path, not at the end of it.

Perhaps you could clarify it by writing "using the separator "/"" or "using "/" as a separator".

index.html
```diff
@@ -460,31 +460,30 @@ <h2>
...
-        <li>Let <var>targetPath</var> be the [=string/concatenation=] of <var>
-          target</var>'s <a data-cite="URL#concept-url-path">path</a>, using
-          U+002F (/).
+        <li>Let |targetPath:string| be the [=string/concatenation=] of
+        |target|'s [=URL/path=] and U+002F (/).
         </li>
-        <li>If <var>target</var> is <a>same origin</a> as <var>scope</var> and
-        <var>targetPath</var> starts with <var>scopePath</var>, return
-        <code>true</code>.
+        <li>If |targetPath| starts with |scopePath:string|, return `true`.
```
mgiuca, line 476: Maybe this and the next line could be combined by saying "Return a Boolean indicating whether |targetPath| starts with |scopePath|." Or even just "Return true if |targetPath| starts with |scopePath|, or false otherwise."

Since if I were writing this in C++, I would write:

```c++
return targetPath.starts_with(scopePath);
```

not

```c++
if (targetPath.starts_with(scopePath))
  return true;
else
  return false;
```

index.html
```diff
@@ -460,31 +460,30 @@ <h2>
...
+        <li>Otherwise, return `false`.
         </li>
       </ol>
       <p>
-        A <a>URL</a> <var>target</var> is said to be <dfn data-lt=
-        "within-scope-manifest">within scope of a manifest</dfn>
-        <var>manifest</var> if <var>target</var> is <a>within scope</a> of the
-        navigation scope of <var>manifest</var>.
+        A [=URL=] |target:URL| is said to be <dfn data-export="" data-dfn-for=
+        "manifest" data-lt="within scope">within the navigation scope of a web
```
mgiuca, line 483: Can this term just be "within scope", otherwise it is again too verbose (as it is intended as a way to write a short-hand).

"**within the navigation scope of a web manifest** _manifest_" is longer than "**within scope** of the navigation scope of _manifest_", so there is no point using this short-hand if its name is that long.

index.html
```diff
@@ -460,31 +460,30 @@ <h2>
...
+        manifest</dfn> {{WebAppManifest}} |manifest:WebAppManifest| if [=URL=]
+        |target:URL| is [=manifest/within scope=] of the |manifest|'s
```
mgiuca, line 485: I think you mean [=URL/within scope=].

index.html
```diff
@@ -460,31 +460,30 @@ <h2>
...
+        {{WebAppManifest/scope}} member (once resolved).
```
mgiuca, line 486: "once resolved" -- this is a colloquial term (URL "resolution") but I don't think it's defined anywhere.

Can you say "once [=processed=]" (linking to "processing a manifest"), because that's really what you're saying: we're talking about the scope of the manifest after processing.

Better yet, move this to the definition of "navigation scope of a manifest" above. That is:

> The navigation scope of a manifest _manifest_ is _manifest_[scope], once manifest has been processed.

or

> The navigation scope of a **processed manifest** _manifest_ is _manifest_[scope].

Then, here, you should say "of the navigation scope of _manifest_" rather than "of the _manifest_'s scope member".

index.html
```diff
@@ -1978,8 +1978,8 @@ <h3>
               URL</var> as the base URL. If the result is failure, <a>issue a
               developer warning</a> and [=iteration/continue=].
               </li>
-              <li>If <var>shortcut</var>["url"] is not <a>within scope</a> of
-              <var>manifest URL</var>, <a>issue a developer warning</a> and
+              <li>If <var>shortcut</var>["url"] is not [=URL/within scope=]
```
mgiuca, line 1981: There seem to be changes leaking both ways between this and #886. I think you should not change this here since you have a separate PR to fix this bug.

index.html
```diff
@@ -1951,11 +1951,11 @@ <h3>
...
         </p>
         <p>
-          The steps for <dfn>processing the <code>shortcuts</code> member</dfn>
-          are given by the following algorithm. The algorithm takes a
+          The steps for <dfn>processing the `shortcuts` member</dfn> are given
+          by the following algorithm. The algorithm takes a
           <a>sequence</a>&lt;<a>ShortcutItem</a>&gt; <var>shortcuts</var> and a
-          <a>URL</a> <var>manifest URL</var>. This algorithm returns a
-          <a>sequence</a>&lt;<a>ShortcutItem</a>&gt;.
+          <a>URL</a> |manifest URL:URL|, and [=URL=] <var>scope URL</var>. This
```
mgiuca, line 1957: Here too (leave this change for #886).

index.html
```diff
@@ -2642,8 +2642,9 @@ <h3>
         </h3>
         <p>
           The <dfn>url</dfn> member of a <a>ShortcutItem</a> is the <a>URL</a>
-          <a data-lt="within scope of a manifest">within the application's
-          scope</a> that opens when the associated shortcut is activated.
+          that opens when the associated shortcut is activated. When resolved,
+          the {{ShortcutItem/url}} must be [=URL/within scope=] of scope URL,
+          it gets ignored.
```
mgiuca, line 2647: This sentence is ungrammatical ("it gets ignored"). Also, do not use the keyword "must" in this definition which is not strictly imposing requirements. Just say "is the URL [=manifest/within scope=] of the manifest that opens when the associated shortcut is activated." (It doesn't have to be explicit about everything, since that's covered by algorithms elsewhere.)

index.html
```diff
@@ -1140,8 +1138,9 @@ <h3>
           member</a> given <var>manifest</var>["<a>related_applications</a>"].
           </li>
           <li>Set <var>manifest</var>["<a>shortcuts</a>"] to the result of
-          running <a>processing the <code>shortcuts</code> member</a> given
-          <var>manifest</var>["<a>shortcuts</a>"] and <var>manifest URL</var>.
+          running <a>processing the `shortcuts` member</a> given
+          <var>manifest</var>["<a>shortcuts</a>"], <var>manifest URL</var>, and
+          <var>scope URL</var>.
```
mgiuca, line 1143: Here too (leave this change for #886).

mgiuca:
(FYI, I'm signing off for the weekend. We'll continue the reviews on Monday! Have a good weekend.)

index.html
marcoscaceres: Well, we don't ever need to write "within the navigation scope of a web manifest"... it's only defined as that here. In every other spec or local instance, one just writes [=manifest/within scope=].

index.html
marcoscaceres: good catch.

index.html
marcoscaceres: D'oh... seems wires got crossed by me jumping back and forth between the two branches... will split this up again...

index.html
marcoscaceres: agree... will drop the redundant part.

index.html
mgiuca: (Not bothered with what we have to write in the source code, but what it reads like.)

Does `[=manifest/within scope]=` render as "within scope" or "within the navigation scope of a web manifest"? If the former, I am happy.

However, I would still like to define this as "within scope of a web manifest" rather than "within the navigation scope of a web manifest" (the latter is too verbose and doesn't match the previous definition, "within scope of a URL").

index.html
```diff
@@ -460,31 +460,29 @@ <h2>
...
+        |scopes|'s [=URL/path=] using U+002F (/) as the separator.
         </li>
-        <li>If <var>target</var> is <a>same origin</a> as <var>scope</var> and
-        <var>targetPath</var> starts with <var>scopePath</var>, return
-        <code>true</code>.
+        <li>Let |targetPath:string| be the [=string/concatenation=] of
+        |target|'s [=URL/path=] using U+002F (/) as the separator.
         </li>
-        <li>Otherwise, return <code>false</code>.
+        <li>Return a boolean indicating if |targetPath| starts with
```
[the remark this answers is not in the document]
marcoscaceres: TIL... will keep lowercase, as it's linking to infra.

index.html
marcoscaceres: I took all the above (plus the a good weekend break) and rewrote various things...

index.html
```diff
@@ -460,31 +460,29 @@ <h2>
...
         </li>
       </ol>
       <p>
-        A <a>URL</a> <var>target</var> is said to be <dfn data-lt=
-        "within-scope-manifest">within scope of a manifest</dfn>
-        <var>manifest</var> if <var>target</var> is <a>within scope</a> of the
-        navigation scope of <var>manifest</var>.
+        A [=URL=] |target:URL| is said to be <dfn data-export="" data-dfn-for=
+        "processed manifest" data-lt="within scope">within the navigation scope
+        of a processed manifest</dfn> {{WebAppManifest}}
```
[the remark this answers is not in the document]
marcoscaceres: Tried to integrate this also... hopefully better now.

index.html
```diff
@@ -1978,7 +1976,7 @@ <h3>
               URL</var> as the base URL. If the result is failure, <a>issue a
               developer warning</a> and [=iteration/continue=].
               </li>
-              <li>If <var>shortcut</var>["url"] is not <a>within scope</a> of
+              <li>If <var>shortcut</var>["url"] is not [=URL/within scope=] of
```
marcoscaceres, line 1979: Did this and the next ones in this PR otherwise we get ReSpec linking errors...

index.html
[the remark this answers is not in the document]
mgiuca: Looks good, but I still would prefer to call it "within scope of a manifest" rather than "within the navigation scope of a processed manifest" (the "navigation scope" is implied, and the "processed" is explicit later).

And I don't understand why you say "within scope of the processed manifest's scope member" when you could say "within scope of the navigation scope of the processed manifest"; now those are defined as equivalent, so you may as well use the term you defined.

marcoscaceres:
@mgiuca looks like we were super close on this one and #880. Let me know how you'd like to proceed,

index.html
```diff
@@ -437,11 +437,11 @@ <h2>
         Navigation scope
       </h2>
       <p data-link-for="WebAppManifest">
-        A <dfn data-export="">navigation scope</dfn> is a <a>URL</a> that
-        represents the set of URLs to which an <a>application context</a> can
-        be navigated while the manifest is <a>applied</a>. The <a>navigation
-        scope</a> of a manifest <var>manifest</var> is
-        <var>manifest</var>["<a>scope</a>"].
+        The {{WebAppManifest/scope}} member of a [=processed manifest=] is the
```
mgiuca, line 440: Nit: I think I'd prefer this to be phrased around the other way to avoid it looking like Yoda wrote it :)

"The **navigation scope** of an application context is the `scope` member of the processed manifest that is applied to it."

(As a bonus: we link this back up to the concept of "applied".)

index.html
```diff
@@ -460,31 +460,28 @@ <h2>
...
+        |scopes|'s [=URL/path=] using U+002F (/) as the separator.
         </li>
-        <li>If <var>target</var> is <a>same origin</a> as <var>scope</var> and
-        <var>targetPath</var> starts with <var>scopePath</var>, return
-        <code>true</code>.
+        <li>Let |targetPath:string| be the [=string/concatenation=] of
+        |target|'s [=URL/path=] using U+002F (/) as the separator.
         </li>
-        <li>Otherwise, return <code>false</code>.
+        <li>Return a [=boolean=] indicating if |targetPath| starts with
```
mgiuca, line 476: s/if/whether

index.html
```diff
@@ -460,31 +460,28 @@ <h2>
...
+        |scopePath:string|.
         </li>
       </ol>
       <p>
-        A <a>URL</a> <var>target</var> is said to be <dfn data-lt=
-        "within-scope-manifest">within scope of a manifest</dfn>
-        <var>manifest</var> if <var>target</var> is <a>within scope</a> of the
-        navigation scope of <var>manifest</var>.
+        A [=URL=] |target:URL| is <dfn data-export="" data-dfn-for=
+        "processed manifest" data-lt="within scope">within navigation scope of
```
mgiuca, line 482: Sorry I just ... can't get my head around GitHub's commenting system. It is truly the worst code review system I have ever used (including email). I don't seem to be able to reply to a comment thread above, so starting a new one...

> Looks good, but I still would prefer to call it "within scope of a manifest" rather than "within the navigation scope of a processed manifest" (the "navigation scope" is implied, and the "processed" is explicit later).
>
> And I don't understand why you say "within scope of the processed manifest's scope member" when you could say "within scope of the navigation scope of the processed manifest"; now those are defined as equivalent, so you may as well use the term you defined.

You didn't address this comment. Do you disagree or did you just miss it?

index.html
```diff
@@ -496,21 +493,21 @@ <h2>
...
-        [=Document/URL=] is not <a data-lt="within-scope-manifest">within
-        scope</a> of the <a>application context</a>'s manifest, the user agent
-        SHOULD show a prominent UI element indicating the [=Document/URL=] or
-        at least its <a>origin</a>, including whether it is served over a
-        secure connection. This UI SHOULD differ from any UI used when the
-        [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        If the [=application context=]'s [=active document=]'s [=Document/URL=]
+        is not [=processed manifest/within scope=] of the [=application
+        context=]'s [=processed manifest=], the user agent SHOULD show a
```
mgiuca, line 498: I noticed that you changed all references to this from "of the application context's manifest" to "of the application context's processed manifest". That has two issues:

1. It's quite verbose to always have to say "processed manifest".
2. It implies that an application context has both an unprocessed and processed manifest applied to it, or that either an unprocessed or a processed manifest can be applied. I don't think it should be necessary to make this distinction.

I think it should be possible to say "within scope of the application context's manifest" and know that that unambiguously means the manifest has been processed. If you don't think it means that, perhaps we can change the definition of "application context" or "applied" to explicitly say that it's a processed manifest that's applied.

index.html
mgiuca: Actually, just realised that this changes "navigation scope" to be a property of the application context, not the manifest. I think that's a bit backwards, because then you have this twisted structure:

- The `scope` is a property of the processed manifest.
- The "navigation scope" is a property of the application context (referring to the processed manifest applied to the application context).
- A URL can be "within scope" of a URL.
- A URL can be "within scope" of a manifest (referring to the navigation scope of the application context the manifest is applied to).
- Lots of definitions are concerned with whether a URL is "within scope of an application context's manifest" (referring to the processed manifest applied to the application context).

So if something says "if a URL is within scope of the application context's manifest", expanding out all of these definitions require that we:

1. Start with an application context,
2. Get the manifest applied to it (argument to "within scope of a manifest"),
3. Get the application context it is applied to (argument to "navigation scope"),
4. Get the manifest applied to it (to get the `scope` property),
5. Check whether the `scope` is a prefix of URL.

We should be able to cut out steps 3 and 4. I think this can be achieved by navigation scope continuing to be a property of a manifest, not an application context.

We could also define "navigation scope of an application context" and "within scope of an application context", but I think that's overkill.

marcoscaceres: > applied to it.

I wanted to avoid "it", as "it" is ambiguous. Will try to rephrase... 

> We should be able to cut out steps 3 and 4. I think this can be achieved by navigation scope continuing to be a property of a manifest, not an application context.

Yes, that could work... although an application context is formed by the applied processed manifest.  

> We could also define "navigation scope of an application context" and "within scope of an application context", but I think that's overkill.

Agree.

index.html
marcoscaceres: I addressed it. It's formally defined as:

 * "within navigation scope of a processed manifest".

However, when exported, referenced, and cited, you just do:

> Some |url| is [=processed manifest/within scope=] of a |processed manifest|.

or, for example:
 
> If |url| is [=processed manifest/within scope=] of |manifestX|, then do something...

index.html
marcoscaceres: I'll have another read... I see value in  having the distinction between the (unprocessed) manifest and the processed manifest, which is applied.

marcoscaceres: I'll see if the unprocessed manifest idea can go away entirely... it's just JSON until it becomes a processed thing, so maybe there is a way to make it go away.

index.html
marcoscaceres: Ok, @mgiuca do you agree? For simplicity: 

```
        The <dfn>navigation scope</dfn> is the {{WebAppManifest/scope}} member of a [=processed
        manifest=]. 

        The navigation scope restricts the set of URLs to which an
        [=application context=] can be [=navigated=] while the manifest is
        [=applied=].
```

index.html
marcoscaceres: Ok, so, I'll go with the less fancy version... "within scope of a manifest".

marcoscaceres:
@mgiuca, I think I've addressed all your feedback... hopefully all good to go!

marcoscaceres:
@mgiuca, II'l really like to get this landed this week, as then we can land #880 ... or would you prefer #880 to land first?

index.html
```diff
@@ -460,31 +459,28 @@ <h2>
...
+        |scopePath:string|.
         </li>
       </ol>
       <p>
-        A <a>URL</a> <var>target</var> is said to be <dfn data-lt=
-        "within-scope-manifest">within scope of a manifest</dfn>
-        <var>manifest</var> if <var>target</var> is <a>within scope</a> of the
-        navigation scope of <var>manifest</var>.
+        A [=URL=] |target:URL| is <dfn data-export="">within scope of a
+        manifest</dfn> if the |target| is [=URL/within scope=] of the
```
mgiuca, line 481: Missing `|manifest|` (i.e., this is a function of two parameters: _target_ and _manifest_).

index.html
```diff
@@ -460,31 +459,28 @@ <h2>
...
+        [=navigation scope=] (i.e., [=URL/within scope=] of the
```
mgiuca, line 482: "of the navigation scope of _manifest_ (i.e., within scope of _manifest_'s `scope` member)."

Slight incongruity with the fact that "navigation scope" isn't defined as being "of a manifest", it just is a thing. But you have to say "of the manifest" here, otherwise we aren't defining "within scope of a manifest", just "within scope (of an implied context)".

Probably, we should change the definition of "navigation scope" to be "the navigation scope of a manifest is ...".

index.html
```diff
@@ -496,21 +492,19 @@ <h2>
...
-        If the <a>application context</a>'s <a>active document</a>'s
-        [=Document/URL=] is not <a data-lt="within-scope-manifest">within
-        scope</a> of the <a>application context</a>'s manifest, the user agent
-        SHOULD show a prominent UI element indicating the [=Document/URL=] or
-        at least its <a>origin</a>, including whether it is served over a
-        secure connection. This UI SHOULD differ from any UI used when the
-        [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        If the [=application context=]'s [=active document=]'s [=Document/URL=]
+        is not [=within scope of a manifest=], the user agent SHOULD show a
```
mgiuca, line 496: I see, this suggests the above change to remove the second _manifest_ parameter from "within scope of a manifest" was deliberate. I think that was a mistake.

This now reads as "if the document URL is within scope of **any** manifest, no need to show UI", which isn't right. This should not change; it should still read "If [the URL] is not within scope of the application context's manifest".

Aside: I previously wanted "within scope [of a manifest]" to be usable without having to write "of a manifest" every time, so it acts as an overloaded function with the same name; I can simply write "within scope of _some-url_" or "within scope of _some-manifest_" and the correct one is invoked, based on context. I thought you agreed that would be possible, by defining it as `[=URL/within scope=]` and `[=manifest/within scope=]` so that when rendered, they both just read as "within scope" but they link to the correct overload. Did this change?

index.html
```diff
@@ -549,9 +543,8 @@ <h3>
           Deep links
         </h3>
         <p>
-          A <dfn>deep link</dfn> is a URL that is <a data-lt=
-          "within-scope-manifest">within scope</a> of an <a>installed web
-          application</a>'s manifest.
+          A <dfn>deep link</dfn> is a URL that is [=within scope of a
+          manifest=].
```
mgiuca, line 547: Similarly, this should not change.

index.html
marcoscaceres: Yes, sorry, we can definitely do that. I'll put those back in.

index.html
marcoscaceres: Ok, so this will become:

```
<dfn data-dfn-for="manifest"
         data-lt="navigation scope"
         data-export="">navigation scope of a |manifest|</dfn>
```

And it can be used like this:

```
The [=manifest/navigation scope=] a |manifest| or [=processed manifest=]. 
```

index.html
marcoscaceres: Ok, addressed how to use the term and made sure we are getting the processed manifest from the application context .

marcoscaceres:
Ok, hopefully all good now @mgiuca? 🤞🤞🥺🤞🤞

mgiuca:
Hrm ... pr-preview doesn't seem to have been updated. It's still stuck on 3ed7528. (Come to think of it ... that's very old, it's _plausible_ that my previous round of reviews I was reviewing an older version than what you had presented..... that might explain why I thought you didn't do some of my suggestions. If so, I'm very sorry.)

index.html
```diff
@@ -460,31 +460,28 @@ <h2>
...
+        |scopePath:string|.
         </li>
       </ol>
       <p>
-        A <a>URL</a> <var>target</var> is said to be <dfn data-lt=
-        "within-scope-manifest">within scope of a manifest</dfn>
-        <var>manifest</var> if <var>target</var> is <a>within scope</a> of the
-        navigation scope of <var>manifest</var>.
+        A [=URL=] |target:URL| is <dfn data-dfn-for="manifest" data-lt=
+        "within scope" data-export="">within scope</dfn> of a |manifest| if the
```
mgiuca, line 482: Could `|manifest|` get a type: `|manifest:processed manifest|` so when you mouse-over it, it says "processed manifest"?

"processed manifest" is a sort of informal type though, I'm not sure if this needs to be a formal WebIDL or infra type.

index.html
```diff
@@ -460,31 +460,28 @@ <h2>
...
+        |target| is [=URL/within scope=] of the [=manifest/navigation scope=]
```
mgiuca, line 483: s/the/_manifest_'s

index.html
```diff
@@ -460,31 +460,28 @@ <h2>
...
+        (i.e., [=URL/within scope=] of the {{WebAppManifest/scope}} member).
```
mgiuca, line 484: s/the/_manifest_'s

or

_manifest_.`scope`

marcoscaceres:
Oh man, that's my fault! argh, how embarrassing. Seems I didn't push.

index.html
```diff
@@ -496,21 +493,20 @@ <h2>
...
       <p>
-        If the <a>application context</a>'s <a>active document</a>'s
-        [=Document/URL=] is not <a data-lt="within-scope-manifest">within
-        scope</a> of the <a>application context</a>'s manifest, the user agent
-        SHOULD show a prominent UI element indicating the [=Document/URL=] or
-        at least its <a>origin</a>, including whether it is served over a
-        secure connection. This UI SHOULD differ from any UI used when the
-        [=Document/URL=] is <a>within scope</a>, in order to make it obvious
-        that the user is navigating off scope.
+        If the [=application context=]'s [=active document=]'s [=Document/URL=]
```
mgiuca, line 496: Neither "active document" nor "URL" is linking for me. Can you check these refs?

index.html
```diff
@@ -496,21 +493,20 @@ <h2>
...
+        is not [=manifest/within scope=] of the [=application context=]'s
+        [=processed manifest=], the user agent SHOULD show a prominent UI
+        element indicating the [=Document/URL=] or at least its [=origin=],
+        including whether it is served over a secure connection. This UI SHOULD
+        differ from any UI used when the [=Document/URL=] is [=manifest/within
+        scope=] of the [=application context=]'s [=processed manifest=], in
```
mgiuca, line 502: Echoing a previous comment: feel free to disagree here, it's a very minor nit: I would prefer to not say "processed manifest" everywhere, since it's implied by "within scope" that the manifest be processed.

I think it would read more cleanly to just say "application context's manifest". Similarly down below for deep links, and shortcut's `url` item, you could say "manifest" instead of "processed manifest".

index.html
marcoscaceres: it's nice to see what to expect, so will add.

index.html
marcoscaceres: ReSpec's xref service is down at the moment, so links are currently broken. Should be back up in a few hours.

index.html
marcoscaceres: I think I'm going to stick with "processed manifest" for these ones if that's ok. Even though you are right that it's implied by "within scope", it reminds the reader that it's not the unprocessed one that was downloaded.

marcoscaceres:
> Great! I've dropped a bunch of comments that are mostly nits. Happy with how this ended up.

Me too! Thanks for all your help. It was a long slog, but this ended up 100 times better!

mgiuca:
Great, thanks @marcoscaceres !

PR #899, opened 2020-06-10
Drop WebIDL enums

### LINKED ISSUE #633: Don't use enums, because they throw
We need to drop all the enums, as they throw.

christianliebel:
Closes #633

This change (choose one):

Commit message:

Drop WebIDL enums, because they throw for invalid values.

***
<a href="https://pr-preview.s3.amazonaws.com/christianliebel/manifest/pull/899.html" title="Last updated on Jun 10, 2020, 2:29 PM UTC (caf69c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/899/632df40...christianliebel:caf69c6.html" title="Last updated on Jun 10, 2020, 2:29 PM UTC (caf69c6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1078,12 +1078,6 @@
           The output from inputting an JSON document into this algorithm is a
           <dfn>processed manifest</dfn>.
         </p>
-        <p class="issue">
-          We need to catch throws associated with enumerations in IDL
-          conversion as the spec might gain new values over time not supported
-          by all existing browsers. This is especially important as we rely on
-          enums not defined in this specification.
-        </p>
         <ol>
           <li>Let <var>json</var> be the result of [=parse JSON from bytes=]
           <var>text</var>. If parsing throws an error:
--- a/index.html
+++ b/index.html
@@ -1106,6 +1106,16 @@
           value|converting=] <var>json</var> to a <a>WebAppManifest</a>
           dictionary.
           </li>
+          <li>Set |manifest|["{{dir}}"] to the result of running [=processing
+          the `dir` member=] given |manifest|["{{dir}}"].
+          </li>
+          <li>Set |manifest|["{{display}}"] to the result of running
+          [=processing the `display` member=] given |manifest|["{{display}}"].
+          </li>
+          <li>Set |manifest|["{{orientation}}"] to the result of running
+          [=processing the `orientation` member=] given
+          |manifest|["{{orientation}}"].
+          </li>
           <li>Set <var>manifest</var>["<a>start_url</a>"] to the result of
           running <a>processing the <code>start_url</code> member</a> given
           <var>manifest</var>["<a>start_url</a>"], <var>manifest URL</var>, and
@@ -1155,6 +1165,27 @@
           <li>Return <var>manifest</var>.
           </li>
         </ol>
+        <section>
+          <h4>
+            Processing enumeration members
+          </h4>
+          <p>
+            The steps for <dfn>processing an enumeration member</dfn> are given
+            by the following algorithm. The algorithm takes a [=DOMString=]
+            |value| and a [=set=] of |supported values|. This algorithm returns
+            a `DOMString?`.
+          </p>
+          <ol>
+            <li>Set |value| to [=ascii lowercased=] |value|.
+            </li>
+            <li>If |supported values| [=set/contain|contains=] |value|, then return
+            |value|.
+            </li>
+            <li>Otherwise, [=issue a developer warning=] that |value| is not
+            supported and return `undefined`.
+            </li>
+          </ol>
+        </section>
       </section>
       <section>
         <h3 id="applying">
@@ -1287,9 +1318,6 @@
         <h3>
           <code>dir</code> member
         </h3>
-        <pre class="idl">
-            enum TextDirectionType { "ltr", "rtl", "auto" };
-        </pre>
         <p>
           The <dfn>dir</dfn> member specifies the <dfn>base direction</dfn> for
           the <a>localizable members</a> of the <a>manifest</a>. The <a>dir</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1363,6 +1385,20 @@ <h3>
...
           </li>
         </ol>
+        <p>
+          The steps for <dfn>processing the `dir` member</dfn> is given by the
+          following algorithm. The algorithm takes a [=DOMString=] |value| as
+          an argument. This algorithm returns a `DOMString`.
+        </p>
+        <ol>
+          <li>Let |value| be [=processing an enumeration member=] given |value|
+          and [=TextDirectionType=].
```
mgiuca, line 1395: `TextDirectionType` and `DisplayModeType` are referenced several times throughout the spec, and are no longer formally defined. Your new structure implies that these are now "sets of DOMStrings" (as opposed to WebIDL enum types), so I think they need to be formally declared as such.

marcoscaceres:
Yeah, I'm not on board on this. My plan is to switch to infra types, but once we do the other more important stuff...

christianliebel:
@mgiuca First of all, thanks for your feedback! I’ve seen your changes in #750 after I created this PR, are they related somehow?

> Yeah, I'm not on board on this. My plan is to switch to infra types, but once we do the other more important stuff...

@marcoscaceres Alright, let's talk about this on Monday. 😇

mgiuca:
> @mgiuca First of all, thanks for your feedback! I’ve seen your changes in #750 after I created this PR, are they related somehow?

Yeah, #750 (thanks for finding this) is what I had in mind when I said "define slightly different parsing rules for WebIDL but keep the same representation". As I recall, that was rejected. @annevk said on https://github.com/heycam/webidl/pull/597 that we could use https://infra.spec.whatwg.org/#parse-json-into-infra-values, but I don't see how that addresses the error problem.

What I was trying to capture with that "[CatchTypeError]" annotation was a declarative way of defining the "limit" of errors. Because we don't just want to say "any manifest field that doesn't have the correct type should just be dropped", because for example if an ImageResource has a bad URL, the whole ImageResource should be dropped, not just the URL field. So we need to manually define on a case-by-case basis the "scope" where failure stops.

It looks silly in my patch that every single element has `[CatchTypeError]` on it. But there's a reason to this. For example, look at icons:

```webidl
[CatchTypeError] sequence<[CatchTypeError] ImageResource> icons;
```

That says if any icon is invalid, the failure bubbles up to the whole ImageResource and we drop that item from the list. But also, if `icons` itself is the wrong type (say, an int), then we drop the icons object, rather than erroring out the whole manifest.

I still like this approach. I think it's better than scrapping the whole WebIDL for hand-written algorithms, or another schema language. We have WebIDL, a beautifully defined way to express a typed data structure that everyone working on the web platform can read and write. I'd rather embrace it than abandon it, simply because we don't have a good way of describing how errors should propagate.

annevk:
See https://github.com/whatwg/infra/issues/159#issuecomment-491338205 for the latest on a schema-language. I'm not entirely convinced it's needed as it seems pretty easy to reject/accept things in prose. Perhaps a list of use cases or scenarios would help there.

marcoscaceres:
@annevk, the use case is to take fetch some JSON -> convert into a neutral set of types (e.g., parsed in C++, Rust, JS, whatever) -> process the data into some canonical data structure while performing error handling and assigning defaults/fallback values, then allow the browser to operate on it. 

In my mind, Infra types are ideal for this because they meet the use case of being programming language neutral.  

Just to reiterate the problem: this spec defines things using WebIDL, but no one actually sends the JSON through a WebIDL processor (so the spec doesn't match reality, so would be pointless to add more error handling or pretend this is WebIDL). Chrome processes the JSON using C++, while Gecko does it in JS.

annevk:
@marcoscaceres well, https://infra.spec.whatwg.org/#convert-a-json-derived-javascript-value-to-an-infra-value addresses that. But I think @mgiuca wants something more and I was asking about that.

mgiuca:
Hi @annevk . The Infra [convert a JSON-derived JavaScript value to an Infra value](https://infra.spec.whatwg.org/#convert-a-json-derived-javascript-value-to-an-infra-value) is necessary but not sufficient, as it's dynamically typed. It will take any valid JSON and convert it into an Infra value, but as far as I can tell, has no schema and no way of checking that the input JSON corresponds to a particular "type signature" (i.e., well-defined data structure). I guess that's what you're trying to cover with whatwg/infra#159. In the interum, I have to check all that myself with prose.

> I'm not entirely convinced it's needed as it seems pretty easy to reject/accept things in prose.

It's "easy" to do in small quantities. It makes it harder to read the overall spec, though, especially when you have a JSON structure as large and complex as that of the Manifest format. I'm really happy with being able to look at [this WebIDL](https://www.w3.org/TR/appmanifest/#webappmanifest-dictionary) and tell at a glance what type is required for each of those fields. It would be much less readable if I had to navigate to the parsing algorithm for each field and then inspect the prose text within to find out what assertions are being made about the data in there, to infer what type is being expected for that field.

I'm not wedded to WebIDL. I just would like some declarative schema language as opposed to manual parsing code.

@marcoscaceres yes, Chromium and Firefox both parse the structure with code, but that isn't particularly readable either, and ideally those implementations would be refactored to parse them using a declarative structure. Either way, it should not stop the specification from being more readable than the code that implements it.

> Perhaps a list of use cases or scenarios would help there.

The use case is simply the need to express the type requirements of the Web App Manifest JSON structure: https://www.w3.org/TR/appmanifest/#webappmanifest-dictionary

Currently it is expressed in WebIDL, which clearly communicates the type of everything, but _doesn't_ communicate what should happen in the event of a type error. Ideally we would be able to fix that problem without making the text less readable and maintainable.

marcoscaceres:
Closed via 32b497c4ca34c9069aa5e545126c252724f370fe

PR #926, opened 2020-07-20
Editorial: clarify when/how 'auto' is applied

### LINKED ISSUE #923: auto value of the dir member
https://www.w3.org/TR/appmanifest/#dom-textdirectiontype-auto

AFAIK the `auto` value for direction is often used when the direction is to be determined programmatically using heuristics. See [HTML](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute) for an example.

For specs that don't use heuristics to determine the base direction (like [CSS](https://drafts.csswg.org/css-writing-modes-3/#direction) and [TTML](https://www.w3.org/TR/ttml2/#style-attribute-direction)), they usually don't include an `auto` value. I wonder what intention of the `auto` value is in the Web App Manifest. Is an `auto` value needed?

marcoscaceres:
Closes #923 

Commit message:
Clarified how "auto" works.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/926.html" title="Last updated on Feb 22, 2021, 6:35 AM UTC (debfa1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/926/140001c...debfa1a.html" title="Last updated on Feb 22, 2021, 6:35 AM UTC (debfa1a)">Diff</a>


index.html
```diff
@@ -1120,21 +1120,21 @@ <h3>
...
-            is <a>rtl</a>.
+        <ol data-link-for="TextDirectionType">
+          <li>If the <a>base direction</a> is <a>ltr</a> or <a>rtl</a>, the
+          user agent MUST override <a data-cite="bidi#P3">Rule P3</a> of
+          [[BIDI]], setting the paragraph embedding level to 0 if the <a>base
+          direction</a> is <a>ltr</a>, or 1 if the <a>base direction</a> is <a>
+            rtl</a>.
           </li>
-          <li data-link-for="TextDirectionType">Otherwise, the user agent MUST
+          <li>Otherwise it's <a>auto</a>, in which case the user agent MUST
```
marcoscaceres, line 1133: Why don't we just make this even more clear... if it's "auto" -> just do P1?

xfq:
The text is still not very clear to me. @r12a do you have any suggestion?

marcoscaceres:
@xfq can you help me understand a bit more about what is confusing. I can try to improve it. The current text does require the reader to go and read the [[BIDI]] spec. We are deliberately trying to avoid repeating what is in [[BIDI]] in our spec, as [[BIDI]] is the authoritative source for how the text will be handled.

r12a:
hi @marcoscaceres .   I think there are a couple of things to discuss.

1. I think you may not need to get to the level of specificity you do wrt the points in the bidi algorithm (UBA), especially since this spec iiuc carries metadata rather than actually implementing bidi (correct me if i'm wrong, i haven't reviewed the spec deeply myself - Fuqiao did that).   I think, at the moment, the steps you're pointing to in the bidi spec are not actually the right ones.  Maybe it's enough to say simply that if the value is set to `ltr` or `rtl`, the consumer will need to ensure that the base direction is set to one of those, respectively, when the text is displayed for a user.  For `auto` it may be enough to say that the base direction will be set according to the first strong directional character. Then mention that all this is as described in the UBA.

2. The text you currently have seems to convey the wrong idea about how `auto` works.  It suggested to me that the direction of the string to which it is attached is determined by looking at the direction of the surrounding text (ie. like an inherited value).  That's not actually the case.  The auto value should force the consuming application to check the first strong directional character in the text to which the value is applied, and determine the base direction from that.  So it looks inwards rather than outwards for cues.  

Does that help?

marcoscaceres:
@r12a that helps a lot! Thanks you. Just to be clear, the "base direction" is just the "ltr" or "rtl" or "auto" values, not the direction of the surrounding text. 

So:

```JSON 
{
   "dir": "auto",
   "description": "set my direction"
} 
```

base direction is "auto". 

So, with the updated text, we are now trying to say is: if "rtl" or "ltr" -> Rule P3 is overridden (as BIDI says to do). Otherwise, it's "auto", so do Rule P1.

r12a:
Not quite, but you're closer.  When it comes to deciding how to display the string, the _actual_ base direction can only be LTR or RTL.  The `dir` _property_ indicates how an application should _work out_ the base direction of the string: if it's set explicitly to `ltr` or `rtl`, that's straightforward; if the value is `auto`, the application has to look for the first strong character in the string to figure out what the base direction will be.

But you're right that the base direction we're concerned with isn't the direction of the surrounding text: it's the direction that should be applied to the string when it is in place for display.

Let's look at an example in HTML.  Let's suppose we have a Hebrew string which will be inserted into an English HTML page (where the overall direction of the paragraph is LTR). Here's the string:

Screenshot 2020-07-27 at 10 20 46

Suppose the application creates the following markup, because, although it knows the language of the string, it doesn't know anything about the appropriate base direction for the string itself:

`<p>We found "<span class="result" lang="he">[Hebrew here]</span>".</p>`

The result will be:

Screenshot 2020-07-27 at 10 14 32

The inserted text has inherited the LTR base direction of the surrounding text, because no other information was to hand. The order of the text (the locations of the comma and 'W3C') is incorrect.

However, if the string has accompanying medata to say that it should be displayed RTL, the application can take that into account while inserting the data and produce the following markup:

`<p>We found "<span class="result" lang="he" dir="rtl">[Hebrew here]</span>".</p>`

Now the Hebrew text will be displayed correctly, because the base direction for the inserted string is correctly established.

Screenshot 2020-07-27 at 10 20 05

This is a simplified example, to get across the ideas about the scope of the base direction. The introduction of an `auto` property value and `bdi` or `dir="auto"` markup introduces various other options, which it's not necessary to go into here, other than to say that an `auto` value in metadata indicates that either the application doing the inserting needs to work out the base direction for the inserted text, or (in HTML) the `dir` attribute should be set to `auto` and the HTML parser will need to work out the text direction for the text inserted into the span.

marcoscaceres:
@r12a really appreciate the clarification and example, what I'm struggling with is what exactly I should change in the proposed pull request. When I read it, it's clear to me as an implementer what I need to do there to achieve what you suggest (of course, I'm biased).

xfq:
The i18n WG is working on a centralized definition of ltr/rtl/auto that specs can refer to.

xfq:
> Note also that i changed SHOULD to MUST, because you really do want this to happen – there are no circumstances i can think of where it's not appropriate.

SHOULD enables other possibilities, such as heuristics based on whether there are more RTL characters in the text than LTR (in some cases it's more accurate than first-strong), but this may have an impact on interoperability, so I am not sure if it is worth it.

PR #928, opened 2020-07-20
Editorial: include JSON Schema into spec

marcoscaceres:
Relates to #611 - we add the JSON schema as a non-normative note to describe the structure and types 

Commit message:
Takes a snapshot of the JSON schema and adds it directly into the spec as a non-normative note.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/928.html" title="Last updated on Jul 20, 2020, 8:50 AM UTC (f168e74)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/928/96246fb...f168e74.html" title="Last updated on Jul 20, 2020, 8:50 AM UTC (f168e74)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2969,9 +2969,9 @@
       </h2>
       <p>
         Developers interested in validating <a>manifest</a> documents can find
-        an unofficial <a href="http://json.schemastore.org/web-manifest">JSON
+        an unofficial <a href="https://json.schemastore.org/web-manifest">JSON
         schema for the manifest format</a> at <a href=
-        "http://schemastore.org/json/">schemastore.org</a>. It is licensed
+        "https://schemastore.org/json/">schemastore.org</a>. It is licensed
         under <a href="https://www.apache.org/licenses/LICENSE-2.0.html">Apache
         2.0</a>. It is kindly maintained by <a href=
         "https://github.com/madskristensen">Mads Kristensen</a>. If you find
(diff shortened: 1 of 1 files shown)


mgiuca:
> Relates to #611 - we add the JSON schema as a non-normative note to describe the structure and types

So are you planning to expand on this and make it the actual normative basis for parsing the manifest? Or will it continue to be a non-normative note?

I don't *really* see the point of having this, when we already have a link out to it. It kind-of endorses it more, which might be a bad thing if it routinely gets out of date. It means there's a piece of spec that isn't controlled by this GitHub repo, and can be dynamically changed by a third party (even though it's in non-normative text). I'm weakly opposed to including it but OK with allowing.

marcoscaceres:
> So are you planning to expand on this and make it the actual normative basis for parsing the manifest? Or will it continue to be a non-normative note?

Non-normative. 

> I don't really see the point of having this, 

I wanted to have a quick alternative to the IDL index - as way for seeing the types. 

> when we already have a link out to it. It kind-of endorses it more, which might be a bad thing if it routinely gets out of date. 

I've updated the PULL_REQUEST_TEMPLATE to prevent that from happening. It shouldn't get out of date as it's our responsibility to keep it updated.  

> It means there's a piece of spec that isn't controlled by this GitHub repo, and can be dynamically changed by a third party (even though it's in non-normative text). I'm weakly opposed to including it but OK with allowing.

We have a good working relationship with the schema folks - so we can ask those folks to always check with us first before updating it.

mgiuca:
I see, so this is basically going to be the "readable" copy of the spec, when WebIDL is deleted, and the rest of the spec will go back to algorithms?

I'm not super happy with that approach (I thought when you [mentioned it previously](https://github.com/w3c/manifest/issues/611#issuecomment-529062071), you were going to have a normative description of the data structure which gets processed algorithmically). I'm afraid that the JSON Schema will get subtly out of date, even with a reminder in the CL description (not from forgetting to update it, but from making subtle differences between the algorithm and the JSON schema).

But anyway, I'm OK with including the JSON Schema in this doc.

marcoscaceres:
Let's try it out. If things go bad, we can rip it out again and just link. I'm also mindful of it getting out date, which is why I also put it into `<details>` to lessen its importance.

Issue #856, opened 2020-03-27
More precise control over app window features


mgiuca:
This is an umbrella issue that keeps coming up over and over over the past few years. It's been discussed in small pockets but I don't think there is a dedicated Manifest issue tracking it, so here is one.

## The problem

Broadly speaking, the problem is this: `display` mode and its associated fallback chain does not give developers enough control over the app window, and it's too inflexible to allow us to build new features on top of. This has manifested in a number of separate issues:

1. When we want to add a new feature, that "looks like" a new display mode, it's unclear how to fit it into the fallback chain. This came up for tabbed application mode (#737) in 2018 (still WIP), and now title bar customization (MicrosoftEdge/MSEdgeExplainers#206). By adding new display modes for new features, it's impossible for sites to mix and match these features.
2. Each display mode is poorly specified and inconsistent across implementations (#673). Developers have no guarantee about what controls they're actually going to get. It's nice in theory to give control over window UI to the user agent, but in practice, it means developers will build with the most popular browser in mind and create de facto standards (for example, if popular browsers add a back button to `display: standalone`, sites will assume it exists). Because of this, we've needed (and have not yet succeeded) to provide ways to query whether these controls are being provided by the user agent (#693).
3. The fallback chain is inflexible. Some sites may want `minimal-ui` falling back to a browser tab. But other sites may only have a slight preference for minimal UI, and be happy to fall "up" to `standalone` if the user agent doesn't support minimal.

On the first issue, a very similar discussion happened in #737 and MicrosoftEdge/MSEdgeExplainers#206. Do we add the new display mode and slot it into the linear fallback chain (e.g., between `fullscreen` and `standalone`?) Or do we break the linearity of the fallback chain, so `fullscreen` still falls back to `standalone` whilst `my_new_feature` also falls back to `standalone`? Or do we add the new feature as a separate Boolean member alongside `display`?

Having `fullscreen` fall back to tabbed mode, or even title bar customization mode, is very problematic, since both of those modes are somewhat special and really should require explicit opt-in from developers. The non-linear fallback chain solves this, but it means you can't mix and match (for example, there would be no way to have a tabbed application with a customizable title bar).

[...]

PR #932, opened 2020-07-28
Adding field `display_override` to the manifest

dmurph:
# Status
Since there were no other implementers that could comment on this, it has been moved into the [manifest-incubations](https://github.com/WICG/manifest-incubations) spec location.

# Pull Request

Closes #???
I'm not sure there is one, but it is related to #856 and could close that - but that issue also goes over some other pieces that this change doesn't solve for yet.

* [ ] Safari (link to issue)
* [x]  Chrome (link to issue): https://bugs.chromium.org/p/chromium/issues/detail?id=1092667
* [ ] Firefox (link to issue)
* [x] Edge (public signal): https://bugs.chromium.org/p/chromium/issues/detail?id=1092667 (they are doing the implementation work)

**I have not done this yet - should I do this before this request is filed?**

Commit message:

Adds the "display_override" manifest field.

This field allows the developer to override the "display" field with a custom fallback chain. It is designed to be both future- and backwards-compatible, and is basically just a logic/algorithm change on how a user agent should determine the requested display mode.

This change also creates an explicit algorithm for determining the display mode, and provides a few new examples using "display_override".

See explainer here:
https://github.com/WICG/display-override/blob/master/explainer.md

TAG Review:
https://github.com/w3ctag/design-reviews/issues/530

Webkit position request:
https://lists.webkit.org/pipermail/webkit-dev/2020-October/031435.html

Mozilla position request:
https://github.com/mozilla/standards-positions/issues/447

* chore:
* editorial:
* BREAKING CHANGE:
* **And use none if it's a normative change**

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest/pull/932.html" title="Last updated on Feb 23, 2021, 6:16 PM UTC (a70561e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/932/7e60981...dmurph:a70561e.html" title="Last updated on Feb 23, 2021, 6:16 PM UTC (a70561e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1283,9 +1283,6 @@
             <dfn data-dfn-for="icon purpose">monochrome</dfn>
           </dt>
           <dd>
-            <aside class="issue atrisk" data-number="905"></aside>
-          </dd>
-          <dd>
             A user agent can present this icon where a <a href=
             "#monochrome-icons-and-solid-fills">monochrome icon with a solid
             fill</a> is needed. The color information in the icon is discarded


dmurph:
@mgiuca can you PTAL?

Note - this is definitely a 'draft' in the sense that I've never done a spec pull request here - sorry if there are a bunch of mistakes here! I would love guidance

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+      </p>
+      </p>
       <div class="example">
         <p>
-          For example, SuperSecure Browser (a fictitious browser) only supports
+          Example: SuperSecure Browser (a fictitious browser) only supports
           the <code>minimal-ui</code> and <code>browser</code> display modes,
           but a developer declares that she wants <code>fullscreen</code> in
-          the manifest. In this case, the user agent will first check if it
+          the manifest by using the <a data-link-for="WebAppManifest">display
```
marcoscaceres, line 648:
```suggestion
          the manifest by using the {{WebAppManifest/display}}
```

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+          </a> property. In this case, the user agent will first check if it
```
marcoscaceres, line 649:
```suggestion
           property. In this case, the user agent will first check if it
```

index.html
marcoscaceres: Nit: Same change globally. Avoid using `data-link-for=` as we now support the new syntax. Quick overview of ReSpec shorthands (and docs): https://respec.org/docs/#Shorthands-Guide

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+          <li>
+            Let |considering_display_mode| be the the value specified by the
+            <a data-link-for="WebAppManifest">display</a> property.
+          </li>
+          <li>
+            If the user agent supports the |considering_display_mode|, then
+            that display mode is used (exit).
+          </li>
+          <li>
+            While |considering_display_mode| has <a>fallback display mode</a>,
```
mgiuca, line 635: I'd like to find a way to express this in a normal while loop rather than effectively using a goto.

I think the right approach here is to essentially allow considering display mode to be undefined. Hasty pseudo-code:

1. Let *considering* be the display property.
2. While *considering* is not undefined:
    1. If the user agent supports *considering*, return *considering*.
    2. Set *considering* to the fallback display mode of *considering* (which is undefined if it has none).

Alternatively, you don't really need a condition, since as your note below says, we're guaranteed to terminate with browser. So you could just do:

1. Let *considering* be the display property.
2. While true:*
    1. If the user agent supports *considering*, return *considering*.
    2. Set *considering* to the fallback display mode of *considering*.

\* I'm not sure what the idiomatic way to do a while-true loop is in [Infra](https://infra.spec.whatwg.org/). There doesn't seem to be any special syntax for "Loop until broken" or "Loop forever". (I've been writing Rust code and I quite like that they have a special keyword `loop {}` for this case, to avoid the somewhat ugly `while true {}` in other languages.)

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+            assign that to |considering_display_mode| and go back to step 3.
+          </li>
+        </ol>
+        This algorithm will end with the final <a>fallback display mode</a> of
```
mgiuca, line 639: Put this in a p class="note" so it's non-normative.

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+        <code>browser</code>, which the user agent MUST support.
+      </p>
+      </p>
       <div class="example">
         <p>
-          For example, SuperSecure Browser (a fictitious browser) only supports
+          Example: SuperSecure Browser (a fictitious browser) only supports
```
mgiuca, line 645: Can just drop "Example:"

index.html
```diff
@@ -1374,6 +1449,20 @@ <h3>
...
         </p>
       </section>
+      <section>
+        <h3>
+          <code>display_override</code> member
+        </h3>
+        <p>
+          The <dfn>display_override</dfn> member is a <a>sequence</a> of<a>
+          DisplayModeType</a>. This item represents the developer's preferred
+          fallback chain for display modes. This field overrides the
```
mgiuca, line 1459: Link "fallback chain"

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+          Example: Same browser as above (SuperSecure Browser),
+          but the developer instead specified the following manifest using
+          display_override:
+          <pre class="example json" title="Using display_override">
+            {
+              display_override: ["fullscreen"],
+              display: "browser"
+            }
+          </pre>
+          This declares that she only wants <code>fullscreen</code>, and if
```
mgiuca, line 666: Avoid using gendered pronouns (in general, best to avoid talking about the developer and just the relationship between the site and the user agent, which is what this spec cares about). Just "This declares that the user agent should use fullscreen, and if that is not supported, it should fall back to browser without considering the other modes."

index.html
```diff
@@ -1374,6 +1449,20 @@ <h3>
           developer's preferred <a>display mode</a> for the web application.
         </p>
       </section>
+      <section>
+        <h3>
+          <code>display_override</code> member
```
mgiuca, line 1454: You also need to add an entry for `display_override` in the Web App Manifest WebIDL (search for `dictionary WebAppManifest`).

index.html
```diff
@@ -207,7 +207,27 @@ <h3>
...
+        <pre class="example json" title="Advanced display requirements manifest">
+          {
+            "name": "Recipe Madness",
+            "description": "All of the recipes!",
+            "icons": [{
+              "src": "icon/hd_hi",
+              "sizes": "128x128"
+            }],
+            "start_url": "/index.html",
+            "display_override": ["minimal-ui", "standalone"],
```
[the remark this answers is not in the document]
dmurph: ah, yes you're right

index.html
dmurph: done - let me know if there are other places you see that you would like me to update :)
or if you see links that I did incorrectly

index.html
```diff
@@ -586,17 +606,72 @@ <h2>
...
+      <p>
+        For more advanced usages, the <a data-link-for="WebAppManifest">
+        display_override</a> member can be used to specify a custom fallback
+        order. This member is valid IFF <a data-link-for="WebAppManifest">display</a> is also present.
+      </p>
+      <p>
+        The algorithm for determing the web app's requested display mode is:
+        <ol>
+          <li>[=list/For each=] <var>considering_display_mode</var> of <var>
+            <a data-link-for="WebAppManifest">display_override</a></var>:
```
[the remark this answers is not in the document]
dmurph: Done.

index.html
dmurph: I don't see this anywhere, what would you like me to link to?

dmurph:
Thanks for the reviews!

I'm not exactly sure how to use github pull request tool to respond to comments, hopefully I did that in an acceptable way

Can you please take a look again?

index.html
mgiuca: I think you just deleted the concept of a fallback chain :) (making it an implicit part of the algorithm).

In that case, link "display modes" to that section, so there's some way to quickly get from here to the section with the algorithm and details.

mgiuca:
Hi,

Yeah GitHub review tool is a big mess; I find it very hard to reply to everything and very easy to lose comments.

I'll do a full review tomorrow; for now just responded to your direct question.

index.html
dmurph: removed this sentence, as the algorithm explicitly states this now

dmurph:
Gentle ping on this request - @mgiuca do you mind giving this another review?

index.html
```diff
@@ -207,7 +207,27 @@ <h3>
               "type": "image/jpeg"
             }]
           }
-      </pre>
+        </pre>
+        <p>
+          The following shows a <a>manifest</a> that prefers the <code>
+          minimal-ui</code> {{DisplayModeType}}
```
mgiuca, line 213: s/`{{DisplayModeType}}`/display mode/ (since this section is quite informal and doesn't refer to specific pieces of spec-language).

index.html
```diff
@@ -207,7 +207,27 @@ <h3>
...
+          to <code>standalone</code>, but wants to support both.
```
mgiuca, line 214: s/wants to support both/if `minimal-ui` isn't supported, fall back to `standalone` as opposed to `browser`.

index.html
```diff
@@ -577,30 +597,88 @@ <h2>
...
-        <dfn>fallback display mode</dfn>, which is the <a>display mode</a> that
-        the user agent can try to use if it doesn't support a particular
-        <a>display mode</a>. If the user agent does support a <a>fallback
-        display mode</a>, then it checks to see if it can use that <a>display
-        mode</a>'s <a>fallback display mode</a>. This creates a fallback chain,
-        with the <a>default display mode</a> (<code>browser</code>) being the
-        last item in the chain.
+        For more advanced usages, the {{WebAppManifest/display_override}}
+        member can be used to specify a custom fallback order. This member is
+        valid IFF {{WebAppManifest/display}} is also present.
```
mgiuca, line 602: I don't think we can have this requirement. Since `display` defaults to "browser" (at the IDL level), having it omitted is exactly the same as writing "display: browser". So we can't write a rule that says "if display is "browser" then display_override is valid, but if display is missing, then display_override is invalid." Thus, just delete this sentence.

Corollary: It is actually valid to have display_override without display, as long as you're happy for legacy browsers to use the browser display mode.

`----` I wrote the below earlier but I don't think it's relevant given what I just said above:

"IFF" isn't technical language in this place. We should reword this as a normative requirement for the user agent, e.g. "If `display` is undefined, set `display_override` to undefined." I think that language belongs in the **steps for processing a manifest**, not here (it's part of the algorithm).

Maybe also add a `<p class="note">` here with the human-understandable form, "The `display_override` member is ignored if `display` is not present."

index.html
```diff
@@ -577,30 +597,88 @@ <h2>
...
+      </p>
+      <p>
+        The steps for determing the web app's requested display mode is given
```
mgiuca, line 605: "web app's requested display mode" is kinda the wrong name, given that it's the *resolved* display mode you're talking about. (The "requested" display mode is the first element of display_override.)

How about "the effective display mode" or "the chosen display mode"?

index.html
```diff
@@ -577,30 +597,88 @@ <h2>
...
+                return |candidate_display_mode|.
+              </li>
+            </ol>
+          </li>
+          <li>
+            Let |candidate_display_mode| be the the value specified by the
+            |manifest|.{{WebAppManifest/display}} property.
+          </li>
+          <li>
+            [=iteration/while=] |considering_display_mode| is not {{browser}}
```
mgiuca, line 625: 1. I think you can capitalize the W of while and it will still link properly.
2. You forgot to update `considering_display_mode` to `candidate_display_mode` here.

index.html
```diff
@@ -577,30 +597,88 @@ <h2>
...
+            <ol>
+              <li>
+                If the user agent supports the |candidate_display_mode|, then
```
mgiuca, line 628: I'm thinking about this while loop. I think it's correct but it's still a bit messy to do a loop with a switch in it when essentially you're just doing a linear scan through a list.

How's this (thinking out loud ... this might be messier, feel free to ignore this and keep the while loop).

First, some definitions:

1. The fallback chain for browser is [browser].
2. The fallback chain for minimal-ui is [minimal-ui] + the fallback chain for browser.
3. The fallback chain for standalone is [standalone] + the fallback chain for minimal-ui.
4. The fallback chain for fullscreen is [fullscreen] + the fallback chain for standalone.

(Or you could explicitly spell out the full fallback chain for each mode.)

Then, in the algorithm:

1. First thing you do is go "the display fallback chain is `display_override` + the fallback chain for `display`".
2. For each candidate of display fallback chain, if the user agent supports candidate, return it.

(Effectively, you combine the entire fallback chain into a single list combining both fields, just like you do in the explainer, and scan over it in one general iteration.)

WDYT?

dmurph: Let me know what you think of the changes :)

dmurph:
Thanks for the review! Please take a look again at the changes :)

index.html
```diff
@@ -577,30 +598,88 @@ <h2>
...
+            |manifest|.{{WebAppManifest/display}} property.
+          </li>
+          <li>
+            [=list/For each=] |fallback_mode:DisplayModeType| of the
+            |fallback_chain| of the |candidate_display_mode|:
+            <ol>
+              <li>
+                If the user agent supports the |fallback_mode|, then
+                return |fallback_mode|.
+              </li>
```
mgiuca, line 649: The above suggestions would collapse this algorithm into just two loops: one to do display_override and one to do the fallback chain of display. I think that would look quite nice.

Arguably it's still cleaner to do my original suggestion, which is to do a single loop over "the [concatenation](https://infra.spec.whatwg.org/#string-concatenate) of *manifest*.`display_override` and the fallback chain of *manifest*.`display`".

But I will leave that up to you, since I can see how it's maybe easier to read having two loops, each with a distinct purpose.

index.html
```diff
@@ -577,30 +598,88 @@ <h2>
...
+            </ol>
+          </li>
+        </ol>
```
mgiuca, line 652: Technically, this algorithm is missing a final return at the end of the loop (as in, something that many compilers would complain about). Infra doesn't require that we return something here, but can we add a `<p class="note">` acknowledging it:

"The above loop is guaranteed to return a value before it terminates, due to the fact that `browser` is in every mode's fallback chain, and the requirement that all user agents support the `browser` display mode."

index.html
```diff
@@ -653,9 +720,8 @@ <h2>
           window).
         </dd>
         <dd>
-          The <code>browser</code> <a>display mode</a> doesn't have a
-          <a>fallback display mode</a> (conforming user agents are required to
-          support the <code>browser</code> <a>display mode</a>).
+          Conforming user agents are required to
```
mgiuca, line 723: I think you can just delete this paragraph. It's redundant for the sentence above: "As such, the user agent MUST support the browser display mode."

I think it was redundantly here before just to explain why it's OK to not have a fallback display mode. That's equivalent to the new note that I suggested you add above.

mgiuca:
@marcoscaceres I think this is close to being ready from a technical standpoint. Can we get a commitment to implement it in Firefox?

@hober I'm not sure if you or anyone from Safari has reviewed this proposal yet. (See the [WICG proposal](https://github.com/WICG/display-override/blob/master/explainer.md); [TAG review](https://github.com/w3ctag/design-reviews/issues/530)). This is just a change to the manifest parser, and doesn't directly involve implementation of new features (though it enables a path to adding new display modes). Do you have any objections, and would Safari be able to commit to implementing this parser change?

(Daniel, could you link to the [TAG review](https://github.com/w3ctag/design-reviews/issues/530) from the PR description?)

index.html
```diff
@@ -577,30 +598,88 @@ <h2>
...
+            |manifest|.{{WebAppManifest/display_override}}:
+            <ol>
+              <li>
+                If the user agent supports the |candidate_display_mode|, then
+                return |candidate_display_mode|.
+              </li>
+            </ol>
+          </li>
+          <li>
+            Define the |fallback_chain| for {{browser}} to be [{{browser}}]
```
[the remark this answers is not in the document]
dmurph: I put this above in the text - hopefully it reads ok?

dmurph:
Updated patch with Matt's comments. It looks like there is an error with something unrelated to my change, but hopefully it is still reviewable?

@hober, @marcoscaceres, can you PTAL? This change is required for future display modes, like tabbed & custom window overlay.

dmurph:
as you can see, I'm struggling with rebasing my fork with changes from here. Any advice appreciated lol. It seems to really want to apply those chore changes in my patch 🤷

index.html
```diff
@@ -207,7 +207,28 @@ <h3>
...
+        </pre>
+        <p>
+          The following shows a <a>manifest</a> that prefers the <code>
+          minimal-ui</code> [=display modes|display mode=]
+          to <code>standalone</code>, but if minimal-ui isn't supported, fall
+          back to standalone as opposed to browser.
+        </p>
+        <pre class="example json" title="Advanced display usage manifest">
+          {
+            "name": "Recipe Madness",
```
mgiuca, line 219: I'm not sure how big of a deal this particular one is, but given the recent efforts around inclusive language, it might be best to avoid the term "madness" here and perhaps just go "Recipe Book", or if you want to spice it up, "Recipe Time", "Recipe Bonanza" or "Recipe Zone", etc.

index.html
```diff
@@ -577,30 +598,92 @@ <h2>
...
+            <ol>
+              <li>
+                If the user agent supports the |candidate_display_mode|, then
+                return |candidate_display_mode|.
+              </li>
+            </ol>
+          </li>
+          <li>
+            [=list/For each=] |fallback_mode:DisplayModeType| of the
+            fallback_chain of |manifest|.{{WebAppManifest/display}}:
```
mgiuca, line 642: Remove the underscore in "`fallback_chain`" since this is a prose term defined above without an underscore.

And link to the above definition (see above comment to use `<dfn>`). I think you can use `{{fallback chain}}` for this. Or maybe `[=fallback chain=]`. I usually guess until it works :)

index.html
```diff
@@ -577,30 +598,92 @@ <h2>
...
       <p>
-        Each <a>display mode</a>, except <code>browser</code>, has a
-        <dfn>fallback display mode</dfn>, which is the <a>display mode</a> that
-        the user agent can try to use if it doesn't support a particular
-        <a>display mode</a>. If the user agent does support a <a>fallback
-        display mode</a>, then it checks to see if it can use that <a>display
-        mode</a>'s <a>fallback display mode</a>. This creates a fallback chain,
-        with the <a>default display mode</a> (<code>browser</code>) being the
-        last item in the chain.
+        The fallback chain for:
```
mgiuca, line 601: Use `<dfn>` around "fallback chain" to define it for use below.

index.html
```diff
@@ -1044,6 +1110,7 @@ <h2>
              DOMString iarc_rating_id;
              USVString start_url;
              DisplayModeType display = "browser";
+             sequence&lt;DisplayModeType&gt; display_override;
```
mgiuca, line 1113: I just thought of a problem (which is not a problem right now, but will be as soon as we add a new display mode): we need a way to say for new display modes that they are allowed in `display_override` but not `display`.

We _either_ need to do this in the type system (so it is type-impossible to use, say, `"display": "tabbed"`, but you can use `"display_override": ["tabbed"]`), or have steps in the algorithm to fail if you use a "new" display mode in `display`. My preference is the former.

I think what we should do is have two enums, for now let's just call them "DisplayModeType" and "EnhancedDisplayModeType". The IDL should have:

```webidl
             DisplayModeType display = "browser";
             sequence<DisplayModeType or EnhancedDisplayModeType> display_override;
```

And never put a new value in `DisplayModeType`.

That allows you to use the new ones in `display_override` but not `display`, and we don't need to mention this in the algorithms at all. Does that sound good? We could also rename the old one to `LegacyDisplayModeType` or something, I don't really mind.

If we do that now, `EnhancedDisplayModeType` will be empty, so maybe we just add a note to do that in the future? Let's add a note just under the listing of `DisplayModeType`'s possible values, that no new values will be added to this, and future display modes will be added to a new enum that can only be used with `display_override` but not `display`, for forwards-compatibility reasons.

index.html
```diff
@@ -577,30 +598,92 @@ <h2>
...
+          <li>
+            [=list/For each=] |fallback_mode:DisplayModeType| of the
+            fallback_chain of |manifest|.{{WebAppManifest/display}}:
+            <ol>
+              <li>
+                If the user agent supports the |fallback_mode|, then
+                return |fallback_mode|.
+              </li>
+            </ol>
+          </li>
```
mgiuca, line 649: Add "[Assert](https://infra.spec.whatwg.org/#assert): false", to explicitly state that you should not be able to exit the second loop.

index.html
```diff
@@ -577,30 +598,92 @@ <h2>
...
+        </ol>
+        <p class="note">
+          The above loop is guaranteed to return a value before it terminates,
```
mgiuca, line 652: s/it terminates/the assertion/

mgiuca (approved):
Great work, Daniel! Sorry this has been so long in the making.

I've approved but added some minor comments. One is a type problem we're going to have in the future, but isn't currently a problem, so I've added a long comment but the effect of it is just to add a note about it, for now.

index.html
dmurph: Sound good, I'll add a note about this under the `DisplayModeType`'s possible values, saying:

> No new values will be added to this enum, and instead future display modes will be added to a new enum that can only be used with `display_override` but not `display`, for forwards-compatibility reasons. `display_override` will then be defined as `sequence<DisplayModeType or EnhancedDisplayModeType>`, assuming the new enum is named `EnhancedDisplayModeType`.

mgiuca: Sounds good. Maybe reword it so it reads more easily: "will be added to a new enum, tentatively named `EnhancedDisplayModeType`, ..." and then you can remove the "assuming" qualifier to the last sentence.

index.html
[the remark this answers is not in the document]
aarongustafson: Please resolve :-)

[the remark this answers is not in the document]
dmurph: Thanks for pinging, fixed.

index.html
dmurph: Based on internal implementation experience, it will be much easier to present this as the other option, where the algorithm fails for any 'new' display mode in `display`.  What do you think about that?


I made a note under the algorithm:
```
          In the future, new values added to {{DisplayModeType}} will not be
          added to the fallback chain, and the algorithm should treat the a
          {{WebAppManifest/display}} field set to a new value as {{browser}}.
```

And a note in DisplayModeType:

dmurph:
I just responded to all comments, ready for review :)

mgiuca:
FYI this was discussed at TPAC 2020, minuted here:

https://www.w3.org/2020/10/26-webapps-minutes.html#t07

(I wasn't present.)

Just wanted to address:

> @slightlyoff : to @NotWoods point, supporting a valid value in _either_ location seems like a good "v2" solution

There is a detailed discussion on this [here](https://github.com/dmurph/display-mode/issues/10).

We explicitly didn't want to support non-legacy values in `display`, because that would not be backwards compatible with older browsers. Two options we considered:

- Allow `display` to have the new modes as an option. e.g., `"display": "window_controls_overlay"`. But if you're a browser that doesn't know about "tabbed", you have to interpret this as "browser", which means developers may _accidentally_ create an experience that has a fancy overlay on supported browsers, and literally doesn't show any window at all on browsers that don't support WCO.
- Allow `display` to have _either_ a string _or_ an array, allowing you to specify the fallback chain directly in there. Again, a legacy browser won't know what to do with the array, so it will just fall back to browser.

If we let developers put the new values in either `display` or `display_override`, then they can work around this backwards compatibility problem by always putting the new values in `display_override`, and never in `display`. But we're providing a footgun by letting them put new values in `display`: if you ever use it, you will see the correct behaviour on your shiny new browser, but incorrect behaviour on the older browser you forgot to test in.

That is why we're proposing to only allow the new values in `display_override`, and not allow any new values to be allowed in `display`.

NotWoods:
I'm OK with not allowing new values in `display`. The resolution from TPAC was that implementors would allow `display` to be omitted if `display_override` is present, allowing it to work like a display v2. I'm concerned about how to teach this to a new dev who isn't concerned about the backwards compat, and requiring two separate display fields makes it more confusing.

dmurph:
FYI @marcoscaceres today is the last day for us to bikeshed on the name :)

mgiuca:
> The resolution from TPAC was that implementors would allow `display` to be omitted if `display_override` is present, allowing it to work like a display v2.

I think that wouldn't be good for compat until _all_ browsers are implementing `display_override` (which is why we designed it as an override, not as a v2). Otherwise, a developer tests against, say, Chrome, uses `display_override` without considering browsers that aren't implementing it, and it doesn't behave properly on those other browsers. I think we should make it mandatory now, and we can address it again in a year or so, removing the need to have a `display` which would be backwards-compatible.

This isn't getting a lot of attention, and I'd like to set a deadline to land it in the Manifest spec. If not, I think we should move this into a WICG repo so that we can point to it without having to refer to this PR. @NotWoods are you planning to implement this in Gecko? If so, I think we can take that as a sign of multi-vendor interest, and land this as-is.

If we only have Chromium on board, then we should move it into WICG.

Should we set a deadline of EOD Friday November 20, to get approvals on this PR?

NotWoods:
@marcoscaceres is usually the one who handles Gecko implementation, I'll defer to him.

I'm OK with adjusting the requirements later, especially since installablity criteria is not defined in the spec anyways.

index.html
```diff
@@ -1374,6 +1449,25 @@ <h3>
           developer's preferred <a>display mode</a> for the web application.
         </p>
       </section>
+      <section>
+        <h3>
+          <code>display_override</code> member
```
marcoscaceres, line 1454:
```suggestion
          `display_override` member
```

marcoscaceres:
If this is a multi-os / multi-browser feature, I think it might be ok to take this on. Can someone verify for me that this works across, say, Chrome (OS?) and Edge with Windows integration? For a feature such as this, I think we can probably argue that OS level integration counts enough as an independent implementation... although we should look at if the Edge/Chrome implementations are truly independent. 

@dmurph, please note that I've removed all the WebIDL from the spec, so this needs some editorial cleanup to get it align with the new stuff in the spec.

dmurph:
So this patch is a bit old - I've since put the spec language in here:
https://github.com/WICG/manifest-incubations
and then landed an editorial change to put an extension point in the display algorithm.

and it looks like the references there are now broken too lol. Cost of having an incubations spec I guess.

If we feel like this should live in w3c/manifest, then I'm super happy to move it into here. We haven't received signals from FF or Safari, and IDK how Edge is being treated now that they are based on Chromium.

marcoscaceres:
> We haven't received signals from FF or Safari, and IDK how Edge is being treated now that they are based on Chromium.

Yeah, it would be good to get another engine onboard before we add this.

diekus:
Hola! Saw this discussion and wanted to ping interested parties as I am looking into adding the "window-controls-overlay" into the display-override, but I am a bit confused as to why this is not in the manifest per se? Should we consider again adding display override to the main spec? What are the parameters that are being taken into account for not allowing it in? I'm thinking it might be confusing to have display in but display-override not, even though their functionalities are highly linked. Also, similar to @NotWoods , it might be confusing for new devs (tbf it is confusing for me atm)? 

To followup on @dmurph 's comments, I have checked with FF and Webkit on positions for WCO and have not gotten any signs of life from either, and I am concerned that  for many upcoming features related to installed Web apps on desktop everything will be thrown into the 'incubations' bag?

I kinda get that it is because there is no other implementation(-ish?), but oddly enough, unless I am mistaken it seems that `shortcuts`](https://caniuse.com/?search=shortcut)" is not implemented in other engines either yet *is* in the manifest spec?

PR #949, opened 2021-01-29
Editorial: Created explicit algorithm for choosing display mode, and add extension point for the display_override incubation.

dmurph:

Commit message:

Editorial: Created explicit algorithm for choosing display mode, and add extension point for the display_override incubation.

* editorial:

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest/pull/949.html" title="Last updated on Feb 18, 2021, 2:55 AM UTC (46466c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/949/0f2fd13...dmurph:46466c9.html" title="Last updated on Feb 18, 2021, 2:55 AM UTC (46466c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -200,7 +200,7 @@
         </pre>
         <p>
           The following shows a <a>manifest</a> that prefers the <code>
-          minimal-ui</code> <a data-link-for="DisplayModeType">display mode</a>
+          minimal-ui</code> {{DisplayModeType}}
           to <code>standalone</code>, but wants to support both.
         </p>
         <pre class="example json" title="Advanced display requirements manifest">
@@ -213,7 +213,7 @@
             }],
             "start_url": "/index.html",
             "display_override": ["minimal-ui", "standalone"],
-            "display": "browser",
+            "display": "minimal-ui",
             "theme_color": "yellow",
             "background_color": "red"
           }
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -555,30 +555,65 @@ <h2>
...
       <p>
-        Each <a>display mode</a>, except <code>browser</code>, has a
-        <dfn>fallback display mode</dfn>, which is the <a>display mode</a> that
-        the user agent can try to use if it doesn't support a particular
-        <a>display mode</a>. If the user agent does support a <a>fallback
-        display mode</a>, then it checks to see if it can use that <a>display
-        mode</a>'s <a>fallback display mode</a>. This creates a fallback chain,
-        with the <a>default display mode</a> (<code>browser</code>) being the
-        last item in the chain.
+        The <dfn>fallback chain</dfn> for:
```
mgiuca, line 558: Historical note: The old text here was wrong (or rather, nonsensical):

> If the user agent does support a fallback display mode, then it checks to see if it can use that display mode's fallback display mode.

This should have read "If the user agent does **not** support a **given** display mode, then it checks to see if it can use that display mode's fallback display mode."

This new text will help clarify it greatly. It might be worth adding a note to the commit message that this actually fixes broken text around how to deal with an unsupported display mode, even though we don't expect it to change implementations since we're preserving the intention behind the old broken text.

index.html
```diff
@@ -555,30 +555,65 @@ <h2>
...
+        </li>
+        <li>{{standalone}} is «{{standalone}}, <a data-link-for=
+        "DisplayModeType">minimal-ui</a>, {{browser}}».
+        </li>
+        <li>{{fullscreen}} is «{{fullscreen}}, {{standalone}},
+          <a data-link-for="DisplayModeType">minimal-ui</a>, {{browser}}».
+        </li>
+      </ol>
+      <p>
+        The <dfn>steps for determing the web app's chosen display mode</dfn> is
```
mgiuca, line 575: s/determing/determining

index.html
```diff
@@ -555,30 +555,65 @@ <h2>
...
+        given by the following algorithm. The algorithm takes a [=processed
+        manifest=] |manifest:processed manifest| and returns a [=display
+        mode=].
+      </p>
+      <ol>
+        <li>
+          <a>Extension point</a>: process any proprietary and/or other
+          supported display mode choosers at this point in the algorithm.
```
mgiuca, line 583: s/display mode choosers/display modes

index.html
mgiuca, line 558: I think we should have a sentence defining fallback chain first. This allows other extensions to add their own (rather than the definition of FC literally being this set of 4 literal ones).

"Every `{{`display mode`}}` has a `<dfn>`fallback chain`</dfn>`, which is a list of `{{`display modes`}}`. The `{{`fallback chain`}}` for:"

index.html
```diff
@@ -555,30 +555,65 @@ <h2>
...
-        the user agent can try to use if it doesn't support a particular
-        <a>display mode</a>. If the user agent does support a <a>fallback
-        display mode</a>, then it checks to see if it can use that <a>display
-        mode</a>'s <a>fallback display mode</a>. This creates a fallback chain,
-        with the <a>default display mode</a> (<code>browser</code>) being the
-        last item in the chain.
+        The <dfn>fallback chain</dfn> for:
+      </p>
+      <ol>
+        <li>{{browser}} is «{{browser}}».
```
mgiuca, line 561: Optional: I think I would prefer if the fallback chain didn't include itself, even if it complicates the algorithm. Otherwise, it's theoretically possible for a display mode to not include itself, which would be weird.

(On the other hand, it allows for "virtual" display modes that don't introduce a new mode, but represent a preset ordered list of other modes. It seems unlikely we'd want this however.)

mgiuca (approved):
lgtm with nits

index.html
dmurph: done :)

PR #961, opened 2021-03-04
Remove query & fragment components from scope URL

### LINKED ISSUE #793: Scope should remove the query and fragment at parse time
Since our [within scope](https://www.w3.org/TR/appmanifest/#dfn-within-scope) algorithm ignores the query and fragment part of the scope, I feel like we should strip those out in the [steps for processing the scope](https://www.w3.org/TR/appmanifest/#dfn-processing-the-scope-member).

Thus if your `scope` is `"/foo/?bar"`, we would canonically say that the scope is `"/foo/"`, reflecting the fact that the query string is irrelevant.

This change would not actually have any direct (specified) user-visible changes, since the only place in the manifest spec that scope is used is the "within scope" algorithm (which ignores query and fragment). However, it might be exposed to the user, for example, through developer tools that displays the canonical parse of the manifest, in which case it would be best to strip the query and fragment for clarity.

marcoscaceres:
Closes #793

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [x] [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1697089)
* [ ] Edge (public signal)

Commit message:

When the scope URL is created, the query and fragment components are removed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/961.html" title="Last updated on Apr 7, 2021, 12:26 AM UTC (7170a48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/961/5a3e8ec...7170a48.html" title="Last updated on Apr 7, 2021, 12:26 AM UTC (7170a48)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -569,8 +569,11 @@
           </li>
           <li>If |scope| is failure, return.
           </li>
-          <li>If |manifest|["start_url"] is not [=URL/within scope=] of |scope
-          URL|, return.
+          <li>From |scope|, remove the [=url/query=] and [=url/fragment=]
+          components.
+          </li>
+          <li>If |manifest|["start_url"] is not [=URL/within scope=] of
+          |scope|, return.
           </li>
           <li>Otherwise, set |manifest|["scope"] to |scope|.
           </li>


marcoscaceres:
Sent patch to Gecko https://bugzilla.mozilla.org/show_bug.cgi?id=1697089

marcoscaceres:
Going ahead and merge this also, as it has no observable behavioral change in implementations.

PR #969, opened 2021-04-12
Adding protocolhandlers

diekus:
Closes #863

This change (choose at least one, delete ones that don't apply):

* Breaks existing normative behavior (please add label "breaking")
* [x] Adds new normative recommendations or optional items
* [x] Makes only editorial changes (only changes informative sections, or changes normative sections without changing behavior)
* Is a "chore" (metadata, formatting, fixing warnings, etc).

Implementation commitment (delete if not making normative changes):

* [ ] Safari ([link to issue](https://bugs.webkit.org/show_bug.cgi?id=211401))
* [x] Chrome ([link to issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1019239))
* [ ] Firefox ([link to issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1635272))
* [x] Edge ([public signal](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/URLProtocolHandler/explainer.md))

If change is normative, and it adds or changes a member:

* [ ] [updated JSON schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/web-manifest.json)

Commit message:

Adds "protocol_handlers" member to the manifest spec, referencing the process that navigator.registerProtocolHandler uses.

contributors:
+@fabiorocha

Person merging, please make sure that commits are squashed with one of the following as a commit message prefix:

* chore:
* editorial:
* BREAKING CHANGE:
* And use none if it's a normative change

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/969.html" title="Last updated on Apr 12, 2021, 6:17 PM UTC (d1c606c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/969/8c7fd7b...d1c606c.html" title="Last updated on Apr 12, 2021, 6:17 PM UTC (d1c606c)">Diff</a>

PR #972, opened 2021-04-26
Adding protocol handlers

diekus:
Closes #863 (#969)

This is an update that brings the protocol_handlers member addition up to date with the new structure of the specification (as stated in #969, it has been reworked since the previous effort was based on a very old version of the manifest spec).

cc @fabiorocha @aarongustafson  

* [ ] [updated JSON 

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/972.html" title="Last updated on Aug 2, 2021, 7:09 PM UTC (ee2674a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/972/c34d432...ee2674a.html" title="Last updated on Aug 2, 2021, 7:09 PM UTC (ee2674a)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/972.html" title="Last updated on Apr 12, 2023, 11:54 PM UTC (b8c938a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/972/c103b3b...b8c938a.html" title="Last updated on Apr 12, 2023, 11:54 PM UTC (b8c938a)">Diff</a>


index.html
```diff
@@ -1004,6 +1006,116 @@ <h3>
...
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;
+          <var>protocol_handlers</var> and a <a>URL</a> <var>manifest
+          URL</var>. This algorithm returns a <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;.
+        </p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new Array object
+          created as if by the expression [].
+          </li>
+          <li>For each <var>protocol_handler</var> (<a>ProtocolHandlerItem</a>)
```
kenchris, line 1042: I think "for each" should link to infra

index.html
```diff
@@ -1004,6 +1006,116 @@ <h3>
...
+          in the sequence:
+            <ol>
+              <li>If <var>protocol_handler</var>["protocol"] or
+              <var>protocol_handler</var>["url"] are undefined, return failure
```
kenchris, line 1046: How can you return and still continue the iteration? Who would ever catch that return value?

index.html
diekus: Hi Kenneth, what is infra?

kenchris: https://infra.spec.whatwg.org/#list-iterate

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          The steps for <dfn>processing the <code>protocol_handlers</code>
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;
+          <var>protocol_handlers</var> and a <a>URL</a> <var>manifest
+          URL</var>. This algorithm returns a <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;.
+        </p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new Array object
```
kenchris, line 1039: You don't need to produce JavaScript values, as these are not web exposed. You are just parsing and you can store the values in infra types like done else where. "be a new list" then link to https://infra.spec.whatwg.org/#list

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          created as if by the expression [].
+          </li>
+          <li><a href="https://infra.spec.whatwg.org/#list-iterate">For each</a> <var>protocol_handler</var> (<a>ProtocolHandlerItem</a>)
```
kenchris, line 1042: [=list/iterate=]

Use xref to find the best way to refer https://respec.org/xref/?term=iterate

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+              </li>
+              <li>Set <var>protocol_handler</var>["url"] to the result of [=URL
+              Parser|parsing=] <var>protocol_handler</var>["url"] using
+              <var>manifest URL</var> as the base URL. If the result is
+              failure, [=iteration/continue=].
+              </li>
+              <li>If <var>protocol_handler</var>["url"] is not
+              [=manifest/within scope=] of <var>manifest URL</var>, [=iteration/continue=].
+              </li>
+              <li>If <var>protocol_handler</var>["url"] already exists in <var>
```
kenchris, line 1056: [=list/exist=] (given you will be using list)

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+                processedProtocolHandlers</var>, [=iteration/continue=].
+              </li>
+              <li>
+                <a>Append</a> <var>protocol_handler</var> to
```
kenchris, line 1060: [=list/append=]
--

index.html
```diff
@@ -1755,6 +1864,100 @@ <h2>
...
+        where we use <code>protocol</code> but the same restrictions apply.
+      </p>
+      <section>
+        <h3>
+          `protocol` member
+        </h3>
+        <p>
+          The <dfn>protocol</dfn> member of a <a>ProtocolHandlerItem</a> is a
+          <a>string</a> that represents the protocol to be handled, such as
+          "mailto" or "web+auth".
```
kenchris, line 1904: `mailto`?

kenchris: using backtick

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          communication where one app directly invokes another and passes data
+          via custom protocol links.
+        </p>
+        <p>
+          How protocol handlers are presented, and how many of them are shown
+          to the user, is at the discretion of the user agent and/or operating
+          system.
+        </p>
+        <p>
+          The steps for <dfn>processing the <code>protocol_handlers</code>
```
marcoscaceres, line 1030:
```suggestion
          The steps for <dfn>processing the `protocol_handlers`
```

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          member</dfn> are given by the following algorithm. The algorithm
+          takes a <a data-cite=
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;
+          <var>protocol_handlers</var> and a <a>URL</a> <var>manifest
+          URL</var>. This algorithm returns a <a data-cite=
```
marcoscaceres, line 1035: If you can avoid returning anything, all the better... see how the other processing steps are done.

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          "WEBIDL#sequence-type">sequence</a>&lt;<a>ProtocolHandlerItem</a>&gt;.
+        </p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new Array object
+          created as if by the expression [].
+          </li>
+          <li><a href="https://infra.spec.whatwg.org/#list-iterate">For each</a> <var>protocol_handler</var> (<a>ProtocolHandlerItem</a>)
+          in the sequence:
+            <ol>
+              <li>If <var>protocol_handler</var>["protocol"] or
```
marcoscaceres, line 1045: ProTip... though just a nit: 

```suggestion
              <li>If |protocol_handler|["protocol"] or
```

You can do the same for other `<var>` things

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+              <var>protocol_handler</var>["url"] are undefined, [=iteration/continue=].
```
marcoscaceres, line 1046:
```suggestion
              <var>protocol_handler</var>["url"] is undefined, [=iteration/continue=].
```

index.html
marcoscaceres, line 1060:
```suggestion
                [=List/Append=] |protocol_handler| to
```

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+        </p>
+        <p>
+          A user agent SHOULD ask users for permission before registering a
+          <a>ProtocolHandlerItem</a> <var>protocol_handler</var>s as the
+          default handler for a protocol with the host operating system. A user
+          agent MAY truncate the list of <a>ProtocolHandlerItem</a>
+          <var>protocol_handlers</var> presented in order to remain consistent
+          with the conventions or limitations of the host operating system.
+        </p>
+        <div class="example">
```
marcoscaceres, line 1081:
```suggestion
        <aside class="example">
```

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+                  "url": "/play?songId=%s"
+                },
+                {
+                  "protocol": "store",
+                  "url": "/buy?songId=%s"
+                }
+              ]
+            }
+          </pre>
+        </div>
```
marcoscaceres, line 1114:
```suggestion
        </aside>
```

index.html
```diff
@@ -1755,6 +1864,100 @@ <h2>
         </ol>
       </section>
     </section>
+    <section>
+      <h2>
+        ProtocolHandler items
+      </h2>
+      <p>
+        Each <dfn>ProtocolHandlerItem</dfn> is an [=object=] that represents
```
marcoscaceres, line 1872: We no longer use IDL , having `ProtocolHandlerItem` looks a little odd compared to other similar objects in the spec... maybe: 
 
```suggestion
        Each <dfn>protocol handler description</dfn> is an [=object=] that represents
```

Or something...

index.html
```diff
@@ -1755,6 +1864,100 @@ <h2>
...
+        handler URL, the user agent SHOULD run <a>Handling a protocol
+        launch</a>.
+      </p>
+      <p class="note">
+        [[HTML]]'s <code>registerProtocolHandler</code> method allows web sites
+        to register themselves as possible handlers for particular protocols.
+        What constitutes valid <code>protocol</code> and <code>url</code>
+        values for <a>ProtocolHandlerItem</a>s is defined in that API. Also
+        note that the [[HTML]] API uses <a href=
+        "https://url.spec.whatwg.org/#concept-url-scheme"><code>scheme</code></a>
```
marcoscaceres, line 1894: Please see https://respec.org/xref/?term=scheme for how to cite scheme :)

index.html
```diff
@@ -1004,6 +1006,113 @@ <h3>
...
+          <pre class="example json">
+            {
+              "protocol_handlers": [
+                {
+                  "protocol": "web+music",
+                  "url": "/play?songId=%s"
+                },
+                {
+                  "protocol": "store",
+                  "url": "/buy?songId=%s"
```
marcoscaceres, line 1109: These are not in the [safe list](https://html.spec.whatwg.org/multipage/system-state.html#safelisted-scheme) ... so this won't work.

index.html
```diff
@@ -1755,6 +1864,100 @@ <h2>
...
+        </p>
+      </section>
+      <section>
+        <h3>
+          Privacy and Security considerations
+        </h3>
+        <p>
+          Privacy concerns for custom scheme handling are detailed in the
+          <a href=
+          "https://html.spec.whatwg.org/multipage/system-state.html#security-and-privacy">
```
marcoscaceres, line 1947: Please avoid direct links to HTML spec.... `data-cite=` here.

index.html
```diff
@@ -1755,6 +1864,100 @@ <h2>
...
+          {{NavigatorContentUtils/registerProtocolHandler()}}.
+        </p>
+        <div class="note">
+          <p>
+            The user agent MUST ask for permission when using a protocol
+            handler for the first time. This feature requires user interaction
+            and a script cannot communicate with another application on its
+            own.
+          </p>
+        </div>
```
marcoscaceres, line 1958: Note: can't have a MUST inside a note.... either it's a normative requirement, or it's not.

marcoscaceres (changes requested):
This is a good start, but it needs a lot tighter integration with registerProtocolHandler on the  security side - they should both use the same underlying model, and I'm not getting a sense this does. The URL replacement parts are also quite scary - all this should go through HTML's "normalize protocol handler parameters". The examples already show how this would be open to abuse (e.g., using random protocols shouldn't work - as that open up a significant attack avenue).

index.html
diekus: fixed

index.html
diekus: Hola Kenny. I'm still quite new to this so sorry for the basic question. 😉

I'm trying to change the IDL stuff to references to the infra stuff. I can't seem to find proper examples of how to do so, I kinda get that I need a data-cite, and trying to do this **<a data-cite="#list-iterate">For each</a>** but it is not working. I also tried **<a data-cite="ref-for-list-iterate⑤">For each</a>** trying to get the id's that they have in the infra spec. Can you guide me how to proceed in this case?

kenchris: https://respec.org/xref/?term=for+each will tell you how :-)

Just use `[=list/for each=]`

marcoscaceres: @diekus, pro tip: in theory, if you need `data-cite="spec#thing"`, there is probably something wrong. Always check https://respec.org/xref/ first. 

If the term is not there, it usually means that there is a spec bug (in our spec). In the rare case where something is missing, you then need to ask the spec that you are trying to reference to export the term by adding `data-export=""` or `export` for BikeShed specs.

Hope that makes sense! 

PS: you can always ping us for help!

aarongustafson:
> This is a good start, but it needs a lot tighter integration with registerProtocolHandler on the security side - they should both use the same underlying model, and I'm not getting a sense this does. The URL replacement parts are also quite scary - all this should go through HTML's "normalize protocol handler parameters". 

@marcoscaceres In line 1070, @diekus’ proposal punts to the HTML spec for the process of registering a protocol handler (see "register a protocol handler"), which is where all of the normalization and security stuff takes place. Do you think that needs to be framed differently to make that connection more clear for implementors?

```html
        <p>
          To process the the <var>processedProtocolHandlers</var> the user
          agent SHOULD [=register a protocol handler=] per item defined in the
          [=sequence=].
        </p>
```

> The examples already show how this would be open to abuse (e.g., using random protocols shouldn't work - as that open up a significant attack avenue).

Perhaps I’m misreading your concern here, but the second bullet in Example 6 specifically calls out that attempted abuse (like attempting to associate with "store") would be thrown out. Am I missing something?

```html
            <li>The second protocol handler would be ignored, as the protocol
            provided does not start with "web+" and is not part of the
            safelist.
            </li>
```

index.html
diekus: yes indeed! the example states that it will be ignored as it does not comply with either starting with "Web+" or on the safelist

index.html
diekus: Quick Q, how do I use the data cite to reference a part of a spec? I've tried looking for the html protocol handler security section but I don't quite get it how it works.  https://www.specref.org/?q=html%20protoco9l%20handler%20security

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+        </p>
+      </section>
+      <section>
+        <h3>
+          Privacy and Security considerations
+        </h3>
+        <p>
+          Privacy concerns for custom scheme handling are detailed in the
+          <a data-cite="HTML/system-state#security-and-privacy">
+          Security and privacy section</a> of
```
marcoscaceres, line 1936: Let's avoid the hard link here... The link to registerProtocolHandler() is probably enough. 

```suggestion
          Security and privacy section of
```

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+          {{NavigatorContentUtils/registerProtocolHandler()}}.
+        </p>
+        <p>
+          The user agent MUST ask for permission when using a protocol
+          handler for the first time. This feature requires user interaction
+          and a script cannot communicate with another application on its
```
marcoscaceres, line 1942: This seems odd (not the asking for permission part - but this doesn't feel like it's wired up properly)...  

Using HTML+JS as an example, can you show me what would this look like in practice? If it's to do with "permissions", then I suspect we need some kind of Permissions Policy or Permissions API thing here.

index.html
```diff
@@ -1004,6 +1006,103 @@ <h3>
           conventions or limitations of the host operating system.
         </p>
       </section>
+      <section>
+        <h3>
+          `protocol_handlers` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">protocol_handlers</dfn></code> member is an <a>array</a>
```
marcoscaceres, line 1015: Don't export until someone asks for it. 

```suggestion
          The [=manifest's=] <code><dfn  data-dfn-for="manifest">protocol_handlers</dfn></code> member is an <a>array</a>
```

index.html
```diff
@@ -1004,6 +1006,103 @@ <h3>
...
+            </ol>
+          </li>
+          <li>Return <var>processedProtocolHandlers</var>.
+          </li>
+        </ol>
+        <p>
+          To process the the <var>processedProtocolHandlers</var> the user
+          agent SHOULD [=register a protocol handler=] per item defined in the
+          [=sequence=].
+        </p>
```
marcoscaceres, line 1062:
```suggestion
          <li>
           [=list/For each=]  |processedProtocolHandlers|, [=register a protocol handler=].
         </li>
      </ol>
```

Note that the above doesn't quite match what HTML expects.

index.html
```diff
@@ -1004,6 +1006,103 @@ <h3>
...
+        <h3>
+          `protocol_handlers` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">protocol_handlers</dfn></code> member is an <a>array</a>
+          of <a>protocol handler description</a>s that allows a web application to
+          handle URL protocols.
+        </p>
+        <p class="note">
```
marcoscaceres, line 1019: This sounds like an example... 

```suggestion
        <p class="example">
```

Given that we have the example below, do we need this one? Could we combine them?

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+        </li>
+      </ul>
+      <p>
+        A user agent SHOULD use these values to register the web application as
+        a handler with the operating system. When the user activates a protocol
+        handler URL, the user agent SHOULD run <a>Handling a protocol
+        launch</a>.
+      </p>
+      <p class="note">
+        [[HTML]]'s <code>registerProtocolHandler</code> method allows web sites
```
marcoscaceres, line 1879:
```suggestion
        [[HTML]]'s {{NavigatorContentUtils/registerProtocolHandler()}} allows web sites
```

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+        <li>[=`url`=]
+        </li>
+      </ul>
+      <p>
+        A user agent SHOULD use these values to register the web application as
+        a handler with the operating system. When the user activates a protocol
+        handler URL, the user agent SHOULD run <a>Handling a protocol
+        launch</a>.
+      </p>
+      <p class="note">
```
marcoscaceres, line 1878: Do we need to talk about `.registerProtocolHandler()` at all? The API is not really relevant here?

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+      <ul>
+        <li>[=`protocol`=]
+        </li>
+        <li>[=`url`=]
+        </li>
+      </ul>
+      <p>
+        A user agent SHOULD use these values to register the web application as
+        a handler with the operating system. When the user activates a protocol
+        handler URL, the user agent SHOULD run <a>Handling a protocol
```
marcoscaceres, line 1875:
```suggestion
        handler URL, the user agent SHOULD run <a>handling a protocol
```

index.html
```diff
@@ -1755,6 +1854,96 @@ <h2>
...
+        <p>
+          The <dfn>protocol</dfn> member of a <a>protocol handler description</a> is a
+          <a>string</a> that represents the protocol to be handled, such as
+          `mailto` or `web+auth`.
+        </p>
+        <p>
+          The <a>protocol</a> member of a <a>protocol handler description</a> is
+          equivalent to {{NavigatorContentUtils/registerProtocolHandler()}}'s
+          <a href="https://url.spec.whatwg.org/#concept-url-scheme">scheme</a>
+          argument defined in [[HTML]].
```
marcoscaceres, line 1899:
```suggestion
          |scheme| argument defined in [[HTML]].
```

index.html
```diff
@@ -1004,6 +1006,103 @@ <h3>
...
+          Protocol handlers could, for instance, be used for web app
+          communication where one app directly invokes another and passes data
+          via custom protocol links.
+        </p>
+        <p>
+          How protocol handlers are presented, and how many of them are shown
+          to the user, is at the discretion of the user agent and/or operating
+          system.
+        </p>
+        <p>To <dfn>process the `protocol_handlers` member</dfn>, given [=object=] |json:JSON|, |manifest:ordered map|, and |manifest URL:URL|:</p>
```
marcoscaceres, line 1029: The main processing steps should call into this.

marcoscaceres (changes requested):

index.html
diekus: The process itself is the same. This is the way to handle the protocol launch from the manifest instead of from the `registerProtocolHandler()` API. Furthermore, following this link explains that the "_same steps used to invoke a protocol handler defined in HTML_" are the ones that apply. These steps are the ones defined in [the `.registerProtocolHandler()` method steps](https://html.spec.whatwg.org/multipage/system-state.html#protocol-handler-invocation). Since this is another entry point to the steps and that it is stated that they are the same ones defined in the HTML spec, I don't think it impacts the mentioned API nor that it is relevant here.

index.html
diekus: Hola Marcos, so the way this would work is that these registrations happen upon installation of the Web App. There is a prompt, and it is similar to the one that appears when registration is invoked from the `.registerProtocolHandler()`. In this case, where the registration happens from the manifest, the prompt appears the first time the protocol handler is invoked for a Web app. (It was a requirement from security/privacy folks).

image

index.html
marcoscaceres: Wait, I'm still confused. First, let's link "Safelist" here to HTML, just so that's clear. 

Secondly, "web+music" is not in the safe list of "[safelisted schemes](https://html.spec.whatwg.org/multipage/system-state.html#safelisted-scheme)"? So, it would still be rejected.

index.html
```diff
@@ -1004,6 +1006,98 @@ <h3>
...
+          <ul>
+            <li>The first protocol handler would register to handle "web+music"
+            URLs (e.g.: web+music://#1234). When activated, the user agent
+            would instantiate a new <a>top-level browsing context</a> and
+            navigate to
+            <samp>https://example.com/play?songId=web+music://%231234</samp>.
+            </li>
+            <li>The second protocol handler would be ignored, as the protocol
+            provided does not start with "web+" and is not part of the
+            safelist.
```
marcoscaceres, line 1082:
```suggestion
            [=safelist=].
```

index.html
marcoscaceres: >  In this case, where the registration happens from the manifest, the prompt appears the first time the protocol handler is invoked for a Web app.

Sorry, I'm still having a hard tome following. Can you walk me through the set up end-to-end? 

How do I "invoke a protocol handler"? Do I click on a link on some website (appears to be the case from the HTML spec)? 

I'm also confused about this in the HTML spec:

```JS
navigator.registerProtocolHandler('web+soup', 'soup?url=%s')
```

If I can register protocol handlers to do the same thing in HTML using JS, why do we need them in the manifest again? Does this just duplicate the functionality? Or does this work differently or provide some additional functionality that I'm not getting?

Sorry for the dumb questions. I'm just trying to understand.

marcoscaceres:
@aarongustafson wrote:

> @marcoscaceres In line 1070, @diekus’ proposal punts to the HTML spec for the process of registering a protocol handler (see "register a protocol handler"), which is where all of the normalization and security stuff takes place. Do you think that needs to be framed differently to make that connection more clear for implementors?

Yes.  At at minimum, we need to deal with the errors thrown by "register a protocol handler" (that should be simple enough... "if it throws, continue".). Also, the "register a protocol handler" takes "normalizedScheme and normalizedURLString", which is normalization we don't do. 

These above small things, but I want to make sure the integration is done as expected. Right now, it feels like we are not taking as much care as we should in passing the right things to HTML.

kenchris:
Screenshot_20210525-130936

web+ is safe listed

index.html
diekus: Hola Marcos! So in this case "Web+music" would be accepted since the two accepted values for a protocol handler parameter during it's normalization are that: 

1. it is a value fomr the safelisted schemes you have share
2. it is a string that starts with "web+". This is the way to define 'custom' protocol handlers currently on the web and works with the `navigator.registerProtocolHandler()`.

index.html
diekus: Hola again Marcos! Not sure which dumb questions you are referring to, they all seem important to be clear for me!

The difference between `navigator.registerProtocolHandler()` and having a `protocol_handlers` member in the manifest file is is mostly where/when is the scheme registration and invocation happening. Let me explain a bit further:

- `navigator.registerProtocolHandler()`: once it is called it will immediately try to register the scheme and show the prompt displayed above. This is handy if I want to register a scheme anytime from HTML/JS. Handy AF.
- `protocol_handlers` member in the manifest file: this will register a/several scheme(s) when a Web application is being installed. (register the installed Web app as a possible handler for that scheme). Here a noteworthy difference is that the prompt itself would appear once a user in a website on the browser invokes one of the registered schemes. For example, I installed the BatmanPWA and when I am browsing online I click on a 'web+batman://' link _and then_ the prompt appears in the browser asking if the user wants to use that (registered) PWA. This happens the first time a user invokes a specific scheme. 

So in summary, it is a matter of _where_ the registration happens, as it is more convenient for a user to decide whether they want to use an installed Web app when they run into a link that would require it instead of getting N prompts when trying to install a Wen app (where N = number of valid items in the `protocol_handlers` member in the manifest).

dmurph: There is also the automatic unregistration that happens when a user uninstalls a webapp, which is nice :)

marcoscaceres: > protocol_handlers member in the manifest file: this will register a/several scheme(s) when a Web application is being installed.

Ok, but isn't this in violation of the web's security/permission model? I.e., we only really allow things to be request permission at use time, not _en mass_ at install time. 

Consider, we've previously rejected similar proposals whereby one would do:

```
  features: ["geolocation", "camera", "something else"]
```

This feels very similar.  

> Here a noteworthy difference is that the prompt itself would appear once a user in a website on the browser invokes one of the registered schemes. For example, I installed the BatmanPWA and when I am browsing online I click on a 'web+batman://' link and then the prompt appears in the browser asking if the user wants to use that (registered) PWA. This happens the first time a user invokes a specific scheme. Very convenient indeed.

This is concerning though. If I install BatmanPWA, it could do:

```
   protocol_handlers: [
       /// ...1000 protocols here... MUAHAHAH! 
       /// Appear to handle everything, and show up everywhere!
    ]
```

@dmurph: 
> There is also the automatic unregistration that happens when a user uninstalls a webapp, which is nice :)

Agree, that's nice. But it does presume some tight integration at the OS level with the browser. IIRC (and this was long ago), Firefox couldn't support such things because it could only put a shortcut icon on the home screen, but it had no way of getting notified if a user had deleted the shortcut.

diekus:
> @aarongustafson wrote:
> 
> > @marcoscaceres In line 1070, @diekus’ proposal punts to the HTML spec for the process of registering a protocol handler (see "register a protocol handler"), which is where all of the normalization and security stuff takes place. Do you think that needs to be framed differently to make that connection more clear for implementors?
> 
> Yes. At at minimum, we need to deal with the errors thrown by "register a protocol handler" (that should be simple enough... "if it throws, continue".). Also, the "register a protocol handler" takes "normalizedScheme and normalizedURLString", which is normalization we don't do.
> 
> These above small things, but I want to make sure the integration is done as expected. Right now, it feels like we are not taking as much care as we should in passing the right things to HTML.

I have changed the processing to better align (uses the same normalization for scheme and url parameters as rph())

index.html
diekus: Hola Marcos,

> > protocol_handlers member in the manifest file: this will register a/several scheme(s) when a Web application is being installed.
> 
> Ok, but isn't this in violation of the web's security/permission model? I.e., we only really allow things to be request permission at use time, not _en mass_ at install time.
> 

It is as you are saying, the permission request happens when the user is using the protocol for the first time. We think the consent dialog on this first use is a good compromise between asking for permission for every protocol that an app wants to register and the security concerns of registering the protocols.
 
> Consider, we've previously rejected similar proposals whereby one would do:
> 
> ```
>   features: ["geolocation", "camera", "something else"]
> ```
> 
> This feels very similar.
> 
> > Here a noteworthy difference is that the prompt itself would appear once a user in a website on the browser invokes one of the registered schemes. For example, I installed the BatmanPWA and when I am browsing online I click on a 'web+batman://' link and then the prompt appears in the browser asking if the user wants to use that (registered) PWA. This happens the first time a user invokes a specific scheme. Very convenient indeed.
> 
> This is concerning though. If I install BatmanPWA, it could do:
> 
> ```
>    protocol_handlers: [
>        /// ...1000 protocols here... MUAHAHAH! 
>        /// Appear to handle everything, and show up everywhere!
>     ]
> ```
> 

The list of safe-listed protocols is rather small and they're very likely to have multiple handlers already defined among browser, native and installed web apps, which makes it unlikely that an app can take over everything. Would setting an arbitrary limit to the number of protocols that can be registered help? 

> @dmurph:
> 
> > There is also the automatic unregistration that happens when a user uninstalls a webapp, which is nice :)
> 
> Agree, that's nice. But it does presume some tight integration at the OS level with the browser. IIRC (and this was long ago), Firefox couldn't support such things because it could only put a shortcut icon on the home screen, but it had no way of getting notified if a user had deleted the shortcut.

diekus:
friendly ping @marcoscaceres 👋🏼

marcoscaceres:
Hi @diekus, thanks for continuing to work on this. I've update the [initial comment](https://github.com/w3c/manifest/pull/972#issue-623155222) to include the things that are missing still for us to proceed. In particular, we need backing from a engine and no objections.

index.html
```diff
@@ -1004,6 +1006,95 @@ <h3>
...
+        <p>To <dfn>process the `protocol_handlers` member</dfn>, given [=object=] |json:JSON|, |manifest:ordered map|, and |manifest URL:URL|:</p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new [=list=].
+          </li>
+          <li>[=list/For each=] <var>protocol_handler</var> (<a>protocol handler description</a>):
+            <ol>
+              <li>If |protocol_handler|["protocol"] or
+                <var>protocol_handler</var>["url"] is undefined, [=iteration/continue=].
+              </li>
+              <li>Let (<dfn>normalizedProtocol</dfn>, <dfn>normalizedUrl</dfn>) be the result of running <a data-cite="HTML/system-state#normalize-protocol-handler-parameters">normalize protocol handler parameters</a> with <var>protocol_handler</var>["protocol"], <var>protocol_handler</var>["url"] using <var>manifest URL</var> as the base URL, and [=this=]'s relevant [=environment settings object=]. If the result is failure, [=iteration/continue=].</li>
```
marcoscaceres, line 1037: Nit: we will need to get "normalize protocol handler parameters" exported from HTML to use it.

marcoscaceres, line 1037:
```suggestion
              <li>Let (<dfn>normalizedProtocol</dfn>, <dfn>normalizedUrl</dfn>) be the result of running [=normalize protocol handler parameters=] with <var>protocol_handler</var>["protocol"], <var>protocol_handler</var>["url"] using <var>manifest URL</var> as the base URL, and [=this=]'s relevant [=environment settings object=]. If the result is failure, [=iteration/continue=].</li>
```

marcoscaceres:
Moving to general discussion here... 

> Would setting an arbitrary limit to the number of protocols that can be registered help?

It feels like it would still be potentially overwhelming. 

This is why registration via the existing APIs seems favorable to me. It already supports arbitrary same-origin URL registrations with good error handling? 

So, fundamental question is: what significant value does this add over the existing protocol handler API? Also, if the PWA is not installed, won't developers need to call `.registerProtocolHandler()` anyway to support users who don't want the app on their home screen, but still want protocol handler capabilities?

index.html
diekus: @marcoscaceres just out of curiosity, why is it preferable to have it defined in HTML instead of linking to it with the data-cite? is the data cite incorrect?

diekus: I've gotten the [=normalize protocol handler parameters=] definition exported in HTML. 👍🏼

diekus:
Hola @marcoscaceres. 

> So, fundamental question is: what significant value does this add over the existing protocol handler API? 

We think it makes a lot of sense to allow a developer to register protocol handlers in the manifest. It is where they will also register file handlers, link handlers and share targets, all in a way that ties them up nicely with the application itself. 

Additionally, it is appropriate because they are linked to the app lifecycle, providing deregistration of those schemes once the app is uninstalled (unlike `registerProtocolHandler()`). 

From the UX side, if an application needed to register to handle N protocols upon installing, it's bad UX to prompt the user N times for the permissions. With the protocols installed from the manifest, the user would only get a permission prompt the first time they link on a link. 

Finally, we think it's about the constant evolution in HTML to make things more declarative. With the `protocol_handlers` member in the manifest, you don't need to rely on JS to register schemes for your installed app.

> Also, if the PWA is not installed, won't developers need to call `.registerProtocolHandler()` anyway to support users who don't want the app on their home screen, but still want protocol handler capabilities?

Yes, but both are valid. I may want to have both the "mailclient.com" in the browser and additionally have the "Mail Client PWA" registered as a handler for mailto://. (one through `registerProtocolHandler()` and the other one through the manifest).

marcoscaceres:
> We think it makes a lot of sense to allow a developer to register protocol handlers in the manifest. It is where they will also register file handlers, link handlers and share targets, all in a way that ties them up nicely with the application itself.

This makes sense, particularly for new types (e.g., share target). There's been hesitancy amongst implementers in the past to duplicate functionality already provided by a JS API. Having said that...   

>Additionally, it is appropriate because they are linked to the app lifecycle, providing deregistration of those schemes once the app is uninstalled (unlike registerProtocolHandler()).

That's compelling, indeed - although the site could call `unregisterProtocolHandler()`. I guess this is interesting if protocols are "provisionally registered", but are inert until activated. 

As an aside, we probably want to be clear with what happens when `.unregisterProtocolHandler()` is called, if it's not already. 

> From the UX side, if an application needed to register to handle N protocols upon installing, it's bad UX to prompt the user N times for the permissions. With the protocols installed from the manifest, the user would only get a permission prompt the first time they link on a link.

And this would only be for that particular link type, right? Not a blanket approval for all registered things? 

> Finally, we think it's about the constant evolution in HTML to make things more declarative. With the protocol_handlers member in the manifest, you don't need to rely on JS to register schemes for your installed app.

There are a lot of pros/cons here (personally, I like the imperative APIs over the declarative ones) - but let's keep this outside.

marcoscaceres:
@diekus, could you please run HTM5 tidy on PR? There seem to be a couple of tiny markup issues: 

line 1035 column 13 - Warning: missing </ol> before </li>
line 1052 column 11 - Warning: discarding unexpected </li>
line 1011 column 7 - Warning: missing </section> before <li>
line 1053 column 11 - Warning: inserting implicit <ul>
line 1056 column 7 - Warning: discarding unexpected </ol>
line 1053 column 11 - Warning: missing </ul> before </section>

diekus:
> > We think it makes a lot of sense to allow a developer to register protocol handlers in the manifest. It is where they will also register file handlers, link handlers and share targets, all in a way that ties them up nicely with the application itself.
> 
> This makes sense, particularly for new types (e.g., share target). There's been hesitancy amongst implementers in the past to duplicate functionality already provided by a JS API. Having said that...
> 
> > Additionally, it is appropriate because they are linked to the app lifecycle, providing deregistration of those schemes once the app is uninstalled (unlike registerProtocolHandler()).
> 
> That's compelling, indeed - although the site could call `unregisterProtocolHandler()`. I guess this is interesting if protocols are "provisionally registered", but are inert until activated.
>
> As an aside, we probably want to be clear with what happens when `.unregisterProtocolHandler()` is called, if it's not already.
> 

So registering protocol handlers from the manifest file would register them for the installed Web app. Registering protocol handlers through `registerProtocolHandler()` would register the website through the browser, therefore calling `unregisterProtocolHandler()` would not unregister any handlers from installed Web apps. In a way, the two ways of registering handlers are different because they act on different scopes.

> > From the UX side, if an application needed to register to handle N protocols upon installing, it's bad UX to prompt the user N times for the permissions. With the protocols installed from the manifest, the user would only get a permission prompt the first time they link on a link.
> 
> And this would only be for that particular link type, right? Not a blanket approval for all registered things?
>

 Correct, it would be only for the particular link type that is triggered.  Security wise, every different protocol registration needs to be granted permission on first use of that particular link type.
 
> > Finally, we think it's about the constant evolution in HTML to make things more declarative. With the protocol_handlers member in the manifest, you don't need to rely on JS to register schemes for your installed app.
> 
> There are a lot of pros/cons here (personally, I like the imperative APIs over the declarative ones) - but let's keep this outside.

lauramorinigo:
+1 to this initiative, it can be a good start point for other similar cases like RSS.

index.html
```diff
@@ -1006,6 +1008,109 @@ <h3>
...
+        </p>
+        <p>
+          To <dfn>process the `protocol_handlers` member</dfn>, given
+          [=object=] |json:JSON|, |manifest:ordered map|, and |manifest
+          URL:URL|:
+        </p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new [=list=].
+          </li>
+          <li>[=list/For each=] <var>protocol_handler</var> (<a>protocol
```
mgiuca, line 1039: Be consistent with variable naming. Looks like we're using camelCase, so `protocolHandler` not `protocol_handler`.

Question for @kenchris : did this change recently? I'm looking at the shortcuts algorithm above which uses variables like _processedShortcuts_. I swear we used to use _variable names with spaces in them_, not _camelCaseVariableNames_.

index.html
```diff
@@ -1006,6 +1008,109 @@ <h3>
...
+          handler description</a>):
```
mgiuca, line 1040: "of _json_["protocol_handlers"]"

index.html
```diff
@@ -1006,6 +1008,109 @@ <h3>
...
+          system.
+        </p>
+        <p>
+          To <dfn>process the `protocol_handlers` member</dfn>, given
+          [=object=] |json:JSON|, |manifest:ordered map|, and |manifest
+          URL:URL|:
+        </p>
+        <ol>
+          <li>Let <var>processedProtocolHandlers</var> be a new [=list=].
+          </li>
```
mgiuca, line 1038: Need to add a step: "Set _manifest_["protocol_handlers"] to _processedProtocolHandlers_". Otherwise this result isn't going anywhere.

index.html
```diff
@@ -1006,6 +1008,109 @@ <h3>
...
+          <var>manifest URL</var>, [=iteration/continue=].
+          </li>
+          <li>If [=normalizedUrl=] already exists in
+          <var>processedProtocolHandlers</var>, [=iteration/continue=].
+          </li>
+          <li>[=List/Append=] |protocol_handler| to
+          <var>processedProtocolHandlers</var>.
+          </li>
+        </ol>
+        <ul>
```
mgiuca, line 1065: Formatting: Why is there a `<ul>` here with one `li` in it, and then a bunch of other things, then a `</ul>`? Was it meant to be closed after the `li` but you forgot, and then tidy-html added it way down there?

index.html
```diff
@@ -1006,6 +1008,109 @@ <h3>
...
+          <li>[=list/For each=] |processedProtocolHandlers|, [=register a
```
mgiuca, line 1066: I don't think this can appear here without context (not part of any algorithm). It doesn't say when it should happen, or what the requirement level is (I think it should be a SHOULD; without that it implies MUST).

At a minimum, this should be in an algorithm "To register the protocol handlers for a processed manifest _manifest_", then you can properly index into that manifest in the for loop, e.g. "For each _protocolHandler_ of _manifest_["protocol_handlers"]".

Ideally, this would be called from an algorithm that's invoked at install time, but I don't see one of those. So maybe just add a note to the "Installable web applications" section to say the user agent MAY `[=`register the protocol handlers`=]`".

index.html
```diff
@@ -1762,6 +1870,96 @@ <h2>
...
+          <var>protocol_handler</var> having [=manifest=] <var>manifest</var>
+          is invoked, it goes through the same steps used to [=invoke a
+          protocol handler=] defined in [=HTML=], where the user agent SHOULD
+          navigate to [=url=] and the appropriate browsing context is set to a
+          new top level browsing context.
+        </p>
+      </section>
+      <section>
+        <h3>
+          Privacy and Security considerations
```
mgiuca, line 1949: I'm not sure this should go here. There's a main P&S section at the end of the document that collects all of these (this would be the only section to have it's own P&S subsection, other than Installable Web Apps, and that is probably a mistake too -- note that it is duplicated text from the end of the document).

However, I think P&S should only have non-normative discussion; a MUST requirement shouldn't be in it. So I would move this MUST requirement above to Handling a protocol launch. Then explain why it is so in the P&S section.

index.html
```diff
@@ -1762,6 +1870,96 @@ <h2>
...
+        </h3>
+        <p>
+          Privacy concerns for custom scheme handling are detailed in the
+          Security and privacy section of
+          {{NavigatorContentUtils/registerProtocolHandler()}}.
+        </p>
+        <p>
+          The user agent MUST ask for permission when using a protocol handler
+          for the first time. This feature requires user interaction and a
```
mgiuca, line 1958: I don't understand this second sentence. Are you trying to write a normative requirement that registration requires user interaction? Are you justifying the previous MUST requirement? Please rephrase this.

index.html
```diff
@@ -2403,6 +2601,7 @@ <h4>
           </p>
           <pre class="example js" title=
           "Accessing 'display-mode' media feature with script">
+
```
mgiuca, line 2604: stray blank line

mgiuca (commented):
Some drive-by comments.

index.html
diekus: seems like it indeed. I fixed the list since I hadn't noticed this one! Thanks

index.html
diekus: I have added the SHOULD requirement level for the registration and fixed the loose ul element that made it appear as orphan. I have also added a note to the installable web applications section indicating that a protocol handler may be installed.

christianliebel:
Closing this as stale

PR #977, opened 2021-05-07
First pass as adding a Client Hint for Display Mode

### LINKED ISSUE #954: Allow setting an HTTP request for server side PWA detection
There is really no way to detect PWA on server side? It makes things very complicated, there is a logic for some ajax requests and content parts that are not needed for PWAs, of course we can use window.navigator.standalone or window.matchMedia() to  detect standalone mode, then use CSR to fetch the appropriate content and for ajax requests just add that is_pwa parameter, but it's just too much not needed things for many startups that want to have their PWAs without doing tons of changes to their current code, if  just there was a way to define a special HTTP request header in manifest.json to use on all requests.

For now thinking about a very quick and messy solution that can work, use JS to rewrite all internal links (hrefs) and add query parameter to them. So, first time when app is opened with the start_url that has a special query parameter, we will use it on server side to know that we have a PWA user, then on client side we will rewrite all links to include that parameter as well, but it's a very poor solution, it will be easier to replace links on front end, but what about back-end redirections and ajax responses that contain a link?

It's a really great technology and I enjoying to work with it, but lack of this feature is much frustrating :(

aarongustafson:
Closes #954

* [ ] Safari (link to issue)
* [ ] Chrome (link to issue)
* [ ] Firefox (link to issue)
* [ ] Edge (public signal)

Commit message:

Adds a Client Hint to reveal the Display Mode being used by the user agent.

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/aarongustafson/manifest/pull/977.html" title="Last updated on May 25, 2021, 6:50 PM UTC (4333c13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/977/80807cf...aarongustafson:4333c13.html" title="Last updated on May 25, 2021, 6:50 PM UTC (4333c13)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/aarongustafson/manifest/pull/977.html" title="Last updated on Apr 12, 2023, 11:47 PM UTC (d336923)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/977/c103b3b...aarongustafson:d336923.html" title="Last updated on Apr 12, 2023, 11:47 PM UTC (d336923)">Diff</a>


aarongustafson:
Had a few moments to put this together, based on what I’ve been working on in the App Info doc.

aarongustafson:
Note: Validation is failing due to a current issue with the spec, where link type "manifest" is undefined.

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
     </section>
+    <section>
+      <h2>
+        Client Hints
+      </h2>
+      <p>
+        Developers often need to know details about how their web app is being
+        rendered in order to make content-negotiation decisions. Though this
+        information is accessible via JavaScript, when sent as part of a
+        {{Request}}, servers are empowered to do the content-negotiation
```
marcoscaceres, line 2420: Request "concept" here: 

```suggestion
       [=request=], servers are empowered to do the content-negotiation
```

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
         </section>
       </section>
     </section>
+    <section>
+      <h2>
+        Client Hints
+      </h2>
+      <p>
```
marcoscaceres, line 2416:
```suggestion
      <p class="note">
```

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+        Developers often need to know details about how their web app is being
+        rendered in order to make content-negotiation decisions. Though this
+        information is accessible via JavaScript, when sent as part of a
+        {{Request}}, servers are empowered to do the content-negotiation
+        earlier in the process of setting up the web app, before JavaScript
+        execution takes place. To convey this information, this document
+        provides for the following <a data-cite=
+        "client-hints-infrastructure#client-hint-token">client hint token</a>:
```
marcoscaceres, line 2424:
```suggestion
        provides for the `Sec-CH-Display-Mode` [=client hint token=]:
```

If client hint token is not exported, we might need to ask that sped to do that.

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+      </p>
+      <ul>
+        <li>`Sec-CH-Display-Mode` whose value is a suitable [=display modes
+        value=]
+        </li>
+      </ul>
+      <section>
+        <h3>
+          The 'Sec-CH-Display-Mode' Header Field
```
marcoscaceres, line 2433: Why "Sec-"? I presume "Sec-" stands for "Security", but I'm not sure how display mode relates to security? Maybe it should just be `CH-Display-Mode` or is that not how client hints are specified?

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+        </h3>
+        <p>
+          The <dfn data-export="" data-lt=
+          "display mode client hint">Sec-CH-Display-Mode</dfn> request header
```
marcoscaceres, line 2437: Let's not export it until someone needs it: 

```suggestion
          The <dfn data-local-lt=
          "display mode client hint">Sec-CH-Display-Mode</dfn> request header
```

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+          field gives a server information about the [=display mode=] in which
+          the user agent plans to display the web app. It is a <a data-cite=
+          "rfc8941#section-3">Structured Header</a> whose value MUST be a valid
+          <a data-cite="rfc8941#section-3.3.3">string</a>.
+        </p>
+        <p>
+          The header's ABNF is:
+        </p>
+        <pre>
+          Sec-CH-Display-Mode = sf-string
```
marcoscaceres, line 2447: This should probably be
```suggestion
          Sec-CH-Display-Mode = "fullscreen" / "standalone" / "minimal-ui" / "browser"
```

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+        <p>
+          The <dfn data-export="" data-lt=
+          "display mode client hint">Sec-CH-Display-Mode</dfn> request header
+          field gives a server information about the [=display mode=] in which
+          the user agent plans to display the web app. It is a <a data-cite=
+          "rfc8941#section-3">Structured Header</a> whose value MUST be a valid
+          <a data-cite="rfc8941#section-3.3.3">string</a>.
+        </p>
+        <p>
+          The header's ABNF is:
```
marcoscaceres, line 2444:
```suggestion
          The header's <abbr title="Augmented BNF for Syntax">ABNF</abbr>, as per [[rfc5234]], is:
```

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
...
+        </p>
+        <pre>
```
marcoscaceres, line 2446:
```suggestion
        <pre class="abnf">
```

marcoscaceres (commented):
Thanks for crafting this @aarongustafson! Looking pretty good.  

I'm wondering if we need to say exactly when this hint is sent (I imagine that must be part of client-hints-infrastructure spec)? Like, how does the user agent know to send it on page load or whatever... 

Once we clean this up a little bit, might be great to have @yoavweiss take a look at it and make sure we have all the right machinery in place.

index.html
tomayac: `Sec-CH-*` is the common prefix for such hints sent by the user agent over secure connections (display mode being `standalone` depends on the PWA to be installed). Also see [User Preference Media Features Client Hints Headers](https://tomayac.github.io/user-preference-media-features-headers/) that we're specifying at the moment.

tomayac:
> I'm wondering if we need to say exactly when this hint is sent (I imagine that must be part of client-hints-infrastructure spec)? Like, how does the user agent know to send it on page load or whatever...

No need. This is specified in [client hints](https://datatracker.ietf.org/doc/rfc8942/). If it needs to be sent upon the first load, it's a [critical client hint](https://tools.ietf.org/html/draft-davidben-http-client-hint-reliability-02). Any client hint can be promoted to be a critical client by mentioning it in the `critical-ch` header.

> Once we clean this up a little bit, might be great to have @yoavweiss take a look at it and make sure we have all the right machinery in place.

(I'm working with Yoav on [User Preference Media Features Client Hints Headers](https://tomayac.github.io/user-preference-media-features-headers/).)

marcoscaceres:
IIUC, we have a concerns from Mozilla (via @annevk) about Client Hints (https://github.com/w3c/manifest-app-info/pull/32#issuecomment-826733759):

> We're not convinced Client Hints is worthwhile and are concerned about its implications on the same-origin policy.

I'd be interested to hear from Apple, maybe @hober or @rniwa if they have an objection or concerns?

index.html
aarongustafson: I’m apprehensive about locking this down to specific values, especially as the display mode may be one of the overrides not defined in this list.

index.html
```diff
@@ -2409,6 +2409,45 @@ <h4>
         </section>
       </section>
     </section>
+    <section>
+      <h2>
+        Client Hints
+      </h2>
```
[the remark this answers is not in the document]
aarongustafson: I figured there might be more in the future, but that’s fine.

aarongustafson:
> IIUC, we have a concerns from Mozilla (via @annevk) about Client Hints ([w3c/manifest-app-info#32 (comment)](https://github.com/w3c/manifest-app-info/pull/32#issuecomment-826733759))

Yes, the concerns are around potential bloat in the request header. That’s not something we can really address in this context (and is being addressed in the Client Hints discussions). As there is no other way to inform the server about the display mode _without_ involving headers, I don’t know that there is another way to solve for this use case.

aarongustafson:
Note to self: Will need to update this PR to reference the `display_override` calculation if #932 lands first.

index.html
aarongustafson: Client Hint Token is not exported, requesting…

aarongustafson: Actually, it is exported: https://respec.org/xref/?term=client+hints+token

I will update to reference the spec

index.html
marcoscaceres: ah, well we should just expand the list. We can update the spec at any time, so there is no reason not to keep this list in sync.

dmurph: I think we would also want to have this possibly be values that end up in manifest-incubations spec - not sure the best way to do this spec-wise, but referencing display-mode is probably a good way.

dmurph (commented):
This change looks good, but as @marcoscaceres referenced - we don't have signals from other vendors. If we need a place for this spec to live in the meantime we can put it in manifest-incubations

aarongustafson:
> in the meantime we can put it in manifest-incubations

https://github.com/WICG/manifest-incubations/blob/gh-pages/display_mode-client-hint.md

marcoscaceres:
I'm going ahead and closing this, as we haven't seen any implementation interest for a few years.

PR #978, opened 2021-06-01
chore: add tidy action

marcoscaceres:
CC @xfq and @siusin .... might be useful for other specs. 

WDYT?


kenchris (approved):
Not an expert on this, but lgtm from me (rubberstamp)

marcoscaceres:
Thanks for checking it @kenchris!

PR #988, opened 2021-07-15
Add id member to manifest

### LINKED ISSUE #586: Add a unique identifier for a PWA
If one is building a PWA Directory or App Store or search engine that detects PWAs one needs a way to uniquely identify a PWA from just the manifest. 

Currently the spec doesn't explicitly say what that identifier or tuple of identifiers should be which leads to issues like: https://github.com/GoogleChrome/gulliver/issues/323

philloooo:
Closes #586 

Explainer https://github.com/philloooo/pwa-unique-id/blob/main/explainer.md

This change:

* [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=230596) 
  * https://lists.webkit.org/pipermail/webkit-dev/2021-August/031949.html
* [x] Chromium (https://crbug.com/1182363)
  * Chrome Status entry https://www.chromestatus.com/feature/6064014410907648
* [ ] [Gecko](https://bugzilla.mozilla.org/show_bug.cgi?id=1731940)
  * https://github.com/mozilla/standards-positions/issues/493
* [x] KaiOS (https://github.com/kaiostech/gecko-b2g/issues/57) 

* [x] [updated JSON schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/web-manifest.json)

Commit message:

Explicitly defines what uniquely identifies a web application and allow developers to specify the identifier in the manifest.

***

### :boom: Error: connect ETIMEDOUT 128.30.52.89:443 :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 5, 2021, 11:30 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fphilloooo%2Fmanifest%2Ff0817f0d9b2eacef261920ae3feae63c658039de%2Findex.html%3FisPreview%3Dtrue)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/manifest%23988.)._
</details>


dmurph:
Nice - I can also see this fleshing out the section "Updating the manifest" as well with the algorithm

philloooo:
> Nice - I can also see this fleshing out the section "Updating the manifest" as well with the algorithm

Will do, thanks for reminding!

index.html
```diff
@@ -1209,6 +1262,13 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents MAY re-fetch and update a manifest for an installed web
+            application and apply updated manifest to its <a>application
+             context</a>. When a manifest is processed, user agents can
+            determine whether it's an installed application that can be updated
+            or a new application by comparing application's [=manifest/id=].
```
philloooo, line 1270: I tried to explain how the id is used for manifest update. I am not sure if I should elaborate on current chrome manifest update lifecycle though as it's specific to chrome. WDTY? @dmurph @marcoscaceres

philloooo:
@mgiuca can you take a look at this?

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+          <li>Otherwise, set |manifest|["id"] to |start URL|.
+          </li>
+        </ol>
+        <aside class="example">
+          <p>
+            For example, if the |start URL| is
+            <samp>https://example.com/resources/start</samp>,
+            and the value of [=manifest/id=] is <samp>"1"</samp>, then the
+            processed result is <samp>https://example.com/1</samp>. If the
+            value of [=manifest/id=] is <samp>""</samp>, then the processed
```
mgiuca, line 853: "is "" **or "/"**"

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+            result is <samp>https://example.com/</samp>. If [=manifest/id=]
+            not specified, then the processed result is
+            <samp>https://example.com/start</samp>.
```
mgiuca, line 856: This should be `https://example.com/resources/start`.

index.html
mgiuca: Yeah I think leaving it intentionally vague is good. We don't want to impose any details on _how_ the update should work.

index.html
```diff
@@ -1209,6 +1262,13 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents MAY re-fetch and update a manifest for an installed web
```
mgiuca, line 1266: "MAY **, at any time,**"

Just to be clear that there is no specific circumstance that should trigger the update. It's allowed to happen whenever the UA wants.

index.html
```diff
@@ -1209,6 +1262,13 @@ <h3>
...
+            application and apply updated manifest to its <a>application
```
mgiuca, line 1267: This is a bit weird, mechanically. An "application context" is a transient thing (it corresponds to a browser window). An installed web application can have zero or more application contexts at any time, so there is no "its application context".

Also, "application context" doesn't technically belong to an "installed web application"; it is a browsing context that has a manifest applied to it. But the manifest is the thing that's being replaced. So let's see if we can get away with just assuming application contexts belong to installed web applications? ¯\\_(ツ)_/¯

"and apply the updated manifest to any current and future _application contexts_ associated with the application."

index.html
```diff
@@ -1209,6 +1262,13 @@ <h3>
...
+             context</a>. When a manifest is processed, user agents can
```
mgiuca, line 1268: Comparing it to what?

Also: "can determine" is non-committal as to what level of requirement this is for user agents. In general, avoid using the word "can" since it isn't an [RFC2119 term](https://datatracker.ietf.org/doc/html/rfc2119) and therefore has no formal meaning. Use "MAY", "MUST", etc.

I think MUST is appropriate here, since the whole point of id is that we don't want to update an app with a manifest with a different ID, even if the manifest has the same URL as before. However, that would be the only MUST that we're introducing in this PR, which may mean we're requiring UAs to change their existing behaviour. So let's start with a MUST and then run it by other browser vendors.

Perhaps rephrase as: "When the user agent fetches a manifest as part of an update check, it MUST compare the fetched processed manifest's `id` with the _identity_ of the app being updated. If it is not equal, the user agent MUST NOT update the installed application's manifest."

If necessary, we could weaken that to a "SHOULD" and "SHOULD NOT", which would mean existing browsers are still compliant without having to change their behaviour.

mgiuca (changes requested):
Very good, overall. Everything's in the right place and largely conveys the meaning I think you intended. So don't be alarmed by the high volume of comments. These are details :)

Now I've made a lot of specific phrasing suggestions. Feel free to reword them. They are by no means perfect or canonical, just the way that I would personally phrase things. Ask if you have questions.

Thanks, it's exciting to see this finally coming together!

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+          to uniquely identify and reference web apps
+        </p>
+        <p>
+          To <dfn>process the `id` member</dfn>, given [=object=]
+          |json:JSON|, [=ordered map=] |manifest:ordered map|.:
+        </p>
+        <ol class="algorithm">
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=]
```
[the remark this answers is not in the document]
philloooo: Yeah I was struggling with which approach to take in the beginning. But your suggestion does make sense. I will update the implementation too then

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+        <h3>
+          `id` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">id</dfn></code> member is a <a>string</a> that
+          represents the <dfn>id</dfn> for the application.
+        </p>
+        <p>The |id| is used  by user agents uniquely identify the application
+          globally. When |id| is changed, it's recognized as a new app and
```
marcoscaceres, line 824: Globally might be confusion (as it might be confused with "global scope"), "universally" maybe? Assuming there cannot be collisions because another app in the same origin will effectively claim the "id".

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+          can be installed separately. When id is not changed, it can be used
```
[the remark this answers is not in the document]
marcoscaceres: In @mgiuca's suggestion above, maybe s/sees/processes ?

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+          to find installed application to perform manifest update.
+        </p>
+        <p>
+          The |id| can also be used by third party entities such as PWA stores
```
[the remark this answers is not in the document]
marcoscaceres: I would prefer we don't mention "such as PWA stores". Trying to focus exclusively on how browsers use this, rather than how stores use this.

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
+          to uniquely identify and reference web apps
+        </p>
+        <p>
+          To <dfn>process the `id` member</dfn>, given [=object=]
+          |json:JSON|, [=ordered map=] |manifest:ordered map|.:
```
marcoscaceres, line 834:
```suggestion
          |json:JSON|, [=ordered map=] |manifest:ordered map|:
```

index.html
marcoscaceres: I guess the only thing to add is maybe what to do if pre-processing and update fails. I guess nothing happens (i.e., just keep using the currently installed manifest) - but it would might be nice to just say that.

marcoscaceres (commented):
Looking really good @philloooo! Excited to implement this in Gecko. I'll make a start once you make the next round of updates.

index.html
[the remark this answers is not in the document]
aarongustafson: "implementors" is a safe general term.

index.html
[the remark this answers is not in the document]
aarongustafson: FWIW, I like the idea of a table of examples showing how the items would be combined either with or without a defined `id`/`identity`.

aarongustafson (approved):
Just some minor suggestions.

index.html
```diff
@@ -808,6 +811,53 @@ <h3>
...
         </section>
       </section>
+      <section>
+        <h3>
+          `id` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">id</dfn></code> member is a <a>string</a> that
+          represents the <dfn>id</dfn> for the application.
```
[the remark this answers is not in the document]
philloooo: I changed it to 

`The [=manifest's=] <code><dfn data-export="" data-dfn-for="manifest" id="identity">id</dfn></code> member`
 
Hope that's the right way to do it?

index.html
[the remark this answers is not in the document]
philloooo: That sounds right to me.

index.html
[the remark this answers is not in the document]
philloooo: sounds great to me, thanks for the suggestions!

index.html
philloooo: thanks for the suggestion!

index.html
philloooo: @marcoscaceres should this be SHOULD behavior for the case you mentioned?

philloooo:
thanks for the review!  @aarongustafson @marcoscaceres @mgiuca

mgiuca:
Apologies, I haven't got around to re-reviewing this. Looks like Aaron and Marcos are happy.

I'd like to take another look at it, but if I don't get to it by ~Monday, then you can probably go ahead and merge it.

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
...
+          The |identity| takes the form of a URL, which is same origin as the
+          start URL.
+        </p>
+        <p>The |identity| is used by user agents to uniquely identify the
+        application universally. When the user agent sees a manifest with an
+        |identity| that does not correspond to an already-installed application,
+        it SHOULD treat that manifest as a description of a distinct
+        application, even if it is served from the same URL as that of another
+        application. When the user agent sees a manifest with an |identity|
+        matching that of an already-installed application, it SHOULD be used as
```
dmurph, line 831: We should create a section defining how to tell if they are matching. Because we are parsing as a url, certain things should be considered the same:

* ASCII characters: `j` and `%6a`
* Special: `a😀b` and `a%F0%9F%98%80b`, `a b` and `a%20b`
* Slash? `a/b` would become `a%2Fb`?

I'm not sure the best place to put this. We could have a section like

```
<p>
  Two manifests are considered <dfn  id="matching">matching</dfn> if performing the [=url equivalence=] process on the |identity| members of the manifests returns true.
</p>
```

Perhaps put this above the existing paragraph here, and then you can link |matching|.

dmurph: @marcoscaceres I wasn't able to find the url-equivalence definition in the respec search tool, but the definition is here:
https://url.spec.whatwg.org/#url-equivalence

Not sure how to link to that?

index.html
```diff
@@ -1210,6 +1300,16 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents MAY, at any time, re-fetch and update a manifest for an
+            installed web application and apply updated manifest to any current
+            and future <a>application contexts</a> associated with the
+            application. When the user agent fetches a manifest as part of an
+            update check, it MUST compare the fetched processed manifest's
```
dmurph, line 1308: |compare| should also link to the matching section.

index.html
marcoscaceres: Sorry for the delay! we need to ask the editor of that spec to "`<dfn export>`" it. 

I would suggest sending a PR to the URL spec, exporting it and pointing to this pull request explaining why we need it.

index.html
marcoscaceres: yes, SHOULD would be good.

marcoscaceres:
@philloooo, I hope to implement this next week. I may have more feedback then! Thanks again for the great work.

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
...
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
+          with |manifest|["start_ur"]'s origin as the base URL.
+          </li>
+          <li>If |id| is not same origin as |manifest|["start_url"], return.
+          </li>
+          <li>If |id| is failure, return.
```
marcoscaceres, line 854: hmm... the algorithm currently doesn't quite do as intended... here is the JS equivalent of what's currently specified:

```JS
/**
 * @param {Object} json
 * @param {Map<string, any>} manifest
 */
function processIdMember(json, manifest){
  // Set |manifest|["id"] to |manifest|["start_url"].
  manifest.set("id", manifest.get("start_url"));

  // If the type of |json|["id"] is not [=string=], return.
  if (typeof json.id !== "string") return;

  // Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
  // with |manifest|["start_ur"]'s origin as the base URL.
  const id = new URL(json.id, manifest.get("start_url"));

  // If |id| is not same origin as |manifest|["start_url"], return.
  if(id.origin !== manifest.get("start_url").origin)) return;

  // If |id| is failure, return.
}
```

I think what we want is:

```JS
/**
 * @param {Object} json
 * @param {Map<string, any>} manifest
 */
 function processIdMember(json, manifest) {
  const startURL = manifest.get("start_url");
  const fallbackId = new URL(startURL);
  let id;
  if (typeof json.id === "string") {
    try {
      id = new URL(json.id, startURL);
    } catch {
      // It failed, so use the fallbackId
      id = fallbackId;
    }
  }

  if (id.origin !== startURL.origin) {
    // user agent would warn here...
    id = fallbackId;
  }
  // finally, we have the id...
  manifest.set("id", id);
}
```

marcoscaceres (changes requested):
Started to implement but hit some issues. 

@philloooo, made some suggested changes to the algorithm. 

I didn't get a chance to feed the table values into it, but they could serve as unit tests for it.

index.html
```diff
@@ -1210,6 +1300,16 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents MAY, at any time, re-fetch and update a manifest for an
```
marcoscaceres, line 1304: Thinking about this some more, the manifest should be re-fetched on every page load, like any other linked resource. If changes are detected, then the manifest should be re-applied and significant changes flagged to the end-user.

We need to better handle what happens when the ID changes. Also, we need to handle what happens if a manifest is in use, but the new manifest is parsed and it fails.

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
           </p>
         </section>
       </section>
+      <section>
+        <h3>
+          `id` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest" id="identity">id</dfn></code> member is a <a>string</a> that
```
marcoscaceres, line 820: I wonder if we should rename this "app_id" instead? I think "id" is overly generic and app_id makes it more clear that this is for the whole "app", not the id of the manifest.

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
...
+        application. When the user agent sees a manifest with an |identity|
+        matching that of an already-installed application, it SHOULD be used as
+        a signal that this manifest is a replacement for the already-installed
+        application's manifest, and not a distinct application, even if it is
+        served from a different URL than the one seen previously.
+        </p>
+        <p>
+          The |identity| MAY be used by a service that collects lists of web
+          applications to uniquely identify applications.
+        </p>
```
marcoscaceres, line 839: These kinds of services are not a conformance class for this specification, so this should be a note.  

```suggestion
        <p class="note">
          The |identity| can be used by a service that collects lists of web
          applications to uniquely identify applications.
        </p>
```

index.html
marcoscaceres, line 854:
```suggestion
          <li>
          Let |fallbackId:URL| be the result of [=URL Parser|parsing=] |manifest|["start_URL"].
          </li>
          <li>
          Set |manifest|["start_url"] to |fallbackId|. 
          </li>
          <li>
          If the type of |json|["id"] is not [=string=], terminate this algorithm.    
          </li>
          <li>
          Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
          with |manifest|["start_url"]'s origin as the base URL.
          </li>
          <li>
             If |id| is failure, terminate this algorithm.
          </li>
          <li>
             If |id| is not same origin as |manifest|["start_url"], terminate this algorithm.
          </li>
          <li>
            Set |manifest|["start_url"] to |id|. 
          </li>
```

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
...
+        <p>
+          To <dfn>process the `id` member</dfn>, given [=object=]
+          |json:JSON|, [=ordered map=] |manifest:ordered map|:
+        </p>
+        <ol class="algorithm">
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
```
marcoscaceres, line 845: I think we might have another issue here: `start_url` might be `undefined`. 

We would need to assure that `start_url` is always assured for this to work. In that case, the document's URL would again need to be the `start_url`.

marcoscaceres:
For start_url, I think we need: https://github.com/w3c/manifest/pull/670

index.html
espinr: In the case of miniapps, all the vendors agreed to use `app_id` as [identifier for the app](https://w3c.github.io/miniapp-manifest/#dfn-app_id).

index.html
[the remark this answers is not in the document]
dmurph: These are great insights, and we are either currently doing this, or plan on doing this, in Chromium.

1. I have a doc describing when exactly Chromium applies the manifest update here [1]. Basically - we wait until there are no app windows open. (edge case - if you load the page and there are no app windows open, then we update immediately). The only caveat is that we have a 1 day (in-memory) throttle, so a manifest cannot be considered for update more than once a day.
2. We currently (generally) do not allow name and icon updates, and we have a project that is currently being worked on the prompt the user for these updates. The user can reject name & icon updates. However, we do automatically update other members of the manifest, like theme color, etc. It might be good to provide language here that a user agent is not required to apply any of the updated manifest fields, which given user agents flexibility for privacy & security concerns.

[1] https://docs.google.com/document/u/1/d/1q5kmxNU7i4eem22LouMaIJ6123jOw5_zuxVQW1wfW8Y/edit

index.html
dmurph: Maybe more like:
```js
/**
 * @param {Object} json
 * @param {Map<string, any>} manifest
 */
 function processIdMember(json, manifest) {
  const startURL = manifest.get("start_url");
  const fallbackId = new URL(startURL);
  if (typeof json.id !== "string") {
    manifest.set("id", fallbackId);
    return;
  }

  // Maye we need to parse the json id as a url first?
  if (json.id.origin !== startURL.origin) {
    manifest.set("id", fallbackId);
    return;
  }

  let id;
  try {
      id = new URL(json.id, startURL.origin);
   } catch {
      id = fallbackId;
   }

  // finally, we have the id...
  manifest.set("id", id);
}
```

index.html
dmurph: Chromium currently fetches on every page load. See my comment above about when the changes apply.

Different ID means that this page is referencing a different webapp, and if user agents choose to they can advertise a different install.

Failed parsing means, I believe, we can't do anything. We don't know if it would apply to a current install or a new install

index.html
dmurph: Yes - this was a bug I noticed recently w/ Chromium & our parsing, and the spec. if start_url is "", then it just isn't set. I am planning on creating a pull request to set start_url to the document url if it is "" - which I believe is the only time it isn't set?

I guess if it isn't in the json - we would want to catch that case here. Perhaps we have no id though - if there is no start_url, there kind of isn't a webapp, just page metadata. So maybe we say id is undefined if start_url is processed to undefined?

dmurph: If we allow start_url to always default to document url, then it could lead to confusion if a developer attaches a manifest on multiple pages (and they don't specify start_url), as, with the `id` member, this could mean each page could trigger an web app update to make that page now the new start_url of the app.

I guess my vote would be to make the behavior more predictable / less potentially weird  by requiring the start_url to be defined for the id to be considered? WDYT @philloooo?

index.html
marcoscaceres: Oh, nice! Looks like we can share the name of this member.

index.html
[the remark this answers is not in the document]
marcoscaceres: Ok, that's great to hear @dmurph. For the spec, I think we should have a working assumption that the manifest is updated on every page load, and then make a provision to support custom update behavior that user agents might use. 

We should document the concerns around icons and name in the privacy and security sections.

index.html
philloooo: I was thinking about it too. But I feel other fields are for the whole "app" too? eg: scope / start_url?

index.html
```diff
@@ -808,6 +811,90 @@ <h3>
...
         </section>
       </section>
+      <section>
+        <h3>
+          `id` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest" id="identity">id</dfn></code> member is a <a>string</a> that
+          represents the <dfn>identity</dfn> for the application.
```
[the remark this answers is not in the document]
philloooo: why is it scoped to web-application that's not used for any other fields?

index.html
philloooo: I updated the spec to add a final statement of "Otherwise, set manifest["id"] to |id|" And reordered the same origin check and URL processing result check.
It should match 
```javascript
/**
 * @param {Object} json
 * @param {Map<string, any>} manifest
 */
function processIdMember(json, manifest){
  // Set |manifest|["id"] to |manifest|["start_url"].
  manifest.set("id", manifest.get("start_url"));

  // If the type of |json|["id"] is not [=string=], return.
  if (typeof json.id !== "string") return;

  // Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
  // with |manifest|["start_ur"]'s origin as the base URL.
  let id;
  try {
    id = new URL(json.id, manifest.get("start_url").origin);
  } catch {
   // if |id| is failure, return.
    return;
  }

  // If |id| is not same origin as |manifest|["start_url"], return.
  if(id.origin !== manifest.get("start_url").origin)) return;
  
  //Otherwise, set |manifest|["id"] to |id|
  manifest.set("id", id);
}
```

index.html
philloooo: #670 changes the default value to manifest's base path. I think alternatively I can update the start_url processing algorithm here to just describe what's happending in all user agents now. Which is do 
```
Set |manifest|["start_ur"] to |document URL:URL| 
``` 
in the beginning of the processing alrogithm, to make it clear that when it's undefined, default to document_url
WDYT?

index.html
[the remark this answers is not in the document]
philloooo: Updated the wording, can you take another look?

index.html
philloooo: clarified with using [=url/equal=]

.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,16 @@
+{
```
dmurph, line 1: delete this file from checkout?

index.html
dmurph: I agree we should resolve this before submitting this pull request. I think there are a couple ideas that @marcoscaceres  is playing with which are looking good. (#670 and #991)

philloooo: oh nice, #991 looks like what I was thinking

johnwilander:
Can this be misused to (also) carry a persistent user ID?

philloooo:
hi @johnwilander.
The id is not exposed to the developers through API  so there is no way to detect the id of your app, if a mal-intented app is trying to use the manifest id to carry the user ID for user tracking.

johnwilander:
> The id is not exposed to the developers through API so there is no way to detect the id of your app, if a mal-intented app is trying to use the manifest id to carry the user ID for user tracking.

Thanks! Is the ID exposed in HTTP requests? Are there side channels such as "thing X will only happen if there's a matching manifest ID" where X can be a network load, a flag being set, an API returning a different value etc?

alancutter:
> Are there side channels such as "thing X will only happen if there's a matching manifest ID" where X can be a network load, a flag being set, an API returning a different value etc?

The [scope_extensions](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md) proposal intends to use app IDs to enable web apps extending their scope to other origins. Combined with a feature like link capturing a web app would know its ID is one listed in the cross origin web-app-origin-association.json file via capturing a user navigation to that other origin as a JS event.

philloooo:
> Thanks! Is the ID exposed in HTTP requests? Are there side channels such as "thing X will only happen if there's a matching manifest ID" where X can be a network load, a flag being set, an API returning a different value etc?

Also a manifest update can only happen when there is a matching ID. But I am not sure if the app can know whether a local web app is updated with latest manifest through network requests.

dmurph:
> > Are there side channels such as "thing X will only happen if there's a matching manifest ID" where X can be a network load, a flag being set, an API returning a different value etc?
> 
> The [scope_extensions](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md) proposal intends to use app IDs to enable web apps extending their scope to other origins. Combined with a feature like link capturing a web app would know its ID is one listed in the cross origin web-app-origin-association.json file via capturing a user navigation to that other origin as a JS event.

Interesting! But as far as I can tell, super infeasible

So the case setup would be:
* Web dev sets id to, say `user_1123314` on app.com
* Web dev sets scope extensions to other origins, say app2.com
  * The hooking of these origins can't happen yet - app2.com has to not serve the association file yet, or other means of delay
* User clears browsing data (uninstalling would completely remove this record, so that case is fine)

The 'attack' - In order for the website to discover the id, they have to:
 1. guess the id in the associations file on app2.com. If the id was 'user_1123314`, then, well, let's say they guess N number of user Ids
 2. get the user to navigate to app2.com after this guess was correct. The fact that app.com gets the launch from app2.com would tell it that it guessed correctly.
 3. However, this would tell them that ONE of the guesses in the N guesses was right. They would have to modify their associations on app2.com to continue narrowing down to the correct id.

So this means that the attack would require:
1. Site to have app2.com have, say, half of their user id space as id associations in the file (thousands? Millions?)
1. the site to ask the you "hey can you click on these thousands/millions of links until I tell you to stop"
1. When a link "works", the site then has to then modify the origin association file to have half again (binary search), and go back to step 1.

This seems like an infeasible attack vector to me, would take absolutely forever & require a ton of user interaction.

If we decide this is somehow feasible (interested, then there are some mitigations):
* Restrict the amount of origin associations you can make in the origin association file.
* Restrict the refresh rate of this file. refresh it once a day? This means it would take 16 days minimum of the above process, including user doing all the guessing, to narrow down to a single id in 100,000. With association restriction, this removes the binary search, and makes it.... 10,000 days (28 years) if we limit to 10 associations? Again assuming the user would play ball and click on 100,000 links? 

Anyways - let me know if my thinking is off here.

dmurph:
> Also a manifest update can only happen when there is a matching ID. But I am not sure if the app can know whether a local web app is updated with latest manifest through network requests.

I think, similar to the 'match' case I outlined above, this would rely on getting a 'matching' id in the served manifest after browsing data was cleared. The serviceworker could tell if there was a match & an update if the icons are fetched, for example.

However, we throttle manifest updates to 1 per day. So the site gets one guess a day at the user id.

johnwilander:
An attacker will likely not have to search its entire user space. We have to assume that 1) the attacker knows exactly which users it hasn’t heard from in a certain time window, 2) the attacker will be able to bucketize its user space on device info, and 3) any fingerprinting available such as IP address, time zone, language settings, and user agent information can be used to further shard the user space.

dmurph:
> An attacker will likely not have to search its entire user space. We have to assume that 1) the attacker knows exactly which users it hasn’t heard from in a certain time window, 2) the attacker will be able to bucketize its user space on device info, and 3) any fingerprinting available such as IP address, time zone, language settings, and user agent information can be used to further shard the user space.

So I guess these attacks can simplify probably to "The attacker would be able to make 1 guess a day at the id". There could be further mitigations around "if the origin association file or manifest file update X days in a row, increase update time" etc. But I guess the first question would be - is 1 guess a day too fast?

philloooo:
So I am trying to clarify how the attack can happen to make sure we are talking about the same scenario:
1. User visits appA.
2. appA assigns user_id_1 to this user, and also dynamically sets  `id` to `user_id_1` in the manifest and also use this id in appB origin association file.
3. User installs appA.
4. User clears the browser session.
5. User visits appA again.
6. appA guesses to set id to  `user_id_x` in manifest and sees if a manifest update happens or if a user can click to follow link capturing to appB, if the detection was successful, the app successfully tracked the user between sessions.
7. If appA failed to guess the user, it sets a new user_id_2 and consider it a new user.

Both manifest update and link capturing scenarios are limited by the manifest update throttling.

dmurph:
Difficult case also, for an attacker, is knowing IF they should search or not. Impossible to know the the visit is a new user or not. Second, if they are serving a new manifest all of the time trying to figure out this id, the id will always be different, allowing any of these 'searching' apps to be installed. This makes the search now inconclusive, as any of those new ids would trigger false positives with the search, and cause the id association to be wrong.

Regarding other fields that are sensitive here - I don't think any of them affect future behavior? I don't see how they would cause privacy exposure.

index.html
mgiuca: I'm not very familiar with the MiniApps work. Are we intentionally aligning with them / is there some assumed compatibility between the two standards? Do we try to align opportunistically (i.e. if there's something that "feels the same" we try to align but we don't go out of our way to make it compatible)?

I think we could align with them on the name, but also we have a somewhat different semantics for our ID.

Ignoring MiniApps, I would have a slight preference for `id` over `app_id`, echoing what @philloooo said. It's called the "Web App Manifest", so "app" is implied, and we could similarly argue that it should be called `app_scope`, `app_name`, etc. So my vote is for `id`.

espinr: > I'm not very familiar with the MiniApps work. Are we intentionally aligning with them / is there some assumed compatibility between the two standards? Do we try to align opportunistically (i.e. if there's something that "feels the same" we try to align but we don't go out of our way to make it compatible)?

The other way round :) MiniApp is trying to reuse/align with the existing standards. As a first approach, the MiniApp vendors agreed to extend the Web App Manifest. 

The purpose and semantics of the MiniApp's [`app_id`](https://w3c.github.io/miniapp-manifest/#dfn-app_id) member are the same (identify the app univocally), but most of the MiniApp vendors don't use URLs to identify the app. I can bring this discussion to the MiniApp WG and see if this member could be reused.

I agree with you @mgiuca @philloooo that the term `id` would be enough since `app` may be redundant.

marcoscaceres: Alright, let's stick with `id`.

index.html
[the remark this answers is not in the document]
marcoscaceres: I guess it's ok to leave it unstopped for now... I have a feeling it will eventually get used in other (external) contexts, which is why I suggested it. However, we can deal with that when it actually happens.

dmurph:
This change LGTM % the comment from marcos about link-ifying [=identity=]

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          To <dfn>process the `id` member</dfn>, given [=object=]
+          |json:JSON|, [=ordered map=] |manifest:ordered map|:
+        </p>
+        <ol class="algorithm">
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
+          with |manifest|["start_ur"]'s origin as the base URL.
```
kenchris, line 856: start_url, missing l

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          Below table shows some example cases of the |identity| processing algorithm.
+          <table class="data">
+            <tr>
+              <th>|json|["id"]</th>
+              <th>|manifest|["start_url"]</th>
+              <th>|manifest|["id"]</th>
+            </tr>
+            <tr>
+              <td><i>undefined</i></td>
+              <td>"https://example.com/resources/start"</td>
```
kenchris, line 874: Could we get some examples with subdomains, like used on GitHub Pages, e.g. w3c.github.io

kenchris: Also, maybe show examples of weirded id's like with spaces in them

mgiuca: Yep, this table is fantastic for illustrating cases, but more cases would be better! I see on the [TAG review](https://github.com/w3ctag/design-reviews/issues/668) they asked about "What constitutes a valid ID?" -- I don't think this needs to be explicitly addressed in normative text, since the fact that it runs through the URL Parser is sufficient to specify that. But some examples would be helpful.

Maybe some Unicode characters. Some percent-encoded sequences to show that this is actually parsed as a URL (e.g. `%6D` should generate an id of ...`m`).

Also I would like to add an example with an input of "`/foo`" in addition to "`foo`", which should generate the same result. (As I have been saying in internal discussions, even though the leading slash is optional, I would very much like us to recommend using it in all examples, since having a URL without a leading slash be parsed relative to the origin is quite confunsing behaviour IMO and differs from how we parse all other URLs in the manifest.)

(Arguably, I would still like to see us parse the ID relative to the manifest URL for consistency with other URLs, and thus soft-mandate the leading slash, but that decision seems to have been made long ago and I'm not sure what the decision was based on.)

index.html
marcoscaceres, line 856:
```suggestion
          with |manifest|["start_url"]'s origin as the base URL.
```

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          The [=identity=] takes the form of a URL, which is same origin as the
+          start URL.
+        </p>
+        <p>The |identity| is used by user agents to uniquely identify the
+        application universally. When the user agent sees a manifest with an
+        |identity| that does not correspond to an already-installed application,
+        it SHOULD treat that manifest as a description of a distinct
+        application, even if it is served from the same URL as that of another
+        application. When the user agent sees a manifest with an |identity|
+        matching that of an already-installed application, it SHOULD be used as
```
marcoscaceres, line 834:
```suggestion
        <p>The [=identity=] is used by user agents to uniquely identify the
        application universally. When the user agent sees a manifest with an
        |identity| that does not correspond to an already-installed application,
        it SHOULD treat that manifest as a description of a distinct
        application, even if it is served from the same URL as that of another
        application. When the user agent sees a manifest with an [=identity=]
        matching that of an already-installed application, it SHOULD be used as
```

index.html
```diff
@@ -1222,6 +1316,27 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents SHOULD re-fetch and process the manifest on every page
+            load. When the processing is successful, user agents MAY apply updated
```
marcoscaceres, line 1321: The first part is covered by HTML... or at least, we should update the HTML spec instead, in the link relation section.  

```suggestion
            When the processing is successful, user agents MAY apply updated
```

index.html
```diff
@@ -1222,6 +1316,27 @@ <h3>
...
+            processed manifest's [=manifest/identity=] with the identity of the
+            app being updated. If it is not equal, the user agent MUST NOT
+            update the installed application's manifest. 
+          </p>
+          <section>
+            <h3>
+              Privacy and security consideration
+            </h3>
+            <p>
+              User agents SHOULD be careful about performing update to manifest
```
marcoscaceres, line 1334: I think we need to just fold this into the section above. We probably don't need to call it out.... in fact, I think an update is really just a "re-install", so we might be able to leverage what's already in the installation section.

index.html
```diff
@@ -1222,6 +1316,27 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            User agents SHOULD re-fetch and process the manifest on every page
+            load. When the processing is successful, user agents MAY apply updated
+            manifest to any current and future <a>application contexts</a>
+            associated with the application. When the user agent fetches a
+            manifest as part of an update check, it MUST [=url/equal|compare=] the fetched
```
marcoscaceres, line 1324: I think this is gives general guidance, but it's still a bit "hand wavy" (which may turn out to be ok, given how reliant this is on browser UI and prompts)... 

We should maybe split this into two parts... just the id member, and then we can tackle updating more specifically.

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          with |manifest|["start_url"]'s origin as the base URL.
+          </li>
+          <li>If |id| is failure, return.
+          </li>
+          <li>If |id| is not same origin as |manifest|["start_url"], return.
+          </li>
+          <li>Otherwise, set |manifest|["id"] to |id|.</li>
+        </ol>
+        <aside class="example">
+          Below table shows some example cases of the |identity| processing algorithm.
```
marcoscaceres, line 865:
```suggestion
          <p>Below table shows some example cases of the |identity| processing algorithm.</p>
```

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+        a signal that this manifest is a replacement for the already-installed
+        application's manifest, and not a distinct application, even if it is
+        served from a different URL than the one seen previously.
+        </p>
+        <p>
+        The |identity| is considered matching when the processed |manifest|["id"] [=url/equal|equals=].
+        </p>
+        <p class="note">
+          The |identity| can be used by a service that collects lists of web
+          applications to uniquely identify applications.
```
marcoscaceres, line 844:
```suggestion
        The [=identity=] is considered matching when the processed |manifest|["id"] [=url/equal|equals=].
        </p>
        <p class="note">
          The |identity| can be used by a service that collects lists of web
          applications to uniquely identify applications.
```

index.html
marcoscaceres: For completeness, I guess we don't mind if the computed ID is not in scope as it's not actually involved in any navigations.  We might want to make a note of that.

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
+          with |manifest|["start_url"]'s origin as the base URL.
+          </li>
+          <li>If |id| is failure, return.
+          </li>
+          <li>If |id| is not same origin as |manifest|["start_url"], return.
```
marcoscaceres, line 860:
```suggestion
          <li>If |id| is not [=same origin=] as |manifest|["start_url"], return.
```

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+            </tr>
+            <tr>
+              <td><i>undefined</i></td>
+              <td>"https://example.com/resources/start"</td>
+              <td>"https://example.com/resources/start"</td>
+            </tr>
+            <tr>
+              <td>""</td>
+              <td>"https://example.com/resources/start"</td>
+              <td>"https://example.com/"</td>
```
marcoscaceres, line 880: Wouldn't this be?: 

```suggestion
              <td>""</td>
              <td>"https://example.com/resources/start"</td>
              <td>"https://example.com/resources/start"</td>
```

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+            </tr>
+            <tr>
+              <td>"/"</td>
+              <td>"https://example.com/resources/start"</td>
+              <td>"https://example.com/"</td>
+            </tr>
+            <tr>
+              <td>"foo"</td>
+              <td>"https://example.com/resources/start"</td>
+              <td>"https://example.com/foo"</td>
```
marcoscaceres, line 890: This one is also incorrect, I believe: 

```suggestion
              <td>"https://example.com/resources/start/foo"</td>
```

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+          </li>
+          <li>Otherwise, set |manifest|["id"] to |id|.</li>
+        </ol>
+        <aside class="example">
+          <p>Below table shows some example cases of the |identity| processing algorithm.</p>
+          <table class="data">
+            <tr>
+              <th>|json|["id"]</th>
+              <th>|manifest|["start_url"]</th>
+              <th>|manifest|["id"]</th>
```
marcoscaceres, line 870: We might want to add a "rationale" or "discussion" `<th>` to explain why for each case.

index.html
```diff
@@ -811,6 +814,94 @@ <h3>
...
+            </tr>
+            <tr>
+              <td><i>undefined</i></td>
+              <td>"https://example.com/resources/start"</td>
+              <td>"https://example.com/resources/start"</td>
+            </tr>
+            <tr>
+              <td>""</td>
+              <td>"https://example.com/resources/start"</td>
```
marcoscaceres, line 879: Nit: let's drop the "resources"... and instead of "foo", let's call it "my-app", just so it's a bit more concrete.

index.html
philloooo: the app's origin is used as the base url for the url parsing instead of start_url. This makes sure that the id stability doesn't depend on start_url.

index.html
philloooo: the app's origin is used as the base url for the url parsing instead of start_url. This makes sure that the id stability doesn't depend on start_url.

index.html
philloooo: It seems a very different UX from re-install as it's performed without user action. I agree the above section is very hand-wavy, and the privacy/security section warns about what NOT to do (update identity related information without user consent). And leave the rest to individual implementation. How should I describe this better? Or I just move this as the next paragraph to:
```
User agents SHOULD re-fetch and process the manifest on every page load.
When the processing is successful, user agents MAY apply updated manifest
to any current and future application contexts associated with the application.
```
?

index.html
philloooo: Cool, https://html.spec.whatwg.org/#fetching-and-processing-a-resource-from-a-link-element mentions:

the element must delay the load event of the element's node document until all the attempts to fetch and process the linked resource and its critical subresources are complete. 

I can update the language to - On every page load, the manifest is re-fetched and processed. When the processing is successful, ...

index.html
philloooo: Added a note for the scope.
Added ASCII character and special character examples.

index.html
philloooo: That will make it exceed the width of the parent element and you have to scroll to see the column. I added a paragraph to explain the main points.

index.html
marcoscaceres: Oh, whoops! Yes, I misread the algorithm. That's easy to miss, so, I think we should pull origin out into a variable and then use that as the base.

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+          |json:JSON|, [=ordered map=] |manifest:ordered map|:
+        </p>
+        <ol class="algorithm">
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
+          with |manifest|["start_url"]'s origin as the base URL.
+          </li>
```
marcoscaceres, line 862:
```suggestion
          <li>Let |base origin| be |manifest|["start_url"]'s [=url/origin=].</li> 
          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"] with  |base origin| as the base URL.
          </li>
```

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+              <td>"https://example.com/my-app/start"</td>
+              <td>"https://example.com/m"</td>
+            </tr>
+            <tr>
+              <td>"😀"</td>
+              <td>"https://example.com/my-app/start"</td>
+              <td>"https://example.com/%F0%9F%98%80"</td>
+            </tr>
+          </table>
+          <p>Since id is resolved against start_url's origin, providing id as
```
marcoscaceres, line 923:
```suggestion
          <p>Since [=manifest/id=] is resolved against [=manifest/start_url=]'s [=URL/origin=], providing
```

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+          "../foo", "foo", "/foo", "./foo" all resolves to the same thing. It is
```
marcoscaceres, line 924:
```suggestion
          "../foo", "foo", "/foo", "./foo" all resolves to the same [=identifier=]. It is
```

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+          RECOMMENDED to always specify with a leading "/" to be explicit that
```
marcoscaceres, line 925:
```suggestion
          As such, best practice is to use a leading "/" to be explicit that
```

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+          the id is a root-relative url path. Also the same url
+          encoding/decoding rules are applied to the id processing algorithm.
```
marcoscaceres, line 927:
```suggestion
          the id is a root-relative url path. Also, standard
          encoding/decoding rules are apply to the id processing algorithm, as per the [[[URL]]].
```

marcoscaceres:
Filed Gecko bug: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1731940

Will post patch soon....

marcoscaceres:
Filed webkit bug:
https://bugs.webkit.org/show_bug.cgi?id=230596

marcoscaceres (approved):
Tentative approval, as technically, this is now looking good. 

There are still a few editorial nits still - and I still we should split updating out to a new PR, but apart from that I'm good with this.

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+              <th>|manifest|["start_url"]</th>
+              <th>|manifest|["id"]</th>
+            </tr>
+            <tr>
+              <td><i>undefined</i></td>
+              <td>"https://example.com/my-app/start"</td>
+              <td>"https://example.com/my-app/start"</td>
+            </tr>
+            <tr>
+              <td>""</td>
```
marcoscaceres, line 883: Hmm... I don't know about this case... I think we should treat the empty string as missing.

marcoscaceres: We do the same for start_url, I think (i.e., treat empty string as whatever the default is).

index.html
```diff
@@ -811,6 +814,120 @@ <h3>
...
+        <p>
+          To <dfn>process the `id` member</dfn>, given [=object=]
+          |json:JSON|, [=ordered map=] |manifest:ordered map|:
+        </p>
+        <ol class="algorithm">
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
```
marcoscaceres, line 860: See below about treating this as missing: 

```suggestion
          <li>If |json|["id"] is the empty string, return.
          </li>
          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
```

index.html
philloooo: I don't have a strong preference here. @dmurph @mgiuca WDYT?

philloooo:
I can split the updating part to a new PR, but then the paragraph about id within the manifest update section won't make much sense independently. Or I should move the whole section to a PR blocked by this one?

index.html
mgiuca: I think I have a weak preference to agree with @marcoscaceres and treat it as missing (fallback to `start_url`).

Overall I think we've gone down a pretty confusing path with this being resolved against the origin but defaulting to `start_url`.

Not sure if it's too late to reconsider, but could we actually **resolve it against `start_url`**.

Pros:

* It just means if it's empty, or missing, it would naturally be equal to `start_url`, without needing a special case for "if it's empty".
* It makes it much more important to add a leading '/' to the `id`, to anchor it to the origin. Which is something I've been, on a few other threads, saying we should be recommending anyway, to really make it clear that this is a URL and the '/' makes it unambiguous that it is relative to the origin.

Cons:

* It means if you don't put a leading '/' on the `id`, if you change the `start_url` your identity will change, which destroys the whole point of having `id`.

So that Con doesn't sound great, right? But if we go with "`""` means `start_url`", that Con also applies, in that one special case — if you want the id to be empty you need to remember to write `"/"` rather than `""`. So maybe my suggestion makes sense and it's just quite important to prefix your `id` with `/`.

mgiuca: If we aren't keen on my above suggestion (a quick chat with Phillis offline suggests not), then I think going with Marcos' suggestion is good. It's actually consistent with how [scope is parsed](https://www.w3.org/TR/appmanifest/#dfn-process-the-scope-member):

- If it's missing, invalid, **or empty**, it defaults to `start_url`.
- If not, it's parsed against manifest URL.

So we've already got that weird condition "if empty default to something other than the base URL it's being parsed against" in the `scope` processor. We would be doing the same thing here.

Ultimately, if we recommend that users always start `id` with '/', then none of this matters, so let's make sure we always recommend that and do this small change and move on.

philloooo: thanks Matt! Yeah my main concern is with the con you already said. If we make start_url as the base_url, then I think the leading slash has to be required, which then basically mean the base_url is the root. Otherwise using start_url as base_url + not requiring to have leading slash make it very error-prone to devs.

I will make the update for the empty string!

index.html
marcoscaceres: > It seems a very different UX from re-install as it's performed without user action. 

Agree. 

>  How should I describe this better? 

I think what you stated in this paragraph is good (in as far that it does highlight what the risks are, and what browsers MAY choose to do about it). So yeah, if you move this up, I think that will give us a good foundation to build on.

marcoscaceres: > It seems a very different UX from re-install as it's performed without user action.

On second thought, I'm unsure if all UAs will do this without user action (although, implemented in multiple OSs... like Chrome OS and Windows). I feel like this might still be an area where we might see some variation on how different implementations handle this. 

In any case, I think the text you've proposed covers the privacy/security aspects that are of most concern with respect to upgrades.

marcoscaceres (commented):
@philloooo I was writing some additional tests for Gecko and realized we didn't deal with URL fragments. I think we should probably remove them? Wdyt?

index.html
```diff
@@ -811,6 +814,123 @@ <h3>
...
+          <li>If |json|["id"] is the empty string, return.
+          </li>
+          <li>Let |base origin| be |manifest|["start_url"]'s [=url/origin=].</li> 
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"] with  |base origin| as the base URL.
+          </li>
+          <li>If |id| is failure, return.
+          </li>
+          <li>If |id| is not [=same origin=] as |manifest|["start_url"], return.
+          </li>
+          <li>Otherwise, set |manifest|["id"] to |id|.</li>
```
marcoscaceres, line 869:
```suggestion
          <li>Remove [=URL/fragment=] from |id|.</li>
          <li>Set |manifest|["id"] to |id|.</li>
```

index.html
```diff
@@ -811,6 +814,123 @@ <h3>
...
+              <td>"https://example.com/my-app/start"</td>
+              <td>"https://example.com/foo"</td>
+            </tr>
+            <tr>
+              <td>"https://anothersite.com/foo"</td>
+              <td>"https://example.com/my-app/start"</td>
+              <td>"https://example.com/my-app/start"</td>
+            </tr>
+            <tr>
+              <td>"%6D"</td>
```
marcoscaceres, line 916: I don't think this one is right... try:

```JS
console.log(new URL("%6D", document.location).href)
```

It just gives back "https://whatever/%6D" (i.e., it doesn't do the URL component decoding)

philloooo: hmm, it's working for me?
Screenshot from 2021-09-22 21-29-52

marcoscaceres: Maybe we should replace this example with a query + fragment example (i.e., query is retained, fragment gets dropped).

index.html
marcoscaceres: You are right. I'd still be inclined to remove it from the id tho, just for simplicity. We could do that in step 2, and then again here?

marcoscaceres: Just to be clear, I'd want to retain the same identity if I updated:


```JSON
{
    "start_url": "foo.html#here",
}
```

To:

```JSON
{
    "start_url": "foo.html#there",
}
```

And not consider that an entirely new application identity.

index.html
marcoscaceres: oh lols no... Firefox and Safari return different results to Chrome on this 🥴

Safari 15: 
Screen Shot 2021-09-23 at 3 27 20 pm

Firefox 92.0: 
Screen Shot 2021-09-23 at 3 28 04 pm

And yes, you are right... Chrome returns "m".

Would need to check URL spec for which is right.

marcoscaceres: @annevk, sorry, I couldn't figure the above out from reading the URL spec. Are Firefox and Safari right? or is Chrome right in the case above?

annevk: Firefox and Safari are correct. https://jsdom.github.io/whatwg-url/#url=aHR0cHM6Ly9leGFtcGxlLmNvbS8lNkQ=&base=YWJvdXQ6Ymxhbms= might be useful for such questions.

marcoscaceres: Oh! Nice!!! I'd not seen that tool. Thank you!

index.html
philloooo: I think as a developer, if I'd want to change the start_url but retaining the same identity, I should specify the id field instead?
```
{
  "start_url": "foo.html#here",
}
```
To:
```
{
  "start_url": "foo.html#there",
  "id": "foo.html#here",
}
```

If we remove the fragment, apps that have fragment in start_url today will get a new id when browser ships this feature.

index.html
philloooo: wow, thanks for finding that! I will update the example and investigate the behavior on Chrome end.

philloooo: Looks like there is a discussion here
https://github.com/whatwg/url/issues/606 

I will remove this example since this is not settled yet.

index.html
marcoscaceres: Sorry, I’m still not following? No app would have an identity today, as we haven’t settled on this feature yet?

index.html
```diff
@@ -811,6 +814,118 @@ <h3>
...
+        it SHOULD treat that manifest as a description of a distinct
+        application, even if it is served from the same URL as that of another
+        application. When the user agent sees a manifest with an [=identity=]
+        matching that of an already-installed application, it SHOULD be used as
+        a signal that this manifest is a replacement for the already-installed
+        application's manifest, and not a distinct application, even if it is
+        served from a different URL than the one seen previously.
+        </p>
+        <p>
+        The [=identity=] is considered matching when the processed |manifest|["id"] [=url/equal|equals=].
```
mgiuca, line 840: This is grammatically incorrect ("equals" what?).

Also "matching" isn't properly defined even though it's used above.

I would just delete this paragraph and then formalize the sentence in the previous paragraph:

> When the user agent sees a manifest with an identity matching that of an already-installed application, ...

becomes

> When the user agent sees a manifest _manifest_ where _manifest_["id"] [equals](https://url.spec.whatwg.org/#concept-url-equals) the identity of an already-installed application, ...

index.html
philloooo: Right, for browsers that use start_url for the identity, this will change app identity from start_url to start_url with fragment removed.
You mentioned you think dropping the fragment is better for simplicity, but it seems to me that following existing processing logic for start_url is simpler?

index.html
philloooo: sounds good, thanks!

index.html
philloooo: Ok I chatted more with dmurph@ offline. It makes sense to drop fragment considering #here or #there both link to the same page so they should be considered identical.
I will make the update to drop the fragment then!

marcoscaceres: Thanks @philloooo. Appreciate you gathering the additional input from @dmurph. I've updated the Gecko implementation to match. 

I know it took us a a while, but really happy with were we are ending up with this 🥰

index.html
```diff
@@ -811,6 +814,116 @@ <h3>
...
+          <li>Otherwise, set |manifest|["id"] to |id|.</li>
+        </ol>
+        <aside class="example">
+          <p>Below table shows some example cases of the |identity| processing algorithm.</p>
+          <table class="data">
+            <tr>
+              <th>|json|["id"]</th>
+              <th>|manifest|["start_url"]</th>
+              <th>|manifest|["id"]</th>
+              <th>Explaination</th>
```
marcoscaceres, line 876: Given the paragraph below the table, Explanation can go away again, right?

philloooo: oops, forgot to remove this, thanks for catching.

index.html
```diff
@@ -1222,6 +1343,30 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
```
marcoscaceres, line 1345:
```suggestion
```

index.html
```diff
@@ -1221,7 +1342,30 @@ <h3 id="applying">
           <h3>
             Updating the manifest
           </h3>
-          <aside class="issue" data-number="446"></aside>
+          <p>
```
marcoscaceres, line 1345:
```suggestion
          <aside class="issue" data-number="446"></aside>
          <p>
```

index.html
```diff
@@ -1222,6 +1343,30 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
```
marcoscaceres, line 1346: @philloooo, would you mind sending the updating part as a separate PR?

marcoscaceres (changes requested):
Request to send update as a different PR.

philloooo:
> Request to send update as a different PR.

Moved to https://github.com/w3c/manifest/pull/1011

philloooo:
Moved manifest update to https://github.com/w3c/manifest/pull/1011

marcoscaceres (approved):

marcoscaceres:
Some of the w3c systems are currently down, which is causing the CI errors.

PR #1005, opened 2021-09-13
Adding file_handlers and launch consumer

### LINKED ISSUE #626: Add a means to associate a PWA with a file extension
It's common on every OS to associate file extensions with some application (e.g., .psd opens in PhotoShop).

It would be great to figure out how we can associate files on a user's device to open with an installed web application.

dmurph:
Closes #626

* [X] Chromium (https://bugs.chromium.org/)

* [X] [updated JSON schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/web-manifest.json)

Commit message:

Adds file_handlers member and launch consumer algorithm to handle file launches.

Explainer: https://github.com/WICG/file-handling/blob/main/explainer.md

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest/pull/1005.html" title="Last updated on Oct 6, 2021, 11:07 PM UTC (d089331)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1005/e127fd7...dmurph:d089331.html" title="Last updated on Oct 6, 2021, 11:07 PM UTC (d089331)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest/pull/1005.html" title="Last updated on Mar 30, 2022, 6:51 PM UTC (0d29fd1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1005/8d0663e...dmurph:0d29fd1.html" title="Last updated on Mar 30, 2022, 6:51 PM UTC (0d29fd1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -17,9 +17,14 @@
       previousMaturity: "WD",
       formerEditors: [
         {
+          name: "Mounir Lamouri",
+          company: "Google Inc.",
+          companyURL: "https://www.google.com/",
+        },
+        {
           name: "Rob Dolin",
           company: "Microsoft Corporation",
-          companyURL: "https://www.microsoft.com/en-us/",
+          companyURL: "https://www.microsoft.com/",
         },
       ],
       editors: [
@@ -34,12 +39,6 @@
           company: "Intel Corporation",
           companyURL: "https://intel.com/",
           w3cid: 57705,
-        },
-        {
-          name: "Mounir Lamouri",
-          company: "Google Inc.",
-          companyURL: "https://www.google.com/",
-          w3cid: 45389,
         },
         {
           name: "Anssi Kostiainen",
--- a/index.html
+++ b/index.html
@@ -41,12 +41,6 @@
           w3cid: 57705,
         },
         {
-          name: "Anssi Kostiainen",
-          company: "Intel Corporation",
-          companyURL: "https://intel.com/",
-          w3cid: 41974,
-        },
-        {
           name: "Matt Giuca",
           company: "Google Inc.",
           companyURL: "https://www.google.com/",
@@ -57,6 +51,18 @@
           company: "Microsoft Corporation",
           companyURL: "https://microsoft.com/",
           w3cid: 43672,
+        },
+        {
+          name: "Daniel Murphy",
+          company: "Google Inc.",
+          companyURL: "https://www.google.com/",
+          w3cid: 90918,
+        },
+        {
+          name: "Anssi Kostiainen",
+          company: "Intel Corporation",
+          companyURL: "https://intel.com/",
+          w3cid: 41974,
         },
       ],
       group: "webapps",
@@ -171,6 +177,8 @@
         </li>
         <li>[=manifest/icons=]
         </li>
+        <li>[=manifest/identity=]
+        </li>
         <li>[=manifest/lang=]
         </li>
         <li>[=manifest/name=]
@@ -232,6 +240,7 @@
                 "sizes": "128x128"
               }],
               "scope": "/",
+              "id": "superracer",
               "start_url": "/start.html",
               "display": "fullscreen",
               "orientation": "landscape",
@@ -809,7 +818,195 @@
             time thereafter, a user agent allows the user to inspect and, if
             necessary, modify the <a>start URL</a> of an application.
           </p>
+          <p>
+            Additionally, developers MUST NOT use the <a>start URL</a> to
+            include information that uniquely identifies a user (e.g.,
+            "?user=123" or "/user/123/", or "https://user123.foo.bar").
+          </p>
         </section>
+      </section>
+      <section>
+        <h3>
+          `id` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">id</dfn></code> member is a <a>string</a> that represents
+          the <dfn>identity</dfn> for the application. The [=identity=] takes
+          the form of a URL, which is same origin as the start URL.
+        </p>
+        <p>
+          The [=identity=] is used by user agents to uniquely identify the
+          application universally. When the user agent sees a manifest with an
+          [=identity=] that does not correspond to an already-installed
+          application, it SHOULD treat that manifest as a description of a
+          distinct application, even if it is served from the same URL as that
+          of another application. When the user agent sees a manifest where
+          |manifest|["id"] [=url/equal|equals=] the [=identity=] of an
+          already-installed application, it SHOULD be used as a signal that
+          this manifest is a replacement for the already-installed
+          application's manifest, and not a distinct application, even if it is
+          served from a different URL than the one seen previously.
+        </p>
+        <p class="note">
+          The [=identity=] can be used by a service that collects lists of web
+          applications to uniquely identify applications.
+        </p>
+        <p class="note">
+          The [=identity=] is processed like a URL but it doesn't point to a
+          resource that can be navigated to, so it's not required to be
+          [=URL/within scope=].
+        </p>
+        <p>
+          To <dfn>process the `id` member</dfn>, given [=object=] |json:JSON|,
+          [=ordered map=] |manifest:ordered map|:
+        </p>
+        <ol class="algorithm">
+          <li>Set |manifest|["id"] to |manifest|["start_url"].
+          </li>
+          <li>If the type of |json|["id"] is not [=string=], return.
+          </li>
+          <li>If |json|["id"] is the empty string, return.
+          </li>
+          <li>Let |base origin| be |manifest|["start_url"]'s [=url/origin=].
+          </li>
+          <li>Let |id:URL| be the result of [=URL Parser|parsing=] |json|["id"]
+          with |base origin| as the base URL.
+          </li>
+          <li>If |id| is failure, return.
+          </li>
+          <li>If |id| is not [=same origin=] as |manifest|["start_url"],
+          return.
+          </li>
+          <li>Set |manifest|["id"] to |id|.
+          </li>
+        </ol>
+        <aside class="example" title="Resulting ids">
+          <p>
+            Below table shows some example `id`s resulting from the [=process
+            the `id` member=] steps.
+          </p>
+          <table class="data">
+            <tr>
+              <th>
+                |json|["id"]
+              </th>
+              <th>
+                |manifest|["start_url"]
+              </th>
+              <th>
+                |manifest|["id"]
+              </th>
+            </tr>
+            <tr>
+              <td>
+                <i>undefined</i>
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                <i>undefined</i>
+              </td>
+              <td>
+                "https://example.com/my-app/#here"
+              </td>
+              <td>
+                "https://example.com/my-app/#here"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                ""
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "/"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "foo"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/foo"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "./foo"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/foo"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "https://example.com/foo"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/foo"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "https://anothersite.com/foo"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+            </tr>
+            <tr>
+              <td>
+                "😀"
+              </td>
+              <td>
+                "https://example.com/my-app/start"
+              </td>
+              <td>
+                "https://example.com/%F0%9F%98%80"
+              </td>
+            </tr>
+          </table>
+          <p>
+            Since [=manifest/id=] is resolved against [=manifest/start_url=]'s
+            [=URL/origin=], providing "../foo", "foo", "/foo", "./foo" all
+            resolves to the same [=identifier=]. It is As such, best practice
+            is to use a leading "/" to be explicit that the id is a
+            root-relative url path. Also, standard encoding/decoding rules are
+            apply to the id processing algorithm, as per the [[[URL]]].
+          </p>
+        </aside>
       </section>
       <section>
         <h3>
@@ -1234,6 +1431,9 @@
             </li>
             <li>[=Process the `start_url` member=] passing |json|, |manifest|,
             |manifest URL|, and |document URL|.
+            </li>
+            <li>[=Process the `id` member=] passing |json|, |manifest| and
+            |document URL|.
             </li>
             <li>[=Process the `scope` member=] passing |json|, |manifest|, and
             |manifest URL|.
@@ -2644,67 +2844,13 @@
     </section>
     <section>
       <h2>
-        Display modes
+        Choosing a display mode
       </h2>
       <p>
-        A <dfn>display mode</dfn> represents how the web application is being
-        presented within the context of an OS (e.g., in fullscreen, etc.).
-        Display modes correspond to user interface (UI) metaphors and
-        functionality in use on a given platform. The UI conventions of the
-        display modes are purely advisory and implementers are free to
-        interpret them how they best see fit.
-      </p>
-      <p>
-        This specification defines the following [=display modes=]:
-      </p>
-      <dl>
-        <dt>
-          <dfn data-dfn-for="display mode">fullscreen</dfn>
-        </dt>
-        <dd>
-          Opens the web application with browser UI elements hidden and takes
-          up the entirety of the available display area.
-        </dd>
-        <dt>
-          <dfn data-dfn-for="display mode">standalone</dfn>
-        </dt>
-        <dd>
-          Opens the web application to look and feel like a standalone native
-          application. This can include the application having a different
-          window, its own icon in the application launcher, etc. In this mode,
-          the user agent will exclude standard browser UI elements such as an
-          URL bar, but can include other system UI elements such as a status
-          bar and/or system back button.
-        </dd>
-        <dt>
-          <dfn data-dfn-for="display mode">minimal-ui</dfn>
-        </dt>
-        <dd>
-          This mode is similar to [=display mode/standalone=], but provides the
-          end-user with some means to access a minimal set of UI elements for
-          controlling navigation (i.e., back, forward, reload, and perhaps some
-          way of viewing the document's address). A user agent can include
-          other platform specific UI elements, such as "share" and "print"
-          buttons or whatever is customary on the platform and user agent.
-        </dd>
-        <dt>
-          <dfn data-dfn-for="display mode">browser</dfn> (default)
-        </dt>
-        <dd>
-          Opens the web application using the platform-specific convention for
-          opening hyperlinks in the user agent (e.g., in a browser tab or a new
-          window).
-        </dd>
-      </dl>
-      <p class="note">
-        The [=display mode/fullscreen=] <a>display mode</a> is orthogonal to,
-        and works independently of, the [[[FULLSCREEN]]]. The [=display
-        mode/fullscreen=] <a>display mode</a> affects the fullscreen state of
-        the browser window, while the [[FULLSCREEN]] API operates on an element
-        contained within the viewport. As such, a web application can have its
-        <a>display mode</a> set to [=display mode/fullscreen=], while
-        `document.fullScreenElement` returns `null`, and `fullscreenEnabled`
-        returns `false`.
+        A [=display mode=], as defined in [[MEDIAQUERIES-5]], represents how
+        the web application is being presented within the context of an OS
+        (e.g., in fullscreen, etc.). Display modes correspond to user interface
+        (UI) metaphors and functionality in use on a given platform.
       </p>
       <p>
         Once a user agent [=applies=] a particular <a>display mode</a> to an
@@ -2784,130 +2930,17 @@
         mode/fullscreen=]", "[=display mode/standalone=]", "[=display
         mode/minimal-ui=]", "[=display mode/browser=]" ».
       </p>
-      <section data-cite="css-conditional-3">
-        <h3>
-          The `'display-mode'` media feature
-        </h3>
-        <table class="def descdef mq">
-          <tbody>
-            <tr>
-              <th>
-                Name:
-              </th>
-              <td>
-                <dfn data-dfn-for="@media" data-dfn-type="descriptor"
-                data-export="">display-mode</dfn>
-              </td>
-            </tr>
-            <tr>
-              <th>
-                For:
-              </th>
-              <td>
-                <a data-xref-type="css-at-rule">@media</a>
-              </td>
-            </tr>
-            <tr>
-              <th>
-                Value:
-              </th>
-              <td class="prod">
-                [=display mode/fullscreen=] | [=display mode/standalone=] |
-                [=display mode/minimal-ui=] | [=display mode/browser=]
-              </td>
-            </tr>
-            <tr>
-              <th>
-                Type:
-              </th>
-              <td>
-                discrete
-              </td>
-            </tr>
-          </tbody>
-        </table>
-        <p>
-          The `display-mode` media feature represents, via a CSS media query
-          [[MEDIAQ]], the <a>display mode</a> of the web application. This
-          media feature applies to the top-level browsing context and any child
-          browsing contexts. Child browsing contexts reflect the <a>display
-          mode</a> of the <a>top-level browsing context</a>.
-        </p>
-        <p>
-          A user agent MUST expose the '`display-mode`' media feature
-          irrespective of whether a manifest is being applied to a browsing
-          context. For example, if the end-user puts the whole user agent into
-          fullscreen, then the user agent would reflect this change to CSS and
-          scripts via the '`display-mode`' media feature.
-        </p>
-        <aside class="note">
-          <p>
-            Please note that the `fullscreen` <a>display mode</a> is not
-            directly related to the CSS `:fullscreen` pseudo-class specified in
-            the [[[FULLSCREEN]]]. The `:fullscreen` pseudo-class matches
-            exclusively when a [[HTML]] element is put into the fullscreen
-            element stack. However, a side effect of calling the
-            `requestFullscreen()` method on an element using the [[FULLSCREEN]]
-            API is that the browser window can enter a fullscreen mode at the
-            OS-level. In such a case, both `:fullscreen` and `(display-mode:
-            fullscreen)` will match.
-          </p>
-          <p>
-            On some platforms, it is possible for a user to put a browser
-            window into fullscreen without the aid of the [[[FULLSCREEN]]].
-            When this happens, the `:fullscreen` pseudo class will not match,
-            but `(display-mode: fullscreen)` will match. This is exemplified in
-            CSS code below.
-          </p>
-          <pre class="example css">
-            /* applies when the window is fullscreen */
-            @media all and (display-mode: fullscreen) {
-                ...
-            }
-
-            /* applies when an element goes fullscreen */
-            #game:fullscreen {
-                ...
-            }
-          </pre>
-        </aside>
-        <p>
-          A user agent MUST reflect the applied <a>display mode</a> of the web
-          application via a CSS media query [[MEDIAQ]].
-        </p>
-        <section>
-          <h4>
-            Examples
-          </h4>
-          <p>
-            An example in CSS:
-          </p>
-          <pre class="example css" title="Using 'display-mode' media feature">
-            @media all and (display-mode: minimal-ui) {
-              /* ... */
-            }
-            @media all and (display-mode: standalone) {
-              /* ... */
-            }
-          </pre>
-          <p>
-            Accessing the display-mode media feature in ECMAScript through
-            `matchMedia()` of [[CSSOM-VIEW]]:
-          </p>
-          <pre class="example js" title=
-          "Accessing 'display-mode' media feature with script">
-            const standalone = matchMedia( '(display-mode: standalone)' );
-
-            standalone.onchange = (e) =&gt; {
-              /* handle changes to display mode */
-            }
-
-            if (standalone.matches) {
-              /* do standalone things */
-            }
-          </pre>
-        </section>
-      </section>
+      <p>
+        A user agent MUST reflect the applied <a>display mode</a> of the web
+        application in the '`display-mode`' media feature [[MEDIAQUERIES-5]].
+      </p>
+      <p class="note">
+        A user agent will expose the '`display-mode`' media feature
+        irrespective of whether a manifest is being applied to a browsing
+        context. For example, if the end-user puts the application into
+        fullscreen, then the user agent would reflect this change to CSS and
+        scripts via the '`display-mode`' media feature.
+      </p>
     </section>
     <section id="priv-sec">
       <h2>
@@ -3012,13 +3045,12 @@
         animated transition, or speaking the text "Launching application X".
       </p>
       <p>
-        The `'display-mode'` media feature allows an origin access to aspects
-        of a user’s local computing environment and, together with the
-        `display` member, allows an origin some measure of control over a user
-        agent’s native UI: Through a CSS media query, a script can know the
-        display mode of a web application. An attacker could, in such a case,
-        exploit the fact that an application is being displayed in fullscreen
-        to mimic the user interface of another application.
+        The `display` member allows an origin some measure of control over a
+        user agent’s native UI. After taking over the full screen, it could
+        attempt to mimic the user interface of another application. This is
+        also facilitated by the `'display-mode'` media feature
+        [[MEDIAQUERIES-5]], through which a script can know the display mode of
+        a web application.
       </p>
     </section>
     <section class="appendix">
--- a/index.html
+++ b/index.html
@@ -1051,8 +1051,8 @@
           </li>
         </ol>
         <p>
-          A user agent SHOULD register the file handlers with the operating
-          system via interactions that are consistent with:
+          On [=installation=], a user agent SHOULD register the file handlers
+          with the operating system via interactions that are consistent with:
         </p>
         <ul>
           <li>The [=file handler/icon=] and [=file handler/name=] appear in the
@@ -1072,6 +1072,53 @@
           the 'default' file handler for a given type if the operating system
           allows.
         </p>
+        <section data-cite="file-system-access">
+          <h2>
+            <dfn>Execute a file handler launch</dfn>
+          </h2>
+          <p>
+            This is the IDL used for executing a file handling launch:
+          </p>
+          <pre class="idl">
+            partial interface Window {
+              readonly attribute LaunchQueue launchQueue;
+            };
+
+            callback LaunchConsumer = any (LaunchParams params);
+
+            [Exposed=Window] interface LaunchQueue {
+              undefined setConsumer(LaunchConsumer consumer);
+            };
+
+            [Exposed=Window] interface LaunchParams {
+              readonly attribute FrozenArray&lt;FileSystemHandle&gt; files;
+            };
+          </pre>
+          <p>
+            When a [=file type=] is registered with a web app and one or more
+            of these registered files are launched on the platform, run the
+            following steps, given the launched {{FrozenArray}} of
+            {{FileSystemHandle}}s |files:FrozenArray&lt;FileSystemHandle&gt;|
+            and the corresponding [=file handler|file_handler=].
+          </p>
+          <ol>
+            <li>Let |url| be the |file_handler|.[=file handler/action=].
+            </li>
+            <li>Let |browsing context| be the result of creating a new
+            [=top-level browsing context=].
+            </li>
+            <li>[=Navigate=] |browsing context| to |url|.
+            </li>
+            <li>XXX Wait for navigation to be complete?
+            </li>
+            <li>Let |params:LaunchParams| be {{LaunchParams}} with
+            {{LaunchParams.files}} set to |files|.
+            </li>
+            <li>Iterate through every {{LaunchConsumer}} set by
+            {{LaunchQueue/setConsumer}} and execute the function with |params|.
+            </li>
+          </ol>
+        </section>
         <section>
           <h3>
             Privacy consideration: Default [=file handler=].
@@ -1915,9 +1962,8 @@
           "file handler">action</dfn></code> member is a <a>string</a> that
           represents a relative URL of the [=manifest/start_url=] origin that
           is [=manifest/within scope=] of a [=Document/processed manifest=].
-          This URL will be navigated to in the [=launch event=] when handling a
-          file. See the process to [=execute a file handler launch=] for more
-          information.
+          This URL will be navigated to in the steps to [=execute a file handler
+          launch=].
         </p>
       </section>
       <section>
@@ -2041,22 +2087,6 @@
           <li>Return |file_handler|.
           </li>
         </ol>
-      </section>
-      <section>
-        <h2>
-          Executing a file handler launched
-        </h2>
-        <p>
-          To <dfn>execute a file handler launch</dfn>, XXX fill out
-        </p>
-        <section>
-          <h3>
-            Launch Event
-          </h3>
-          <p>
-            A <dfn>launch event</dfn> is ....
-          </p>
-        </section>
       </section>
     </section>
     <section>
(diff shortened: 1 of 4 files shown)


dmurph:
@mkruisselbrink  and @inexorabletash can you PTAL for structural & spec-language feedback?

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
         </p>
       </section>
+      <section>
+        <h3>
+          `file_handlers` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">`file_handlers`</dfn></code> member is a [=list=] that
+          represents urls within the app that can handle launches of a given
```
tomayac, line 1022:
```suggestion
          represents URLs within the app that can handle launches of a given
```

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          file type (TODO: link to definition of file type).
+        </p>
+        <p>
+          How file handlers are registered and disambiguated are at the
```
tomayac, line 1026:
```suggestion
          How file handlers are registered and disambiguated is at the
```

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+            </ol>
+          </li>
+        </ol>
+        <p>
+          A user agent SHOULD register the file handlers with the operating
+          system via interactions that are consistent with:
+        </p>
+        <ul>
+          <li>The [=file handler/icon=] and [=file handler/name=] appearing in
+          an "Open with.." menu in the context menu of associated files.
```
tomayac, line 1059:
```suggestion
          an "Open with…" menu in the context menu of associated files.
```

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          <li>The [=file handler/icon=] appearing as the icon of associated
+          files.
+          </li>
+          <li>File types with multiple possible handlers can be disambiguated
+          by UX provided by the operating system.
+          </li>
+        </ul>
+        <p>
+          A user agent MAY (TODO: should?) prevent registered file handlers
+          from registering as the 'default' file handler for a give type if the
```
tomayac, line 1070:
```suggestion
          from registering as the 'default' file handler for a given type if the
```

tomayac (commented):
Just a quick drive-by review…

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+              <li>If |file_handler| is failure, continue.
+              </li>
+              <li>[=list/Append=] |file_handler| to |processedFileHandlers|.
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p>
+          A user agent SHOULD register the file handlers with the operating
+          system via interactions that are consistent with:
```
inexorabletash, line 1055: While I like lists, this phrasing makes the bullets into sentence fragments. Can you reword the bullets so they're more complete sentences?

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+        </p>
+        <ul>
+          <li>The [=file handler/icon=] and [=file handler/name=] appearing in
```
inexorabletash, line 1058: It's more spec-like to describe the abstract concept, then give examples. e.g.

"The icon and name appear in the host operating system surfaces presenting options for opening files, for example within "Open with..." context menus for associated file types."

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          from registering as the 'default' file handler for a give type if the
+          operating system allows.
+        </p>
+      </section>
+      <section>
+        <h3>
+          Privacy consideration: Default [=file handler=].
+        </h3>
+        <p>
+          Depending on the operating system capabilities that the
```
inexorabletash, line 1079: Grammar: I think you want: "Depending on the operating system capabilities, the file handler could..."

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          [=manifest/file handler=] could become a 'default' handler (e.g.
+          handling launches of a given file type automatically) of a given file
+          type (todo define? link?) or [=mime type=] without the explicit
+          knowledge of the user, depending on the operating system. To protect
```
inexorabletash, line 1083: ... and don't also end the sentence with "depending on the operating system."

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+      </section>
+      <section>
+        <h3>
+          Privacy consideration: Default [=file handler=].
+        </h3>
+        <p>
+          Depending on the operating system capabilities that the
+          [=manifest/file handler=] could become a 'default' handler (e.g.
+          handling launches of a given file type automatically) of a given file
+          type (todo define? link?) or [=mime type=] without the explicit
```
inexorabletash, line 1082: Maybe go ahead and create a stub definition for "file type" in this doc, and put a TODO there, then have these links point at it?

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          knowledge of the user, depending on the operating system. To protect
+          against possible mis-use, user agents MAY choose to offer the user
+          the functionality to not launch the web application, or even
+          unregister a given file handler for a web application.
+        </p>
+      </section>
+      <section>
+        <h3>
+          Privacy consideration: Name & icon.
```
inexorabletash, line 1091: nit: I'd spell out "and"

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+        </h3>
+        <p>
+          The name and icon of each file handler can be sensitive to privacy
+          and security, as there isn't a specified way for the user to see &
```
inexorabletash, line 1095: nit: spell out "and"

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
+          confirm these. Due to this, user agents MAY choose to replace these
+          with name and icons derived from the application's icon and name in
```
inexorabletash, line 1097: consistency: "icon" above vs. "icons" here.

index.html
```diff
@@ -1118,6 +1204,9 @@ <h3>
             <li>[=Process the `shortcuts` member=] passing |json|, |manifest|,
             and |manifest URL|.
             </li>
+            <li>[=Process the `file_handlers` member=] passing |json|,
```
inexorabletash, line 1207: nit: ", and" or just "and"

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
         </ol>
       </section>
     </section>
+    <section>
+      <h2>
+        File Handler Items
+      </h2>
+      <p>
+        Each <dfn data-local-lt="file handler's">file handler</dfn> represents
```
inexorabletash, line 1884: Does ReSpec require the lt? Can you put the `'s` outside the link where used instead?

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+        a url in the scope of the application that can handle launches with
```
inexorabletash, line 1885: nit: URL

inexorabletash, line 1885: Maybe just "handles launches" instead of "can handle launches" ? Less passive voice, I think.

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+        <li>[=file handler/accepts=]
+        </li>
+        <li>[=file handler/icons=]
+        </li>
+      </ul>
+      <p>
+        A user agent can use these members to associate the web application
+        with specific file extensions or [=mime types=] on the operating
+        system. When a user launches this association When a user launches a
+        file that is associated with A user agent can use these members to
```
inexorabletash, line 1902: Looks like the first sentence doesn't get finished?

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+        <li>[=file handler/name=]
+        </li>
+        <li>[=file handler/accepts=]
+        </li>
+        <li>[=file handler/icons=]
+        </li>
+      </ul>
+      <p>
+        A user agent can use these members to associate the web application
+        with specific file extensions or [=mime types=] on the operating
```
inexorabletash, line 1900: This might be a good candidate to hoist into a definition of "file type" (covering extensions + mime types) and then linking from here and elsewhere.

index.html
inexorabletash: FWIW, GitHub is rendering @tomayac's suggestion as a skinny ellipsis character (`…`) rather than just three periods (`...`). I'm not sure which was intended.

inexorabletash: ... but be consistent in the doc.

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+        </p>
+      </section>
+      <section>
+        <h3>
+          `icons` member
+        </h3>
+        <p>
+          The <a>file handler</a>'s <code><dfn data-dfn-for=
+          "file handler">icons</dfn></code> member lists images that serve as
+          iconic representations of file. This MAY be replaced by the user
```
inexorabletash, line 1960: "files" or "file types"

Also, maybe replace "iconic" with "graphical" or "visual" ?

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+          [=URL=] |manifest URL:URL|.
+        </p>
+        <ol class="algorithm">
+          <li>Return failure if it's the case that:
+            <ul>
+              <li>|item| is not [=ordered map=].
+              </li>
+              <li>|item|["action"] doesn't [=map/exist=].
+              </li>
+              <li>|item|["action"] is not a [=string=].
```
inexorabletash, line 1980: WDYT about combining these pairs, e.g. "item["action"] exists and is a string" ?

index.html
```diff
@@ -1787,6 +1876,172 @@ <h2>
...
+              </li>
+              <li>|item|["name"] doesn't [=map/exist=].
+              </li>
+              <li>|item|["name"] is the empty string.
+              </li>
+              <li>|item|["accepts"] doesn't [=map/exist=].
+              </li>
+              <li>|item|["accepts"] is not a [=sequence=].
+              </li>
+              <li>|item|["accepts"] is an empty [=sequence=].
```
inexorabletash, line 1990: Linkify empty

inexorabletash (commented):
Initial feedback. I didn't get all the way to the end, LMK when it's ready for another look.

index.html
```diff
@@ -1012,6 +1012,92 @@ <h3>
...
       </section>
+      <section>
+        <h3>
+          `file_handlers` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">`file_handlers`</dfn></code> member is a [=list=] that
+          represents urls within the app that can handle launches of a given
+          file type (TODO: link to definition of file type).
```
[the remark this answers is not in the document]
dmurph: This is just my own TODO for things I need to do in this pull request :)

index.html
dmurph: I'm matching how it works for shortcut items in this spec here.

index.html
dmurph: I put it in a 'concepts' section. WDYT?

index.html
dmurph: Not sure there is something I can link to? only see list/empty

inexorabletash: Ah, okay, there's subtlety here. A "sequence" is a WebIDL idea. item and the stuff inside it are spec concepts, not IDL things. So this should be a list, not a sequence.

dmurph:
Alright I think I have it kind of all in place? The execution algorithm isn't good yet - but I think this is the right structure?

@inexorabletash PTAL again? @evanstade PTAL?

index.html
```diff
@@ -1787,6 +1942,147 @@ <h2>
...
+          </li>
+          <li>Let |file_handler:ordered map| be |ordered map| «[ "action" →
+          |url|, "name" → |item|["name"] ]».
+          </li>
+          <li>[=Process image resources=] passing |item|["icons"],
+          |file_handler|, |manifest URL|, and "icons".
+          </li>
+          <li>The user agent MAY replace |file_handler|["icons"] and
+          |file_handler|["name"] with the application icon and a string
+          containing the application name, respectively.
```
evanstade, line 2079: The user agent should not replace `name` with the application's name. It may fall back to something else, such as "\<Application name\> file", but the application name by itself is a bad fallback.

Also this fallback is more intended for a missing `name`; I don't think there's a reason to ever replace the name provided by the app.

index.html
```diff
@@ -1012,6 +1012,158 @@ <h3>
...
+        <ul>
+          <li>The [=file handler/icon=] and [=file handler/name=] appear in the
+          host operating system surfaces presenting options for opening files,
+          for example within "Open with..." context menus for associated file
+          types.
+          </li>
+          <li>The [=file handler/icon=] appears as the icon of associated files
+          in the host operating system.
+          </li>
+          <li>File types with multiple possible handlers can be disambiguated
```
evanstade, line 1084: I think instead of "disambiguated" I would choose a different word here.

The OS allows the user to *select* which app should open a file, and also allows the user to set a default app.

index.html
```diff
@@ -1012,6 +1012,158 @@ <h3>
...
+          </p>
+          <ol>
+            <li>Let |url| be the |file_handler|.[=file handler/action=].
+            </li>
+            <li>Let |browsing context| be the result of creating a new
+            [=top-level browsing context=].
+            </li>
+            <li>[=Navigate=] |browsing context| to |url|.
+            </li>
+            <li>XXX Wait for navigation to be complete?
```
evanstade, line 1130: We do wait for the navigation to "complete", whatever that means technically.

Currently we also have to wait for the user to approve the action (at least sometimes).

However, we're interested in changing this such that user approval, if any, occurs before the creation of the browsing context and subsequent navigation. This sidesteps concerns like what happens after a navigation has occurred, but the user has not allowed the open-file action. It sounds like you're describing this potential future improvement rather than the current state of Chrome. Is that accurate?

index.html
```diff
@@ -1012,6 +1012,158 @@ <h3>
...
+        <section>
+          <h3>
+            Privacy consideration: Default [=file handler=].
+          </h3>
+          <p>
+            Depending on the operating system capabilities, the [=manifest/file
+            handler=] could become a 'default' handler (e.g. handling launches
+            of a given file type automatically) of a given [=file type=]
+            without the explicit knowledge of the user. To protect against
+            possible mis-use, user agents MAY choose to offer the user the
```
evanstade, line 1149: maybe,

> MAY require explicit user consent to proceed with the launch action when an app is launched by handling a file.

index.html
```diff
@@ -1012,6 +1012,158 @@ <h3>
...
+            functionality to not launch the web application, or even unregister
+            a given file handler for a web application.
```
evanstade, line 1151: > If consent is denied, the user agent MAY remove the web application's OS registration as a file handling entity.

evanstade (changes requested):

index.html
```diff
@@ -1787,6 +1942,147 @@ <h2>
...
+          strings that start with a "." and only contain [=valid suffix code
+          points=]. Additionally extensions are limited to a length of 16 code
+          points.
+        </p>
+        <p>
+          In addition to complete MIME types, "*" can be used as the subtype of
+          a MIME type to match, for example, all image formats with "image/*".
+        </p>
+        <p>
+          Websites SHOULD always provide both [=MIME types=] and file
```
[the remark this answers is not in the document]
dmurph: Let me know what you think now

index.html
dmurph: I added some notes about what we might want to do here instead - let me know what you think.

index.html
dmurph: Done.

dmurph:
@marcoscaceres do you think you could give this a review? I'm currently not sure about:

1. How to do the algorithm for execution of the file handler - I put some thoughts there
2. How to cite the mimetype standard (currently just linking to the RFC)

Do you have some cycles to take a look?

Thanks
Daniel

index.html
dmurph: PTAL

dmurph:
@evanstade @inexorabletash @tomayac @marcoscaceres PTAL :)

index.html
```diff
@@ -1012,6 +1019,144 @@ <h3>
...
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">`file_handlers`</dfn></code> member is a [=list=] that
+          provides instructions for how the app handles file-opening actions
+          that originate outside of the app itself.
+        </p>
+        <p>
+          How file handlers are registered and disambiguated is at the
+          discretion of the operating system.
```
evanstade, line 1034: It's not clear to me what this sentence is intended to convey. I guess it means that control over which app is used to open a file is an OS-level setting, not a user agent setting? If that's the case, maybe something like:

> The management, presentation and selection of registered file-handling applications is at the discretion of the operating system.

index.html
```diff
@@ -1012,6 +1012,158 @@ <h3>
...
+              </li>
+              <li>If |file_handler| is failure, continue.
+              </li>
+              <li>[=list/Append=] |file_handler| to |processedFileHandlers|.
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p>
+          On [=installation=], a user agent SHOULD register the file handlers
```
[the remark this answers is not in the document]
evanstade: @marcoscaceres wdyt?

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+        </p>
+        <p>
+          In addition to complete [=MIME types=], "*" can be used as the
+          subtype of a MIME type to match, for example, all image formats with
+          "image/*". However, the top-level type MUST be defined in [[rfc2046]]
+          and listed in [[IANA_MEDIA_TYPES]].
+        </p>
+        <p>
+          Websites MUST always provide both [=MIME types=] and file extensions
+          for each [=file handler/accept=] entry. This format is required by
```
evanstade, line 2032: So technically, the file extensions don't matter on Linux (as long as the extensions and mime type are standard, like .csv and text/plain). Whereas on Windows, the MIME type shouldn't matter. So I would say this format is required to support the variety of operating systems, each of which may rely on MIME /or/ extension.

It's also required for custom extensions/MIMEs to tell Linux that a file ending in ".foo" is application/foo.

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
+          "file handler">accept</dfn></code> member is a <a>dictionary</a>
+          mapping [=MIME types=] to a list of extensions. Extensions have to be
+          strings that start with a "." and only contain [=valid suffix code
+          points=]. Additionally extensions are limited to a length of 16 code
+          points.
+        </p>
+        <p>
+          In addition to complete [=MIME types=], "*" can be used as the
```
evanstade, line 2025: While the wildcard in the subtype would work in some OSes, on others only the extension matters. Thus it's probably worth noting that

`"image/*": [ ".png" ]`

would /not/ handle all image files on all OSes. (In fact there's no enforcement of these two lists having anything to do with each other, so

`"image/png": [ ".txt" ]`

would be accepted; it would just behave differently across OSes.

index.html
```diff
@@ -1012,6 +1019,144 @@ <h3>
...
+                      "src": "/csv-file.png",
+                      "sizes": "144x144"
+                    }
+                  ]
+                },
+                {
+                  "action": "/open-svg",
+                  "name": "Image",
+                  "accept": {
+                    "image/svg+xml": ".svg"
```
evanstade, line 1136: nit: the file extension needs to be in a list

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+          </li>
+          <li>If |url| is not [=manifest/within scope=] of |scope|, return
+          failure.
+          </li>
+          <li>Let |accept:ordered map| be an [=ordered map=].
+          </li>
+          <li>[=map/iterate=] |mime_type_string:string| → |extensions| of
+          |item|["accept"]
+            <ol>
+              <li>If |extensions| is not a [=string=] or [=list=], return
```
evanstade, line 2089: I don't think we intend to support a string that's not in a list?

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/manifest/manifest_parser.cc;l=1048-1057;drc=6fc55b5ee6ccec48d2e0ee47fb0e4bed0e1f8040

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+            </ol>
+          </li>
+          <li>Let |file_handler:ordered map| be |ordered map| «[ "action" →
+          |url|, "name" → |item|["name"], "accept" → |accept| ]».
+          </li>
+          <li>[=Process image resources=] passing |item|["icons"],
+          |file_handler|, |manifest URL|, and "icons".
+          </li>
+          <li>The user agent MAY replace |file_handler|["icons"] with the
+          application icon.
```
evanstade, line 2115: Should it be this specific?

One proposal that has floated around before has been to use the application icon badged with the file type icon.

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+        A {{Window/window}} has an associated {{LaunchConsumer}} <dfn>assigned
+        launch consumer</dfn>.
+      </p>
+      <p>
+        A {{Window/window}} has an associated [=list=] of {{LaunchParams}}
+        <dfn>unconsumed launch params</dfn>.
+      </p>
+      <p>
+        When a [=file type=] is registered with a web app and one or more of
+        these registered files are launched on the platform, run the following
```
evanstade, line 2139: these steps are also run after every same-origin page navigation (this supports refreshing the page or redirecting)

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+            readonly attribute FrozenArray&lt;FileSystemHandle&gt; files;
+          };
+        </pre>
+        <section>
+          <h2>
+            <dfn>files</dfn> attribute
+          </h2>
+          <p>
+            The {{LaunchParams/files}} attribute is a {{FrozenArray}} of
+            {{FileSystemHandle}}s.
```
evanstade, line 2176: worth noting these handles are read only? (but can be converted to readWrite via requestPermission)

evanstade (changes requested):

index.html
```diff
@@ -1787,6 +1935,296 @@ <h2>
...
+          mapping [=MIME types=] to a list of extensions. Extensions have to be
+          strings that start with a "." and only contain [=valid suffix code
+          points=]. Additionally extensions are limited to a length of 16 code
+          points.
+        </p>
+        <p>
+          In addition to complete [=MIME types=], "*" can be used as the
+          subtype of a MIME type to match, for example, all image formats with
+          "image/*". However, the top-level type MUST be defined in [[rfc2046]]
+          and listed in [[IANA_MEDIA_TYPES]].
```
marcoscaceres, line 2028: This should be enforced in some algorithm (if it isn't already), but not as a MUST requirement here.

index.html
dmurph: Agreed. I'll just say something vague.

index.html
dmurph: Done

index.html
dmurph: Hm.... @inexorabletash or @marcoscaceres I need some spec help here - how do you think I should specify this? do the above "assigned launch consumer" and "unconsumed launch params" live across same-origin redirects?

index.html
dmurph: @mkruisselbrink I tried to state this in a way that made sense, but you'll know better how to specify this.

dmurph (commented):
Thanks!

index.html
dmurph: Note from @mkruisselbrink - I should use language from https://wicg.github.io/file-system-access/#local-file-system-permissions

index.html
inexorabletash: I think you'll [browsing session](https://html.spec.whatwg.org/multipage/history.html#browsing-sessions) is where you'll need to have the launch params live (rather than window), and then a window has access to the params via its browsing session. 

I have low confidence in my answer here, maybe @jakearchibald can share some wisdom?

index.html
dmurph: Done.

index.html
```diff
@@ -1787,6 +1941,303 @@ <h2>
...
+        launch consumer</dfn>.
+      </p>
+      <p>
+        A {{Window/window}} has an associated [=list=] of {{LaunchParams}}
+        <dfn>unconsumed launch params</dfn>.
+      </p>
+      <p>
+        When a [=file type=] is registered with a web app and one or more of
+        these registered files are launched on the platform, run the following
+        steps, given the launched {{FrozenArray}} of {{FileSystemHandle}}s
```
jakearchibald, line 2149: It's unclear where this object comes from (eg what JS realm)

index.html
```diff
@@ -1787,6 +1941,303 @@ <h2>
...
+      <h2>
+        <dfn>Execute a file handler launch</dfn>
+      </h2>
+      <p>
+        A {{Window/window}} has an associated {{LaunchConsumer}} <dfn>assigned
+        launch consumer</dfn>.
+      </p>
+      <p>
+        A {{Window/window}} has an associated [=list=] of {{LaunchParams}}
+        <dfn>unconsumed launch params</dfn>.
```
jakearchibald, line 2144: What causes this list to contain more than one item?

jakearchibald (changes requested):

index.html
evanstade: After some more thought, I think this makes more sense if it's based on the app scope. That is, any navigation or reload that remains within app scope gets the file handles. (Originally I was thinking of origin because it's the standard security boundary, but if we're supporting scope extensions then I suppose the security boundary ought to be a set of origins.)

dmurph: @jakearchibald do you have thoughts?

@alancutter does the existing spec here generally work for you other than the navigation scope part?

alancutter: I wrote a small [doc](https://docs.google.com/document/d/1EeweceXvS55VLzj-YYgUPIa8sq3vHnYEcy5yvoYX54A/edit) for removing the re-enqueuing on navigation implementation currently in Chrome.

The LaunchQueue/LaunchParams API surface LGTM otherwise.

index.html
```diff
@@ -1787,6 +1951,310 @@ <h2>
...
+          <h2>
+            <dfn>files</dfn> attribute
+          </h2>
+          <p>
+            The {{LaunchParams/files}} attribute is a {{FrozenArray}} of
+            {{FileSystemHandle}}s. For every |file handle:FileSystemHandle| in
+            {{LaunchParams/files}}, the [=permission state=] for a descriptor
+            with {{FileSystemPermissionDescriptor/handle}} set to the |file
+            handle| and {{FileSystemPermissionDescriptor/mode}} set to
+            "{{FileSystemPermissionMode/"read"}} should be "granted".
```
evanstade, line 2204: this has been changed to readwrite: crbug.com/1264489

marcoscaceres (commented):
Couple of drive-by comments...

index.html
```diff
@@ -1012,6 +1019,160 @@ <h3>
...
+        <section>
+          <h3>
+            Privacy consideration: Default [=file handler=].
+          </h3>
+          <p>
+            Depending on the operating system capabilities, the [=manifest/file
+            handler=] could become a 'default' handler (e.g. handling launches
+            of a given file type automatically) of a given [=file type=]
+            without the explicit knowledge of the user. To protect against
+            possible mis-use, user agents MAY require explicit user consent to
```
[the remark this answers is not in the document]
evanstade: In general, triggering this API requires user action at the OS level, e.g. `file browser > file > context menu > open with > app`. There's a similar amount of user interaction as with `<input type="file">`. If a user agent is careful about how it presents the app in OS level surfaces and depending on details of the OS (such as whether it might silently make a new app the default handler for a file type) then this added confirmation step is arguably not necessary.

That said, the current plan for Chromium **is** to show this extra user consent step.

index.html
```diff
@@ -1787,6 +1951,310 @@ <h2>
...
+        </h3>
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
+          "file handler">accept</dfn></code> member is a <a>dictionary</a>
+          mapping [=MIME types=] to a list of [=file extensions=].
+        </p>
+        <p>
+          In order to [=register file handlers=], some operating systems
+          require [=MIME types=], some require [=file extensions=], and some
+          support both. Thus websites MUST always provide both for each [=file
```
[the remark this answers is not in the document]
evanstade: Right, so, if an app used:

`  'image/jpeg': ['.txt']`

then it would work with .txt files on Windows and not work with anything on Linux.

> can the file extension be made to win on Linux

No, because Linux (`xdg-mime`) only associates mime types to apps. If we tried to use some custom MIME type, e.g. telling `xdg-mime` that this app supports `application/x-appid-type1` the problem would be that we couldn't convince `xdg-mime` to map `foo.txt` to this custom MIME type; it would still map it to `text/plain`.

index.html
```diff
@@ -1787,6 +1942,147 @@ <h2>
...
+      <section>
+        <h3>
+          `name` member
+        </h3>
+        <p>
+          The [=file handler's=] <code><dfn data-dfn-for=
+          "file handler">name</dfn></code> member is a <a>string</a> that
+          represents a name to be provided to the user's operating system for
+          use with "Open with..." UX flows. This MAY be replaced by the user
+          agent for privacy & security reasons.
```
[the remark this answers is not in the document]
evanstade: fiddled with this section a bit.

index.html
```diff
@@ -1787,6 +1951,310 @@ <h2>
...
+        </p>
+      </section>
+      <section data-cite="MIMETYPE">
+        <h3>
+          `accept` member
+        </h3>
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
```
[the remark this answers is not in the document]
evanstade: Yes. This is why the file extensions are required. I've attempted to clarify the wording here.

The goal is greater consistency of user experience across OSes and user agents.

Because certain operating systems require mime types for the association (linux via xdg-mime), we enforce that site authors must provide mime types. Any attempt by the user agent to smartly map from extension to mime type or vice versa would cause different user agents to behave differently since the mappings would not be identical.

However, it's also enforced that the file extension matches what's provided in the manifest. This may happen at the OS level or at the user agent level. For example, `"text/*": [".txt"]` would not match a `.csv` file anywhere, even on Linux where the file is recognized to match that MIME type. This would cause different OSes to behave differently.

index.html
```diff
@@ -1787,6 +1941,303 @@ <h2>
...
+            <li>|item|["accept"] is [=list/empty].
+            </li>
+          </ul>
+        </li>
+        <li>Let |url:URL| be the result of [=URL Parser|parsing=]
+        |item|["action"] with |start_url URL| as the base URL.
+        </li>
+        <li>If |url| is failure, return failure.
+        </li>
+        <li>If |url| is not [=manifest/within scope=] of |scope|, return
```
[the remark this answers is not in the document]
evanstade: seems to me it takes a URL

index.html
```diff
@@ -1787,6 +1941,303 @@ <h2>
...
+            <li>If |extensions| is an [=list/empty=] [=list=], return failure.
+            </li>
+            <li>If |extensions| is a [=list=] with items that are not
+            [=string=]s, return failure.
+            </li>
+            <li>Let |mime_type_parsed:mime type| be the result of running the
+            steps of [=parse a mime type=] on |mime_type_string|.
+            </li>
+            <li>If |mime_type_parsed| is not defined in [[rfc2046]], return
+            failure.
```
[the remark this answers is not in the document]
evanstade: I've removed the rfc reference. The parse a mime type algorithm and the IANA-MEDIA-TYPES reference should be sufficient.

index.html
evanstade: updated

evanstade (commented):
Hello @marcoscaceres @jakearchibald Daniel and I have done some load balancing and I will be working on this PR going forward. PTAL the latest version. I have attempted to respond to all outstanding comments.

evanstade:
Hi @marcoscaceres and @jakearchibald --- the PR has been updated and the sections on launching a web app with file handles is now also ready for review. Our intent is to leave the launchqueue/launchparams section in this spec for now and move it into a different doc while preparing the spec changes for `launch_handler`.

index.html
```diff
@@ -1209,6 +1209,148 @@ <h3>
           conventions or limitations of the host operating system.
         </p>
       </section>
+      <section>
+        <h3>
+          `file_handlers` member
+        </h3>
+        <p>
+          The [=manifest's=] <code><dfn data-export="" data-dfn-for=
+          "manifest">`file_handlers`</dfn></code> member is a [=list=] that
```
marcoscaceres, line 1218:
```suggestion
          "manifest">file_handlers</dfn></code> member is a [=list=] that
```

index.html
```diff
@@ -1209,6 +1209,148 @@ <h3>
...
+          provides instructions for how the app handles file-opening actions
+          that originate outside of the app itself.
+        </p>
+        <p>
+          The management, presentation and selection of registered
```
marcoscaceres, line 1223:
```suggestion
          The management, presentation, and selection of registered
```

index.html
```diff
@@ -1012,6 +1019,160 @@ <h3>
...
+        <p>
+          To <dfn>process the `file_handlers` member</dfn>, given [=ordered
+          map=] |json:ordered map|, [=ordered map=] |manifest:ordered map|,
+          |manifest_url::URL|:
+        </p>
+        <ol class="algorithm">
+          <li>Let |processedFileHandlers:list| be a new [=list=].
+          </li>
+          <li>Set |manifest|["file_handlers"] to |processedFileHandlers|.
+          </li>
```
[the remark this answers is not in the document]
marcoscaceres: > That leaves |manifest|["file_handlers"] as uninitialized if there's an early return. Is that OK?

Yeah, good call. I wasn't expecting an early return.

index.html
```diff
@@ -1012,6 +1019,160 @@ <h3>
...
+              </li>
+              <li>If |file_handler| is failure, continue.
+              </li>
+              <li>[=list/Append=] |file_handler| to |processedFileHandlers|.
+              </li>
+            </ol>
+          </li>
+        </ol>
+        <p>
+          On [=installation=], a user agent SHOULD run the process to
```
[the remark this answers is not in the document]
marcoscaceres: > Can you clarify how you suggest the text here should change?

If we wanted to get fancy, we could have throw these into some kind of "steps to run on installation" which would have a queue of actions to run. The intent here is clear enough for now though.   

>  The user agent might not register these associations with the OS at installation time, but pop up a toast when the app is later launched to confirm the user wants the associations.

Right, I see...

index.html
```diff
@@ -1012,6 +1019,160 @@ <h3>
...
+          [=register file handlers=]
+        </p>
+        <section>
+          <h3>
+            Registering file handlers
```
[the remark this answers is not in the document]
marcoscaceres: Thanks for the clarification, @evanstade. This may become contentious as it may be that there are certain file types that a web application might never be allowed to handle (e.g., system files)... but this may be a UA-level restriction or os-level.

So, question is: would we allow registering anything? or would it be limited to a few well-known types (which yes, massively restricts the utility of this).

index.html
```diff
@@ -1209,6 +1209,148 @@ <h3>
...
+            native file browser and other surfaces that display lists of apps
+            that can handle a given file or file type.
+            </li>
+            <li>The app is a candidate for becoming the system's default
+            handler for associated file types.
+            </li>
+          </ul>
+          <p>
+            A user agent MAY truncate the total set of [=file extensions=] to
+            preserve functionality and prevent abuse.
```
marcoscaceres, line 1276: I wonder if we should explicitly ban some types (e.g., .sys, .exe, etc.)?

marcoscaceres: actually, it would probably be enough to say that the UA may prevent certain file associations...

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+        with [=file type=] on the operating system.
+      </p>
+      <p>
+        A <dfn>file type</dfn> can be defined by a [=mime type=] and/or [=file
+        extension=]. A file belongs to a file type if the OS determines it to
+        have a mime type and/or its name ends with a certain [=file
+        extension=]. A <dfn>file extension</dfn> is a string that start with a
+        "." and only contains [=valid suffix code points=]. A <dfn>valid suffix
+        code point</dfn> is a [=code point=] that is [=ASCII alphanumeric=],
+        U+002B (+), or U+002E (.). Additionally, [=file extensions=] are are
```
marcoscaceres, line 2165:
```suggestion
        U+002B (+), or U+002E (.). Additionally, [=file extensions=] are
```

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+        limited to a length of 16 code points.
```
marcoscaceres, line 2166: Just wondering, what's are the restrictions in this section based on?

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+      </p>
+      <aside class="note">
+        <p>
+          Note: These code points were chosen to support most pre-existing file
```
marcoscaceres, line 2170:
```suggestion
          These code points were chosen to support most pre-existing file
```

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+          formats. The vast majority of file extensions are purely
+          alphanumeric, but compound extensions (such as .tar.gz) and
+          extensions such as .c++ for C++ source code are also fairly common,
+          hence the inclusion of + and . as allowed code points.
+        </p>
+      </aside>
+      <aside class="note">
+        <p>
```
marcoscaceres, line 2178: This note sounds somewhat normative....

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+        </p>
+      </aside>
+      <section>
+        <h3>
+          `action` member
+        </h3>
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
+          "file handler">action</dfn></code> member is a <a>string</a> that
+          represents a relative URL of the [=manifest/start_url=] origin that
```
marcoscaceres, line 2191: This should probably be resolved relative to the manifest's URL instead (like other URLs in the manifest). 

```suggestion
          represents a relative URL of the [=manifest/start_url=] origin that
```

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+          handler launch=].
+        </p>
+      </section>
+      <section>
+        <h3>
+          `name` member
+        </h3>
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
+          "file handler">name</dfn></code> member is a <a>string</a> that
```
marcoscaceres, line 2203: I think this should work: 

```suggestion
          "manifest/file_handler">name</dfn></code> member is a <a>string</a> that
```

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+          describes the file type. User agents MAY pass this information to the
```
marcoscaceres, line 2204: Something doesn't seem right here (might just be the wording)... it's a "name" that "describes". 

Also, I'm a bit unsure where this gets shown to a user? Is there potential abuse here? Like `"name": "Pictures"` and `"type": "application/whatever"`.

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+          operating system during file handler registration.
+        </p>
+        <aside class="note">
+          The [=file handler/name=] is used to describe the file type in a
+          human readable format, and is typically only displayed in OS surfaces
+          (such as a file browser) if the web app has become the default
+          handler for that file type. It's most useful for custom file types.
```
marcoscaceres, line 2211:
```suggestion
          <p>
          The [=file handler/name=] is used to describe the file type in a
          human readable format, and is typically only displayed in OS surfaces
          (such as a file browser) if the web app has become the default
          handler for that file type. It's most useful for custom file types.
          </p>
```

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+        <h3>
+          `launch_type` member
+        </h3>
+        <p>
+          The [=file handler=]'s <code><dfn data-dfn-for=
+          "file handler">launch_type</dfn></code> member is a <a>string</a>
+          that determines how the app is launched for files routed to this
+          handler. The possible values are "single-client" and
+          "multiple-clients". If not provided, the default behavior matches
+          "single-client".
```
marcoscaceres, line 2275:
```suggestion
          "multiple-clients". If not provided, it defaults to
          "single-client".
```

marcoscaceres: multiple-clients and single-client should probably link to their definitions and behavior.

index.html
evanstade: > Something doesn't seem right here (might just be the wording)... it's a "name" that "describes".

Being descriptive is a property of a good name, I think :) But I shall change this to "identifies".

> Also, I'm a bit unsure where this gets shown to a user?

The non-normative note right below describes where it's shown to the user. Do you have suggestions on how to make that text more clear?

> Is there potential abuse here? Like "name": "Pictures" and "type": "application/whatever"

I can't see a security threat here, but then it's not my specialty. We have, however, subjected this API to several rounds of internal security review and while spoofing in general is a keen concern, the name is not as much of a focus as the icon. I left the wording here more relaxed ("MAY") to accommodate that, depending on the level of trust the user agent places in the app. For example, a user agent with a lot of ceremony around installation may trust installed apps enough to display the name and icon, whereas a user agent where installation is totally frictionless may not want to display name and icon at all.

index.html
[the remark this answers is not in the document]
evanstade: I don't think there's any _file type_ that we wouldn't want to allow access to. Access control should be applied to individual files that a user shouldn't be able to edit, as you say, by the OS. If a user is allowed to edit a file, I think there's a business justification for wanting to do it via a PWA. If we block specific file types from being edited, we run the risk of forcing apps/users to other ecosystems which don't have this restriction and are much less secure overall. And the greatest threat to user security/privacy is probably in the divulgence of PII or other secrets, which are more likely to be in .doc or .txt or .xlsx or .pdf than some obscure type --- these are clearly types we need to support.

index.html
evanstade: done.

evanstade (commented):
Marcos, thank you for the thoughtful and constructive feedback. I've updated the PR, but due to the lack of signals from other vendors, for now we'll put this in manifest incubations.

index.html
```diff
@@ -1987,6 +2132,432 @@ <h2>
...
+                </ol>
+              </li>
+            </ol>
+          </li>
+        </ol>
+      </section>
+    </section>
+    <section data-cite="file-system-access permissions">
+      <h2>
+        Launch queue and launch params
```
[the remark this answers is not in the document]
alancutter: By having a persistent launchQueue we avoid racing with JS event registration during page load.

CodeDoctorDE:
Hi,
why was this pr closed?

dmurph:
File handlers have been put into WICG in manifest-incubations:
https://github.com/WICG/manifest-incubations

PR #1011, opened 2021-09-29
Describe manifest update behavior

### LINKED ISSUE #446: Specify how updates work

philloooo:
closes #446 
This change is dependent on https://github.com/w3c/manifest/pull/988

Commit message:

difference will be.)

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/manifest/pull/1011.html" title="Last updated on Nov 2, 2021, 10:00 PM UTC (ff204a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1011/8d81603...philloooo:ff204a0.html" title="Last updated on Nov 2, 2021, 10:00 PM UTC (ff204a0)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/manifest/pull/1011.html" title="Last updated on Feb 17, 2022, 12:41 AM UTC (751e534)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1011/8d0663e...philloooo:751e534.html" title="Last updated on Feb 17, 2022, 12:41 AM UTC (751e534)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1430,11 +1430,11 @@
             Updating the manifest
           </h3>
           <p>
-            For installed web application, the manifest is fetched and processed
-            on every page load. When the processing is successful, user agents
-            MAY apply updated manifest to any current and future <a>application
-            contexts</a>
-            associated with the application. 
+            As specified for [^link/rel/manifest^] link relation, the manifest
+            is fetched and processed on every page load. When the [=processing
+            a manifest=] is successful, user agents MAY apply updated manifest
+            to any current and future <a>application contexts</a> associated
+            with the application.
           </p>
           <p>
             User agents SHOULD be careful about performing update to manifest
--- a/index.html
+++ b/index.html
@@ -1464,7 +1464,7 @@
             does not contain changes to [=security-sensitive members=].
           </p>
           <aside class="note" title=
-          "A user agent won't not apply a partial update">
+          "A user agent will not apply a partial update">
             <p>
               For example, the user agent could present options to the user:
             </p>
--- a/index.html
+++ b/index.html
@@ -1216,23 +1216,18 @@
             associated with the application. 
           </p>
           <p>
+            User agents SHOULD be careful about performing update to manifest
+            fields like app name and icons. Users should be fully aware if
+            such fields are to be be updated to avoid web applications being
+            installed and silently changed to appear as a different one.
+          </p>
+          <p>
             When the user agent fetches a manifest as part of an update check,
             it MUST [=url/equal|compare=] the fetched processed manifest's
             [=manifest/identity=] with the identity of the app being updated. If
             it is not equal, the user agent MUST NOT update the installed
             application's manifest. 
           </p>
-          <section>
-            <h3>
-              Privacy and security consideration
-            </h3>
-            <p>
-              User agents SHOULD be careful about performing update to manifest
-              fields like app name and icons. Users should be fully aware if
-              such fields are to be be updated to avoid web applications being
-              installed and silently changed to appear as a different one.
-            </p>
-          </section>
         </section>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -1221,7 +1221,6 @@
           <h3>
             Updating the manifest
           </h3>
-          <aside class="issue" data-number="446"></aside>
           <p>
             The manifest is fetched and processed on every page load. When the
             processing is successful, user agents MAY apply updated manifest to
--- a/index.html
+++ b/index.html
@@ -1437,11 +1437,46 @@
             with the application.
           </p>
           <p>
-            User agents SHOULD be careful about performing update to manifest
-            fields like app name and icons. Users should be fully aware if
-            such fields are to be be updated to avoid web applications being
-            installed and silently changed to appear as a different one.
+            For the purpose of updating, the following member are
+            <dfn>security-sensitive members</dfn>, as they are presented during
+            installation and on launch surfaces:
           </p>
+          <ol>
+            <li>[=manifest/short_name=],
+            </li>
+            <li>[=manifest/icons=]
+            </li>
+            <li>[=manifest/name=],
+            </li>
+          </ol>
+          <p data-cite="permissions">
+            User agents SHOULD NOT automatically apply changes to
+            [=security-sensitive members=] without [=express permission=] from
+            the user.
+          </p>
+          <p>
+            Instead, user agents SHOULD present changes to [=security-sensitive
+            members=] with appropriate management options, so the user can make
+            an informed decision about updating the web application.
+          </p>
+          <p>
+            The user agent MAY automatically apply the changes if the update
+            does not contain changes to [=security-sensitive members=].
+          </p>
+          <aside class="note" title=
+          "A user agent won't not apply a partial update">
+            <p>
+              For example, the user agent could present options to the user:
+            </p>
+            <ol>
+              <li>Accept the update
+              </li>
+              <li>Uninstall the web app, or
+              </li>
+              <li>Report the website as abusive/spam.
+              </li>
+            </ol>
+          </aside>
         </section>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -1232,6 +1232,7 @@
               such fields are to be be updated to avoid web applications being
               installed and silently changed to appear as a different one.
             </p>
+          </section>
         </section>
       </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1209,6 +1209,29 @@ <h3>
...
+          <section>
+            <h3>
+              Privacy and security consideration
+            </h3>
+            <p>
+              User agents SHOULD be careful about performing update to manifest
+              fields like app name and icons. Users should be fully aware if
+              such fields are to be be updated to avoid web applications being
+              installed and silently changed to appear as a different one.
+            </p>
```
NotWoods, line 1234: Missing `</section>` tag, which is causing later items to get different section numbers in the table of contents.

index.html
```diff
@@ -1222,6 +1222,25 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
```
marcoscaceres, line 1224:
```suggestion
```

index.html
```diff
@@ -1222,6 +1222,25 @@ <h3>
...
+            associated with the application. 
+          </p>
+          <p>
+            User agents SHOULD be careful about performing update to manifest
+            fields like app name and icons. Users should be fully aware if
+            such fields are to be be updated to avoid web applications being
+            installed and silently changed to appear as a different one.
+          </p>
+          <p>
+            When the user agent fetches a manifest as part of an update check,
```
marcoscaceres, line 1238: I think we should  move this section to the "processing a manifest" section as part of the algorithm: i.e. something like: 
 
1. If the document "processed manifest" is not null, and document [=processed manifest=]'s id member [=URL/equals=] manifest's id member, return. 

(i.e., prevent overriding an existing "processed manifest").

index.html
```diff
@@ -1222,6 +1222,25 @@ <h3>
             Updating the manifest
           </h3>
           <aside class="issue" data-number="446"></aside>
+          <p>
+            The manifest is fetched and processed on every page load. When the
```
marcoscaceres, line 1226: I think we need to update the HTML spec to say this instead. I.e., it should say that, for regular web applications, the manifest is loaded when needed, but for "installed" web applications, the manifest is fetched on page load. 

Then here we can just say point to the HTML spec.

marcoscaceres (changes requested):
@philloooo, happy to work on this together... I can take the HTML spec updates if you'd like.

index.html
marcoscaceres: I just realized that [=URL/equals=] allows us to "exclude fragments" as an argument, which might mean we can keep the fragment in the `id`, but ignore it for these comparisons. That might simplify `id` a little bit, because we don't need to chop the fragments anymore.

index.html
philloooo: Updated the html, but didn't find a way to reference it. Do you know how can I reference that?

marcoscaceres:
From our call today:

For the purpose of updating, the following member are <dfn>security-sensitive members</dfn>, as they are presented during installation and on launch surfaces:

  * [=manifest/name=],
  * [=manifest/short_name=],
  * [=manifest/icons=]

User agents SHOULD NOT automatically apply changes to [=security-sensitive members=] without [=express permission=] from the user.

Instead, user agents SHOULD present changes to [=security-sensitive members=] with appropriate management options, so the user can make an informed decision about updating the web application.

The user agent MAY automatically apply the changes if the update does not contain changes to [=security-sensitive members=].


Aside: A user agent won't not apply a partial update.

For example, the user agent could present options to the user:
Accept the update
uninstall the web app, or
report the website as abusive.

The user agent could present
Appropriate user agent options for the user , for example, 

Appropriate user agent actions could be, for example, to present the change to the user and provide the user with a means to uninstall the application or even report the website as abusive.

aarongustafson:
Sounds reasonable. Is there any option for a UA to make a determination on the relative substance of the change? For example, if the name went from Company to Company™, that's not terribly significant. Same could be said for changing the colors of an icon while maintaining the same design. I'm not suggesting how the UA might determine the substance of a change and where they should peg the threshold of needing to inform the user, but it could cut down on unnecessary notices.

marcoscaceres:
> Is there any option for a UA to make a determination on the relative substance of the change?

Sure, it's all just "recommendations" after all. However, implementers would do that at their own risk because an attacker would try to abuse such heuristics if/where possible. 

We might need to add that to the Security Considerations.

index.html
```diff
@@ -1421,7 +1429,19 @@ <h3 id="applying">
           <h3>
             Updating the manifest
           </h3>
-          <aside class="issue" data-number="446"></aside>
+          <p>
+            For installed web application, the manifest is fetched and processed
+            on every page load. When the processing is successful, user agents
+            MAY apply updated manifest to any current and future <a>application
+            contexts</a>
+            associated with the application.
```
marcoscaceres, line 1437:
```suggestion
            As specified for [^link/rel/manifest^] link relation, the manifest
            is fetched and processed on every page load. When the [=processing
            a manifest=] is successful, user agents MAY apply updated manifest
            to any current and future <a>application contexts</a> associated
            with the application.
```

index.html
```diff
@@ -1421,7 +1429,19 @@ <h3 id="applying">
...
+            MAY apply updated manifest to any current and future <a>application
+            contexts</a>
+            associated with the application. 
+          </p>
+          <p>
+            User agents SHOULD be careful about performing update to manifest
+            fields like app name and icons. Users should be fully aware if
+            such fields are to be be updated to avoid web applications being
+            installed and silently changed to appear as a different one.
+          </p>
```
marcoscaceres, line 1444:
```suggestion
            For the purpose of updating, the following member are
            <dfn>security-sensitive members</dfn>, as they are presented during
            installation and on launch surfaces:
          </p>
          <ol>
            <li>[=manifest/short_name=],
            </li>
            <li>[=manifest/icons=]
            </li>
            <li>[=manifest/name=],
            </li>
          </ol>
          <p data-cite="permissions">
            User agents SHOULD NOT automatically apply changes to
            [=security-sensitive members=] without [=express permission=] from
            the user.
          </p>
          <p>
            Instead, user agents SHOULD present changes to [=security-sensitive
            members=] with appropriate management options, so the user can make
            an informed decision about updating the web application.
          </p>
          <p>
            The user agent MAY automatically apply the changes if the update
            does not contain changes to [=security-sensitive members=].
          </p>
          <aside class="note" title=
          "A user agent won't not apply a partial update">
            <p>
              For example, the user agent could present options to the user:
            </p>
            <ol>
              <li>Accept the update
              </li>
              <li>Uninstall the web app, or
              </li>
              <li>Report the website as abusive/spam.
              </li>
            </ol>
          </aside>
```

index.html
```diff
@@ -1421,7 +1429,57 @@ <h3 id="applying">
...
+            an informed decision about updating the web application.
+          </p>
+          <p>
+            The user agent MAY automatically apply the changes if the update
+            does not contain changes to [=security-sensitive members=].
+          </p>
+          <aside class="note" title=
+          "A user agent will not apply a partial update">
+            <p>
+              When the update contains changes both to [=security-sensitive
```
philloooo, line 1469: @marcoscaceres I've added some  context to the partial update note, let me know what you think.

index.html
```diff
@@ -1421,7 +1429,57 @@ <h3 id="applying">
...
+              members=] and other members. A user agent SHOULD not automatically
```
marcoscaceres, line 1470:
```suggestion
              members=] and other members. A user agent won't automatically
```

index.html
```diff
@@ -1421,7 +1429,57 @@ <h3 id="applying">
...
+              partially update. For example, the user agent could present
+              options to the user:
+            </p>
+            <ol>
+              <li>Accept the update
+              </li>
+              <li>Uninstall the web app, or
+              </li>
+              <li>Report the website as abusive/spam.
+              </li>
```
marcoscaceres, line 1480:
```suggestion
              <li>Report the website as malicious.
              </li>
```

marcoscaceres (approved):

dmurph:
Adding myself in comments for searching purposes

PR #1013, opened 2021-10-11
use url equivalence excluding fragments for id comparison

philloooo:
The https://url.spec.whatwg.org/#url-equivalence defaults to exclude fragments = false. Clarify in `id` member that the equality comparison here excludes fragments.

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/manifest/pull/1013.html" title="Last updated on Oct 11, 2021, 7:54 PM UTC (804f2c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1013/8d81603...philloooo:804f2c1.html" title="Last updated on Oct 11, 2021, 7:54 PM UTC (804f2c1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -831,11 +831,12 @@
           application, it SHOULD treat that manifest as a description of a
           distinct application, even if it is served from the same URL as that
           of another application. When the user agent sees a manifest where
-          |manifest|["id"] [=url/equal|equals=] the [=identity=] of an
-          already-installed application, it SHOULD be used as a signal that
-          this manifest is a replacement for the already-installed
-          application's manifest, and not a distinct application, even if it is
-          served from a different URL than the one seen previously.
+          |manifest|["id"] [=URL serializer/exclude fragment|excluding
+          fragments=] [=url/equal|equals=] the [=identity=] of an
+          already-installed application, it SHOULD be used as a signal that this
+          manifest is a replacement for the already-installed application's
+          manifest, and not a distinct application, even if it is served from a
+          different URL than the one seen previously.
         </p>
         <p class="note">
           The [=identity=] can be used by a service that collects lists of web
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -831,11 +831,12 @@ <h3>
           application, it SHOULD treat that manifest as a description of a
           distinct application, even if it is served from the same URL as that
           of another application. When the user agent sees a manifest where
-          |manifest|["id"] [=url/equal|equals=] the [=identity=] of an
-          already-installed application, it SHOULD be used as a signal that
-          this manifest is a replacement for the already-installed
-          application's manifest, and not a distinct application, even if it is
-          served from a different URL than the one seen previously.
+          |manifest|["id"] [=URL serializer/exclude fragment|excluding
```
marcoscaceres, line 834: I think we need to move this part up to the [processing the manifest](https://www.w3.org/TR/appmanifest/#processing) section.

It should go:

21. Process the id member passing json, manifest and document URL.
22. If document's [=processed manifest=] is not `null`:
22.1 If [=processed manifest=]["id"] member is not  [=url/equal|equals=] with exclude fragment true to |manifest|["id"], then return. (We might need a note here to clarify if anything should happen when there is a mismatch - it's not an error, we just drop the new manifest on the floor).
22.2  If ???? members have changed, signal to the end user that manifest is a replacement for the already-installed application's manifest.
23. Let document's processed manifest be manifest.

Now 22.2 might have special rules, like:
 * How do we actually if the manifest has changed? Do we JSON-serialize them and do text comparisons?  
 * what happens if the `scope` changes, and the current page is out of scope?
 * Is there any situation where we can do a silent update? (e.g., only the theme-color changed... my gut feeling is no).
 * If the icons change, do we need to say MUST prompt? What else would should cause a prompt? 
 
Do we need final step "24", that once we have updated the manifest, we need to re-[apply](https://www.w3.org/TR/appmanifest/#applying) the manifest. However, that causes the entire application to reload... what does that mean? Do we need to navigate back to the start URL?

philloooo:
Merged this to #1011

PR #1034, opened 2022-03-16
Add scope conflict section about same-origin and nested scopes

### LINKED ISSUE #539: Relationship between nestedly scoped web app manifests?
There is an explicit section [match-service-worker-registration](https://w3c.github.io/ServiceWorker/#match-service-worker-registration) defining how nestedly scoped service workers work, where a SW with longest scope_url would "override" all the others:

>  5. Set matchingScopeString to the **longest** value in scopeStringSet which the value of clientURLString starts with, if it exists.

But of web app manifest, how nestedly scoped web app (manifest) effect with each other? Taking `/` and `/sub-app/` as an example, is the one with shorter path (which has a wider scope) totally bypassed? OR the longer-path one (the sub one) can inherit all missing members from the shorter/parent one? There seems no definition exist about this scenarios in the current spec.

Furthermore, there might exist some scenarios that a `/sub-app/` want treat `/` as sort of "parent app", where the sub-app can be navigated to from the "parent app" (in current browsing context) but has its own needs for a separated manifest (e.g. different themes/bg colors, different orientation, different desceiption etc. I think It's common?), thus should allow the navigation back from sub-app to "parent app".

There might be more complicated scenarios in practice, I believed It's worth discussion ;)

dmurph:
Closes #539

* [x] Chromium (https://bugs.chromium.org/)

Commit message:

Elaborate recommendations around scope conflicts like same-origin scopes, nested scopes, etc.

* chore:
* editorial:
* BREAKING CHANGE:

***

### :boom: Error: 502 Bad Gateway :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 2, 2023, 12:15 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fdmurph%2Fmanifest%2F76e211629d4e66082c926ccde42ed2fbdcad1175%2Findex.html%3FisPreview%3Dtrue)

```
<html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>cloudflare</center>
</body>
</html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/manifest%231034.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2501,8 +2501,11 @@
           (detailed below) and is thus considered bad practice.
         </p>
         <p>
-          Same-origin scopes are not recommended due to origin-based settings
-          that will affect all apps installed under that origin. Settings like:
+          One origin per application is ideal because it isolates common user
+          agent settings and controls to apply to only one web applications.
+          Without this isolation, these settings or controls can apply to more
+          than one web application, which will likely be unexpected by the
+          user. Examples:
         </p>
         <ul>
           <li>Permissions
@@ -2533,8 +2536,8 @@
           </li>
           <li>Notifications and notification attribution.
           </li>
-          <li>And current or future APIs that somehow correspond to a single web app identity
-          (e.g. badging).
+          <li>And current or future APIs that somehow correspond to a single
+          web app identity (e.g. badging).
           </li>
         </ul>
         <aside class="example">
--- a/index.html
+++ b/index.html
@@ -2467,8 +2467,7 @@
           </li>
         </ul>
         <p>
-          Nested scopes are strongly not recommended. They have the above
-          complications of same-origin scope, along with the following UX and
+          Further, overlapping, nested, or duplicate scopes can have the following UX and
           API problems or inconsistencies among other possible consistencies:
         </p>
         <ul>
--- a/index.html
+++ b/index.html
@@ -2434,6 +2434,59 @@
           </p>
         </aside>
       </section>
+      <section class="informative">
+        <h3>
+          Web Apps with scope conflicts
+        </h3>
+        <p>
+          It is possible for two web apps to be installed where there are
+          conflicts between the scope. These conflicts come with disadvantages
+          outlined below and they are generally not recommended.
+        </p>
+        <ul>
+          <li>The scopes of the two web apps can be on the same origin. Not
+          recommended.
+          </li>
+          <li>The scope of one web app can be nested inside the scope of the
+          other. Strongly not recommended.
+          </li>
+          <li>The scopes of the two web apps can be the same. Strongly not
+          recommended.
+          </li>
+        </ul>
+        <p>
+          Same-origin scopes are not recommended due to origin-based settings
+          that will affect all apps installed under that origin. Settings like:
+        </p>
+        <ul>
+          <li>Permissions
+          </li>
+          <li>Storage and storage quota
+          </li>
+          <li>User settings (e.g. font size)
+          </li>
+        </ul>
+        <p>
+          Nested scopes are strongly not recommended. They have the above
+          complications of same-origin scope, along with the following UX and
+          API problems or inconsistencies among other possible consistencies:
+        </p>
+        <ul>
+          <li>Installation prompting may not work for the nested app if the
+          outer app is installed.
+          </li>
+          <li>User-agent UX around launching an app for a browsing context may
+          be inconsistent or not appear.
+          </li>
+          <li>Badging API calls will not be able to consistently update the
+          correct web app badge.
+          </li>
+          <li>Notifications may have incorrect attribution or not appear.
+          </li>
+          <li>Future APIs may not work at all in this configuration.
+          </li>
+        </ul>
+      </section>
     </section>
     <section>
       <h2>
--- a/index.html
+++ b/index.html
@@ -2495,21 +2495,11 @@
           Web Apps with scope conflicts
         </h3>
         <p>
-          Because scopes are based on URL matching, it is possible for a developer to create
-          multiple web applications with the same, overlapping, or nested scopes. Doing so creates
-          several issues (detailed below) and is thus considered bad practice.
+          Because scopes are based on URL matching, it is possible for a
+          developer to create multiple web applications with the same,
+          overlapping, or nested scopes. Doing so creates several issues
+          (detailed below) and is thus considered bad practice.
         </p>
-        <ul>
-          <li>The scopes of the two web apps can be on the same origin. Not
-          recommended.
-          </li>
-          <li>The scope of one web app can be nested inside the scope of the
-          other. Strongly not recommended.
-          </li>
-          <li>The scopes of the two web apps can be the same. Strongly not
-          recommended.
-          </li>
-        </ul>
         <p>
           Same-origin scopes are not recommended due to origin-based settings
           that will affect all apps installed under that origin. Settings like:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+        <h3>
+          Web Apps with scope conflicts
+        </h3>
+        <p>
+          It is possible for two web apps to be installed where there are
+          conflicts between the scope. These conflicts come with disadvantages
+          outlined below and they are generally not recommended.
+        </p>
+        <ul>
+          <li>The scopes of the two web apps can be on the same origin. Not
```
aarongustafson, line 2447: Do you think the "Not recommended" (and similar) bits should be _emphasized_?

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+          recommended.
+          </li>
+          <li>The scope of one web app can be nested inside the scope of the
+          other. Strongly not recommended.
+          </li>
+          <li>The scopes of the two web apps can be the same. Strongly not
```
aarongustafson, line 2453: The other two have expansions below. Should this one be bundled with nested scoping in that expansion?

aarongustafson:
@dmurph I had two recommendations masked as questions. Feel free to take ’em or leave ’em.

aarongustafson (changes requested):

index.html
marcoscaceres: We should avoid RFC2119 language in non-normative sections. 

Instead, we should be clear on what the implications are. 

I think this list can go as it's covered by the paragraph suggested above.

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+          </li>
+          <li>Storage and storage quota
+          </li>
+          <li>User settings (e.g. font size)
+          </li>
+        </ul>
+        <p>
+          Nested scopes are strongly not recommended. They have the above
+          complications of same-origin scope, along with the following UX and
+          API problems or inconsistencies among other possible consistencies:
```
marcoscaceres, line 2472:
```suggestion
          Further, overlapping, nested, or duplicate scopes can have the following UX and
          API problems or inconsistencies among other possible consistencies:
```

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+        </p>
+        <ul>
```
marcoscaceres, line 2474: I think this is getting into user agent behaviour... and that might be a good thing for us to specify.

The simplest model might be last installed and/or most specific wins.

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+          <li>Installation prompting may not work for the nested app if the
+          outer app is installed.
+          </li>
+          <li>User-agent UX around launching an app for a browsing context may
+          be inconsistent or not appear.
+          </li>
+          <li>Badging API calls will not be able to consistently update the
+          correct web app badge.
+          </li>
```
marcoscaceres, line 2483: This is overly specific (not all UAs) support badging... should be generalised that API calls might fail.

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
+          <li>Notifications may have incorrect attribution or not appear.
+          </li>
+          <li>Future APIs may not work at all in this configuration.
```
marcoscaceres, line 2486: Hmm... so yeah... we might need to deal with the reality that that overlapping scopes can happen and specify what the AU should actually do. The focus here needs to be on user expectation that things should work. 

In theory, it should be possible to go from most specific scope to least specific.... we might need to look at what Service Workers did here, as, in theory, they have the same problem.

index.html
```diff
@@ -2434,6 +2434,59 @@ <h3>
...
       </section>
+      <section class="informative">
+        <h3>
+          Web Apps with scope conflicts
+        </h3>
+        <p>
+          It is possible for two web apps to be installed where there are
+          conflicts between the scope. These conflicts come with disadvantages
+          outlined below and they are generally not recommended.
+        </p>
```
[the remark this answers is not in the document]
aarongustafson: Agreed. Adding "_with_ the same":

```suggestion
        <p>
          Because scopes are based on URL matching, it is possible for a developer to create
          multiple web applications with the same, overlapping, or nested scopes. Doing so creates
          several issues (detailed below) and is thus considered bad practice.
        </p>
```

index.html
```diff
@@ -2434,6 +2434,58 @@ <h3>
...
+          <li>The scopes of the two web apps can be on the same origin. Not
+          recommended.
+          </li>
+          <li>The scope of one web app can be nested inside the scope of the
+          other. Strongly not recommended.
+          </li>
+          <li>The scopes of the two web apps can be the same. Strongly not
+          recommended.
+          </li>
+        </ul>
```
marcoscaceres, line 2456: This is effectively said above (they are all bad practice), so it can all go.

```suggestion
```

index.html
```diff
@@ -2434,6 +2434,58 @@ <h3>
...
+        <p>
+          Same-origin scopes are not recommended due to origin-based settings
```
marcoscaceres, line 2458: This one doesn't quite make sense as written. It reads like we are discouraging same-origin scopes? It would be better to frame this explaining why one origin per application is ideal.

marcoscaceres:
Hi @dmurph, just went over this again... I think all of my comments, along with @aarongustafson' comments, still stand.

index.html
dmurph: I tried creating a definition here. I'm wondering if we should pull this out of 'normative', and if we should maybe enumerate all APIs here that affect the web app's OS identity? e.g. badging, notifications.... those are the only two I can think of right now.

index.html
dmurph: As mentioned in another comment - I'm playing around with adding a definition here, but perhaps we want to pull out a section per API that is affected, and then reference the "controlling app" concept in that section? What do you think?

index.html
dmurph: Done. I could also call out examples of each of the list items being bad. e.g. Clearing storage for app A will clear storage for app B, Giving app A location permission possibly giving app B location permission, etc.

index.html
dmurph: I now mention it in an example list - let me know if you think I should remove it there.

dmurph:
FYI I just received feedback from @marcoscaceres offline and I'm going to redo this to make it a whole lot shorter.

index.html
```diff
@@ -2490,6 +2490,72 @@ <h3>
...
+          <li>Permissions
+          </li>
+          <li>Storage and storage quota
+          </li>
+          <li>User settings (e.g. font size, zoom level)
+          </li>
+        </ul>
+        <p>
+          To aide with disambiguation apps with scope conflicts, user agents
+          can determine the <dfn>controlling app</dfn> for a given document
```
mgiuca, line 2520: +1 this feels like getting into normative (optional) requirements in a non-normative section. Look for instances of the word "can" and "may" in the above paragraphs and ask yourself whether you're specifying things for the user agent - I think you are, so those words should be capitalized ("can" should become "MAY") and moved into a normative section, or this section can become normative with notes for non-normative bits.

Conversely, the "may" in "The following scenarios may require" below is not a normative requirement so should be reworded to avoid the word "may" - e.g. "The following scenarios could require". Or turn it into a proper normative statement as Marcos is suggesting.

index.html
mgiuca: +1 for specifying the behaviour around this (perhaps starting with MAY requirements so we don't require browsers to update right away). It doesn't feel good to say "not work" as a possible outcome.

For simplicity, I'm OK with leaving this here for now, but long-term as you say it would be good for each section to specify "Choose the controlling app".

There's actually prior art for this which we might want to merge into this concept: in the Badging API there's "[determining the set of matching applications](https://w3c.github.io/badging/#determining-the-set-of-matching-applications)" which is a bit weird in that it returns a set but usually only with one item in it. If you just look at the document case (the top half), it's explicitly picking the one with the most specific scope.

dmurph:
Because Safari & Chromium differ in their behavior here (Safari creates a totally isolated world for each app), this guidance probably shouldn't live here. There already is guidance for Chromium on web.dev [here](https://web.dev/building-multiple-pwas-on-the-same-domain/), so that is probably fine for now.

PR #1069, opened 2023-01-20
Allow manifest processing to be invoked without going through an HTML document

### LINKED ISSUE #1068: Allow manifest processing to be invoked without going through an HTML document
Currently, per the spec, manifests can **only** be fetched via a `<link rel="manifest">` element in an HTML document. This is (as of #894 — July 2020) tightly coupled into the manifest spec, as the "processing a manifest" algorithm takes an HTML `link` element and a `Response` object as inputs. So the only way to invoke these steps is to have an open HTML document with a `link` to the manifest, and fetch the manifest using the HTML fetch machinery, and have a corresponding `Response` object.

However, this excludes or complicates a number of key use cases (which are in use today in various browsers), which fall under the umbrella of installing a web app when you don't already have the web page open. Use cases include:

- Installation directly from a "store" page (e.g. an app listing site or app, has a button to directly install a web app without first navigating to the app in a browser tab).
- Installation from another site (a possible future API where an app installation can be triggered from a third party origin).
- Automatic installation via sync (the user manually installs an app on device X, it syncs across to device Y and automatically gets installed without having to load the page on device Y).
- Automatic installation via admin policy (apps get installed when the user first opens their device).
- Automatic installation of a hard-coded set of default apps for a device or browser.

At the moment, browsers implement these features using a fairly clunky hack: they record a document URL. At install time, the browser creates an invisible web contents, loads the full HTML page, then triggers the app install, as required by the spec. There are a number of major downsides to this process:

- Having to load a full HTML page just to trigger the install then discarding it wastes a lot of bandwidth and time.
- Loading a page on startup without user consent is seen as a privacy issue. The user agent is sending requests to a third-party website without consent. (Granted, we would still have to send a request to fetch the manifest and icons, but these could be cached by the user agent ahead of time, e.g. the full text of these could be sent through the sync system.)
- It is necessarily flaky. Since pages can take an arbitrary amount of time to load the `<link>` element, we need to wait an arbitrary amount of time. In practice, we have a long (~1 minute) timeout, which means installations can fail because of network slowness (whereas if we directly loaded the manifest, we could just continue to wait as long as the connection remains open).

This hack even extends to developers and administrators. To get around the timing issue above, we have suggested that developers create a hidden "install page" - a blank HTML page that simply contains a `<link>` to the manifest, just so we use that as the document URL and avoid loading a full website. Then we have to tell administrators to use the "install page" instead of the regular start URL. These hacks create extra complexity for everyone.

Therefore, I propose that we:

1. Partially revert #894 to have "processing a manifest" take a document URL and manifest URL again (and let the HTML spec extract the URLs from the relevant objects and pass them in). This would allow the user agent to invoke "processing a manifest" from outside the HTML document.
2. Explicitly state (in normative text) that the user agent can do this without loading a document. However, add a requirement that if the user agent does this, they must in some way verify that there is a links back from the document to the manifest (otherwise a manifest on one origin could apply unauthorized metadata to app pages on someone else's origin).

Note: I am not currently proposing that we remove the document URL (which is what I proposed in #668, and could be a follow-up step). I am proposing that we remove the requirement that the document is _loaded_ in order to process the manifest. (This way, the user agent still has to verify the bidirectional link between the two origins, but it can be done ahead of time, not at install time -- for example if you are using sync, it has already been verified on the other device.) So we still need a document URL for the time being.

Note: This is related to #668 but not identical (it can be considered a stepping stone to that more advanced and risky change).

Note for completeness that I originally wrote this up in [this doc](https://docs.google.com/document/d/1NUsoTWkWrPFFNnxR00r1NF-MXzq-_rjl2ldAhZm-TEw/edit#) but all relevant details appear above.

mgiuca:
Closes #1068

Commit message:

Allow manifest processing to be invoked without going through an HTML document.

Manifest processing: Replaces the link and response parameters with document URL and manifest URL. This was limiting the ability to call the processing algorithm from outside an HTML document context.

Note that the only call to this algorithm is in the HTML spec, which needs to be updated simultaneously to use the new interface.
 
Adds new normative text allowing user agents to invoke the processing steps without a document, provided that
they supply a valid document URL and if the document URL is on a different origin to the manifest, there is a page on the document's origin that links to the manifest.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/1069.html" title="Last updated on Apr 27, 2023, 3:04 AM UTC (3d03340)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1069/0252939...mgiuca:3d03340.html" title="Last updated on Apr 27, 2023, 3:04 AM UTC (3d03340)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1383,14 +1383,14 @@
             The [=processing a manifest=] steps are invoked by [[HTML]]'s
             processing steps for the [^link^] element, but MAY also be invoked
             by the user agent to process a manifest without an associated
-            document.
+            [=document=].
           </p>
           <p>
-            In this case, the user agent SHOULD ensure that, at least at some
+            In this case, the user agent SHOULD ensure that at least at some
             point in the past:
           </p>
           <ul>
-            <li>there was at least one page which is [=same origin=] as
+            <li>there was at least one [=document=] that is [=same origin=] as
             |document URL:URL| that has a [^link^] element
             |linkElement:HTMLLinkElement| with a [^link/rel^] of
             [^link/rel/manifest^] and a [^link/href^] that resolves to
@@ -1407,7 +1407,7 @@
           <p>
             (The [[HTML]] steps automatically ensure this.)
           </p>
-          <div class="note">
+          <aside class="note">
             <p>
               This provision allows user agents to perform app installations
               using a manifest without having to first load a document that
@@ -1429,7 +1429,7 @@
               the manifest is fetched with or without credentials, as agreed by
               both origins, as it would if going directly through the document.
             </p>
-          </div>
+          </aside>
         </section>
         <section>
           <h3 id="applying">
--- a/index.html
+++ b/index.html
@@ -1254,15 +1254,10 @@
           <p>
             The steps for <dfn data-export="" data-local-lt=
             "processing">processing a manifest</dfn> are given by the following
-            algorithm. The algorithm takes [^link^] |el:HTMLLinkElement|, a
-            [=Response=] |response|, and a [=byte sequence=] |bodyBytes|.
+            algorithm. The algorithm takes a [=URL=] |document URL:URL|, a
+            [=URL=] |manifest URL:URL|, and a [=byte sequence=] |bodyBytes|.
           </p>
           <ol class="algorithm">
-            <li>Let |document URL:URL| be |el|'s [=Node/node document=]'s
-            [=Document/URL=].
-            </li>
-            <li>Assert: |document URL:URL| is not null.
-            </li>
             <li>Let |json| be the result of [=parse JSON bytes to an Infra
             value=] passing |bodyBytes|.
             </li>
@@ -1284,8 +1279,6 @@
             </li>
             <li>[=Process a text member=] passing |json|, |manifest|, and
             "short_name".
-            </li>
-            <li>Let |manifest URL:URL| be |response|'s [=response/URL=].
             </li>
             <li>[=Process the `start_url` member=] passing |json|, |manifest|,
             |manifest URL|, and |document URL|.
--- a/index.html
+++ b/index.html
@@ -1238,8 +1238,8 @@
             to process a manifest without an associated document. In this case,
             the user agent MUST supply a |document URL| which is either <a>same
             origin</a> as |manifest URL|, or <a>same origin</a> as at least one
-            page with a [^link^] element of type "manifest" that references the
-            given |manifest URL|.
+            page with a [^link^] element of type [^link/rel/manifest^] that
+            references the given |manifest URL|.
           </p>
           <p class="note">
             This provision allows user agents to perform app installations
--- a/index.html
+++ b/index.html
@@ -1385,16 +1385,16 @@
             by the user agent to process a manifest without an associated
             document. In this case, the user agent MUST supply a |document URL|
             which is either <a>same origin</a> as |manifest URL|, or <a>same
-            origin</a> as at least one page with a [^link^] element of type
-            [^link/rel/manifest^] that references the given |manifest URL|.
+            origin</a> as at least one page on origin |linkOrigin| with a
+            [^link^] element |linkElement| of type [^link/rel/manifest^] that
+            references the given |manifest URL|.
           </p>
           <p>
             Additionally, the user agent SHOULD ensure that the |bodyBytes|
             data was fetched with a [=CORS Request=] whose <a data-xref-type=
-            "http-header">`Origin`</a> matches a page with a [^link^] element
-            of type [^link/rel/manifest^] that references the given |manifest
-            URL|, and that its [=Request/credentials mode=] is set to the
-            [=CORS settings attribute credentials mode=] for that [^link^]'s
+            "http-header">`Origin`</a> is |linkOrigin|, and whose
+            [=Request/credentials mode=] is set to the [=CORS settings
+            attribute credentials mode=] for |linkElement|'s
             [^link/crossorigin^] attribute. (The [[HTML]] steps automatically
             ensure this.)
           </p>
--- a/index.html
+++ b/index.html
@@ -1383,35 +1383,53 @@
             The [=processing a manifest=] steps are invoked by [[HTML]]'s
             processing steps for the [^link^] element, but MAY also be invoked
             by the user agent to process a manifest without an associated
-            document. In this case, the user agent MUST supply a |document URL|
-            which is either <a>same origin</a> as |manifest URL|, or <a>same
-            origin</a> as at least one page on origin |linkOrigin| with a
-            [^link^] element |linkElement| of type [^link/rel/manifest^] that
-            references the given |manifest URL|.
+            document.
           </p>
           <p>
-            Additionally, the user agent SHOULD ensure that the |bodyBytes|
-            data was fetched with a [=CORS Request=] whose <a data-xref-type=
-            "http-header">`Origin`</a> is |linkOrigin|, and whose
-            [=Request/credentials mode=] is set to the [=CORS settings
-            attribute credentials mode=] for |linkElement|'s
-            [^link/crossorigin^] attribute. (The [[HTML]] steps automatically
-            ensure this.)
+            In this case, the user agent SHOULD ensure that, at least at some
+            point in the past:
           </p>
-          <p class="note">
-            This provision allows user agents to perform app installations
-            using a manifest without having to first load a document that links
-            to the manifest (e.g. when installing an app onto the user's device
-            via a sync service). But the above requirement forces the user
-            agent to verify that, when an application scope is on origin A and
-            the manifest is on a different origin B, there is a bidirectional
-            relationship between the two origins: that at least some page on
-            origin A links to the manifest on origin B. (Otherwise, it would be
-            possible for a manifest on origin B to control the metadata for an
-            unaffiliated app on origin A.) Also, the user agent is expected to
-            fetch the manifest using the same CORS settings as it would if
-            going directly through the document.
+          <ul>
+            <li>there was at least one page which is [=same origin=] as
+            |document URL:URL| that has a [^link^] element
+            |linkElement:HTMLLinkElement| with a [^link/rel^] of
+            [^link/rel/manifest^] and a [^link/href^] that resolves to
+            |manifest URL:URL|, and that
+            </li>
+            <li>if |manifest URL| is not [=same origin=] as |document URL|, the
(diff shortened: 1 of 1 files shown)


mgiuca:
@marcoscaceres @dmurph adding both; let's take our time with this one.

I'm not sure if you want to just start with the interface change and add the text later, if the text is controversial. I wanted to capture that this is explicitly allowed now (invoking this from outside a document) but I also want to make sure user agents verify the bidirectional link. This is not just a security consideration, it is a requirement. (It could also be downgraded to a security consideration if you like.)

It was difficult to phrase the requirement. Note that I don't expect user agents to literally verify the document links to the manifest at install time (otherwise we effectively require that you load the whole document), just that it is known to have linked to the manifest in the past. This means, e.g. if you're doing sync, then you install through the normal HTML flow on device X, recording both the document URL and manifest URL in the sync state, then when you sync to device Y, the user agent on Y can invoke the processing steps with the stored document URL and manifest URL, knowing that they had been bidirectionally associated in the past. Do you think the text adequately captures this?

mgiuca:
Wrote whatwg/html#8754 to update HTML to use the new interface. (Which also moves some of the deleted lines from Manifest into HTML.)

marcoscaceres (approved):
I'm still a bit worried about the security implications of this (in that the document provides the security/cors policy), so I wonder if we should add a note that this is a bit experimental?

dmurph (approved):
This LGTM - the other patch is approved - can we modify the descriptions & submit?

aarongustafson:
LGTM as well. I welcome this change! It'll be a boon for app stores, packaging, & so much more. Thanks for doing this @mgiuca!

mgiuca:
Hi folks,

Sorry for taking awhile to circle back on this. I wanted to take time to self-review since it was a long time since I uploaded the PRs, and that meant pushing it back until I found some free time.

So now that I've paged it back in, the security concern Marcos had is basically that the HTML applies lots of CORS checks before downloading the manifest. This change (particularly, what's implied by the non-normative note) allows user agents to magically supply a manifest from anywhere, not follow the HTML download steps. In some sense, that's the whole point (e.g., we are trying to use this to create a service that installs apps from pre-cached manifests, we explicitly _do not want_ to have to follow the HTML download steps to fetch it from its actual URL). But in another sense, that opens up the possibility of a UA not following CORS.

It's possible we want stronger language, like "the user agent MUST use a manifest that it previously downloaded using the CORS stuff from HTML" but I don't really know how to phrase that or what the implications are. For now, I think (if you all agree) it's OK to leave that up to user agents to work out, and trust that they will avoid spoofs. Maybe this can just be a security consideration (in a follow-up? Or should I add it in this PR?).

Since several of you have pinged me to say "please merge this so we can get on with our lives", I'm thinking we just merge it for now and add security considerations as a followup. Unless someone objects, I'll do that as soon as I get an approval on whatwg/html#8754. I have rebased.

****

I haven't used "merge queue" before. I don't see a way that I can enter a commit message -- how do you do that now? (Is it going to use the original post text? That contains stuff I don't want including the preview and draft links.)

aarongustafson:
@mgiuca Perhaps adding another RECOMMENDED to the [Privacy & Security Considerations](https://www.w3.org/TR/appmanifest/#priv-sec) section that recommends UAs that make use of Manifest files not bound to an HTML document apply the same rigor & protections they would under CORS? Since it's possible the Manifest could be part of a package, I'm not really sure what the mechanism should be. How does the Web Packaging spec deal with this?

As for the merge queue, I have no idea. I don't think I've paid attention to that before, I've mainly squashed (which I thought was what we had on this repo previously).

mgiuca:
Thanks @aarongustafson - I'm re-reading up on CORS now so I can do it justice.

I filed #1075 about the merge queue thing.

mgiuca:
I did a lot of thinking about CORS and here's what I came up with. Note that I've moved all the new text to its own section, "Processing a manifest without a document" which you can see in the preview. It now basically has prose that says "you SHOULD do CORS the way HTML does it".

Note that this is blocked on whatwg/html#9204 to export the definition of "CORS settings attribute credentials mode".

Also all the links to the link rel "`manifest`" value (both here, and existing already in the document) link to the wrong place due to w3c/respec#4435. Were it not for that, I'd have used `[^link/rel/manifest^]` more to directly link to the algorithm in HTML, but I can't get a direct link to that to work at the moment.

mgiuca:
I've done another pass over this section with @alancutter (thanks).

- Changed MUST into a SHOULD. We can't really expect all uses to directly verify this.
- Removed the "or" clause that the document be same-origin as manifest; you still want a link from the document to the manifest either way.
- Added "at least at some point in the past", to acknowledge that you don't need to verify this at install time, just whenever you did the caching.
- Clarify that the CORS request is only needed if the manifest is not same-origin as the document.

I think this basically captures the way that I at least expect this interface to be used outside of HTML. In short, it says "this should be called as-if you found it in an HTML document and invoked the 'steps to process this type of linked resource'". It captures the need to have a link from somewhere on this origin, and make the request with CORS.

Would appreciate another review from at least one of { @marcoscaceres , @dmurph , @aarongustafson } since I've rewritten the prose section in its entirety.

On the plus side, when this was originally proposed we were thinking we'd need a follow up to think about CORS better - that has now been done!

aarongustafson (approved):

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
         </section>
+        <section>
+          <h3>
+            Processing the manifest without a document
+          </h3>
+          <p>
+            The [=processing a manifest=] steps are invoked by [[HTML]]'s
+            processing steps for the [^link^] element, but MAY also be invoked
+            by the user agent to process a manifest without an associated
+            document.
```
marcoscaceres, line 1386:
```suggestion
            [=document=].
```

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+          </p>
+          <p>
+            In this case, the user agent SHOULD ensure that, at least at some
+            point in the past:
```
marcoscaceres, line 1390:
```suggestion
            In this case, the user agent SHOULD ensure that at least at some
            point in the past:
```

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+          </p>
+          <ul>
+            <li>there was at least one page which is [=same origin=] as
```
marcoscaceres, line 1393:
```suggestion
            <li>there was at least one [=document=] which is [=same origin=] as
```

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+            |document URL:URL| that has a [^link^] element
+            |linkElement:HTMLLinkElement| with a [^link/rel^] of
+            [^link/rel/manifest^] and a [^link/href^] that resolves to
+            |manifest URL:URL|, and that
```
marcoscaceres, line 1397: Nit: not sure if we have a concept of this... but it feels like we should say something like "a valid link type manifest"... or just a "link type "manifest"", as per HTML.

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+            </li>
+            <li>if |manifest URL| is not [=same origin=] as |document URL|, the
+            |bodyBytes:byte sequence| data was fetched with a [=CORS Request=]
```
marcoscaceres, line 1400: Maybe "is fetched"?

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+            whose <a data-xref-type="http-header">`Origin`</a> is |document
+            URL|'s [=url/origin=], and whose [=Request/credentials mode=] is
+            set to the [=CORS settings attribute credentials mode=] for
+            |linkElement|'s [^link/crossorigin^] attribute.
+            </li>
+          </ul>
+          <p>
+            (The [[HTML]] steps automatically ensure this.)
```
marcoscaceres, line 1408: I wonder if you can can say this directly in the above (something like, "In keeping with [[HTML]] behavior..." or something). It's a bit risky in case it gets moved or whatever, but should be ok... it will also make it more clear what "it" is.

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+          </p>
+          <div class="note">
```
marcoscaceres, line 1410: Might be good to add a "`title="some title for the note"`"

marcoscaceres, line 1410:
```suggestion
          <aside class="note">
```

index.html
```diff
@@ -1382,6 +1375,62 @@ <h2>
...
+              a manifest on origin B to control the metadata for an
+              unaffiliated app on origin A).
+            </p>
+            <p>
+              The second check ensures that origin B allows (via the [=CORS
+              protocol=]) its manifest to be applied by origin A, and also that
+              the manifest is fetched with or without credentials, as agreed by
+              both origins, as it would if going directly through the document.
+            </p>
+          </div>
```
marcoscaceres, line 1432:
```suggestion
          </aside>
```

marcoscaceres (changes requested):
couple of little things, but otherwise looks good I think.

index.html
mgiuca: By the way, the "at least some point in the past" clause was to acknowledge that we don't need the user agent to verify this exactly at install time. It's sufficient to have verified this at the time you cached it (e.g. if you install on Machine A, you verify this, then sync the metadata to Machine B and install it, you don't need to verify it again on Machine B).

We thought about writing "at the time of fetching the metadata" but we wanted to also allow a situation where you verify it on Machine A, then you sync the manifest URL to Machine B, Machine B fetches the manifest but does not re-verify this condition. So we settled for the more permissive "at some point in the past".

index.html
mgiuca: HTML doesn't export a concept of "link type". Internally, they just link "link type" to the [#linkTypes](https://html.spec.whatwg.org/multipage/links.html#linkTypes) heading, but it isn't exported.

I think it's reasonable to keep this the way it is; it's just literally stating what it is instead of referring to a concept.

index.html
mgiuca: Done.

index.html
mgiuca: Done.

index.html

mgiuca:
Done the requested changes (see comments).

mgiuca:
Friendly ping, @marcoscaceres - do the updates look good now?

marcoscaceres (approved):

marcoscaceres:
Oops! sorry for the delay @mgiuca. Was traveling last week.

PR #1101, opened 2023-10-09
Add members for localization

### LINKED ISSUE #676: Add optional `translations` member
Sometimes it's difficult to find an app or game on a specific (especially, unpopular) language.

Conversely some apps (games) are available only in one language (e.g. Chinese).

Since we want PWAs to be presented in app stores, what about adding optional `"translations"`/`"languages"`/`"supported_languages"` array of strings member?

Something like:

```json
"translations": ["en","fr","it"]
```

It should help to find (filter/exclude) apps by supported languages.

### LINKED ISSUE #1077: No way to localize icons?
Declaring multiple icons
https://w3c.github.io/manifest/#declaring-multiple-icons

> In the following example, the developer has made the following choices about the icons associated with the web application (etc.)

There are ways to include multiple icons with different `sizes`, but there is no way to indicate when these should be used with specific locales, such as when the icon contains localizable text or when there are regional variations in the icon.

### LINKED ISSUE #1078: No language/direction metadata for shortcuts?
1.1.4 Creating shortcuts 
https://w3c.github.io/manifest/#creating-shortcuts

Multiple examples are provided of creating a "shortcut" in a manifest. Here is one such:

```
  "shortcuts": [
    {
      "name": "Play Later",
      "description": "View the list of podcasts you saved for later",
      "url": "/play-later",
      "icons": [
        {
          "src": "/icons/play-later.svg",
          "type": "image/svg+xml"
        }
      ]
    },
```

The `name` and `description` fields do not have their own `lang` and `dir` values nor does the "shortcut" structure have the ability to provide an override. The whole manifest does have a `lang` and `dir`, but this doesn't allow for separate localizations nor for local overrides of the values. I18N normally recommends the ability to provide a local override when there is file-level metadata.

This design also suggests that some language is more privileged than all others or that large sections of the file must be repeated in order to localize the full experience described by the manifest.

### LINKED ISSUE #1080: "localizable members" should be defined separately?
`dir` member
https://w3c.github.io/manifest/#dir-member

> The localizable members are:
> 
>    [name](https://w3c.github.io/manifest/#dfn-name) member.
>    [short_name](https://w3c.github.io/manifest/#dfn-short_name) member.
>    [Shortcut item's](https://w3c.github.io/manifest/#dfn-shortcut-item) [name](https://w3c.github.io/manifest/#dfn-name-0) member.
>    [Shortcut item's](https://w3c.github.io/manifest/#dfn-shortcut-item) [short_name](https://w3c.github.io/manifest/#dfn-short_name-0) member.
>    [Shortcut item's](https://w3c.github.io/manifest/#dfn-shortcut-item) [description](https://w3c.github.io/manifest/#dfn-description) member.

This is more of an editorial comment than it is a technical objection.

The section about the `dir` member defines the term "localizable member" and includes the above definition of which fields are localizable. It seems like "localizable" should be defined separately so that it can be referred to by various members and sub-members as appropriate and so that the list of items belonging to it doesn't inadvertently become stale.

Note that "localization" encompasses more than just translation of strings. It can and often does include changing other items found in a manifest, such as icons, colors, and such. The textual items are important, of course, and are related to direction and language metadata more directly than (say) an icon, but this shouldn't be overlooked.

### LINKED ISSUE #1084: Use of `CanonicalizeUnicodeLocaleId`
`lang` member
https://w3c.github.io/manifest/#lang-member

> Set manifest["lang"] to the result of calling the [CanonicalizeUnicodeLocaleId](https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid) abstract operation with json["lang"].

The instructions on processing the `lang` member are inconsistent with the description of what a `lang` member can contain. The ECMA-402 function `CanonicalizeUnicodeLocaleID` is invoked and this results in a variety of potential changes to a few otherwise-valid values. This is probably the right thing to do when actually consuming the `lang` member's _value_ (particularly with `Intl.Locale` and friends in JavaScript), but its sudden appearance buried in the processing instructions makes this surprising. This suggests that `lang` should actually be defined as a Unicode Locale ID??

It's not clear if referencing Unicode Locale IDs is the right thing for the manifest spec to do, since the manifest is mainly concerned with transmitting data values, not actually performing runtime processing.

### LINKED ISSUE #1085: `name` and `short_name` (and other localizable member's) relationship to localization and accessibility
`name` member
https://w3c.github.io/manifest/#name-member

> The [manifest's](https://w3c.github.io/manifest/#dfn-manifest) name member is a [string](https://infra.spec.whatwg.org/#string) that represents the [name](https://w3c.github.io/manifest/#dfn-name) of the web application as it is usually displayed to the user (e.g., amongst a list of other applications, or as a label for an icon).
>
> The name member serves as the [accessible name](https://www.w3.org/TR/accname-1.2/#dfn-accessible-name) of an [installed web application](https://w3c.github.io/manifest/#dfn-installed-web-application).

The "accessible name" and the display name (or its `short_name` equivalent) really should be localized to match the expectation of the user. Many apps are localized into many languages, but the design of manifest (with `lang` and `dir` as top level items applying to *all* values inside the manifest and an on-going discussion about localization) suggests that this is not yet fully baked.

### LINKED ISSUE #1086: "security sensitive" items interaction with localization
Updating the manifest
https://w3c.github.io/manifest/#updating

> For the purpose of updating, the following member are security-sensitive members, as they are presented during installation and on launch surfaces:
>
> *    [short_name](https://w3c.github.io/manifest/#dfn-short_name),
> *    [icons](https://w3c.github.io/manifest/#dfn-icons)
> *    [name](https://w3c.github.io/manifest/#dfn-name),

These items are all localizable. If an app adds a new locale to the list of supported locales and this locale matches that of the user, should the app be required to report updating the name/short name/icon? Probably the answer is yes, but I18N wants to call out the interaction. The spec text ***assumes*** that there is only one (unlocalized) value for the name and short name. Also, note that users sometimes change their device locale...

### LINKED ISSUE #1087: shortcuts and localizability
Shortcut items
https://w3c.github.io/manifest/#shortcut-items

Shortcuts are a separate structure inside the manifest that include a `name`, `short_name`, `description`, `icon`, and `url` for some operation that can be invoked on an app or invoke an app with additional arguments. The same concerns about localizability of the shortcuts applies as it does to the `name` and `short_name` at the manifest level, although with the additional note that shortcuts often appear inside the system's user interface (such as context menus) and want to be localized to match the system whenever possible (and that the system locale can be changed by the user after app installation)

### LINKED ISSUE #1088: Application name section might need localization support
Application's name
https://w3c.github.io/manifest/#application-s-name

> The application's name is derived from either the [name](https://w3c.github.io/manifest/#dfn-name) member or [short_name](https://w3c.github.io/manifest/#dfn-short_name) member.

This section describes how the name is assigned from the `name` and `short_name` values (and how to name the application when these are not available). There is no discussion of localization here.

christianliebel:
Closes #1077, closes #1078, closes #1080, closes #1085, closes #1087, closes #1086, closes #1084, closes #1088, closes #676

* [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=276463)

Commit message:

Add members for localization

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1101.html" title="Last updated on Sep 28, 2024, 3:00 PM UTC (c400170)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1101/3e1fd57...c400170.html" title="Last updated on Sep 28, 2024, 3:00 PM UTC (c400170)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -446,29 +446,14 @@
         </h3>
         <p>
           The [=manifest's=] <code><dfn data-export="" data-dfn-for=
-          "manifest">dir</dfn></code> member specifies the default string
-          direction for the <a>localizable members</a> of the <a>manifest</a>.
+          "manifest">dir</dfn></code> member specifies the <dfn>default
+          direction</dfn> for the [=localizable members=] of the [=manifest=].
           The [=manifest/dir=] member's value can be set to a
-          <a>text-direction</a>.
+          [=text-direction=].
         </p>
-        <p>
-          The <dfn>localizable members</dfn> are:
-        </p>
-        <ul>
-          <li>[=manifest/name=] member.
-          </li>
-          <li>[=manifest/short_name=] member.
-          </li>
-          <li>[=Shortcut item's=] [=shortcut item/name=] member.
-          </li>
-          <li>[=Shortcut item's=] [=shortcut item/short_name=] member.
-          </li>
-          <li>[=Shortcut item's=] [=shortcut item/description=] member.
-          </li>
-        </ul>
         <p>
           The <dfn>text-directions</dfn> are the following, implying that the
-          value of the <a>localizable members</a> is by default:
+          value of the [=localizable members=] is by default:
         </p>
         <dl>
           <dt>
@@ -559,9 +544,10 @@
           "ECMA-402#sec-isstructurallyvalidlanguagetag">IsStructurallyValidLanguageTag</a>
           with |json|["lang"] returns `false`, return.
           </li>
-          <li>Set |manifest|["lang"] to the result of calling the
-            <a data-cite="ECMA-402#sec-canonicalizeunicodelocaleid">CanonicalizeUnicodeLocaleId</a>
-            abstract operation with |json|["lang"].
+          <li>[=Map/Set=] |manifest|["lang"] to the result of calling the
+          <a data-cite=
+          "ECMA-402#sec-canonicalizeunicodelocaleid">CanonicalizeUnicodeLocaleId</a>
+          abstract operation with |json|["lang"].
           </li>
         </ol>
       </section>
@@ -575,6 +561,9 @@
           represents the name of the web application as it is usually displayed
           to the user (e.g., amongst a list of other applications, or as a
           label for an icon).
+        </p>
+        <p>
+          The [=manifest/name=] member is a [=localizable member=].
         </p>
         <p>
           The [=manifest/name=] member serves as the <a data-cite=
@@ -596,6 +585,9 @@
           represents a short version of the name of the web application. It is
           intended to be used where there is insufficient space to display the
           full name of the web application.
+        </p>
+        <p>
+          The [=manifest/short_name=] member is a [=localizable member=].
         </p>
         <p class="note" title="Processing the `short_name` member">
           When [=processing a manifest=], the [=process a text member=]
@@ -655,6 +647,9 @@
           list of other applications, or to integrate the web application with
           an <abbr title="Operating system">OS</abbr>'s task switcher and/or
           system preferences.
+        </p>
+        <p>
+          The [=manifest/icons=] member is a [=localizable member=].
         </p>
         <aside class="note">
           <p>
@@ -1238,7 +1233,7 @@
         <ol class="algorithm">
           <li>Let |processedShortcuts:list| be a new [=list=].
           </li>
-          <li>Set |manifest|["shortcuts"] to |processedShortcuts|.
+          <li>[=Map/Set=] |manifest|["shortcuts"] to |processedShortcuts|.
           </li>
           <li>If |json|["shortcuts"] doesn't [=map/exist=] or
           |json|["shortcuts"] is not a [=list=], return.
@@ -1246,7 +1241,8 @@
           <li>[=list/For each=] |entry:ordered map| of |json|["shortcuts"]:
             <ol>
               <li>Let |shortcut:ordered map| be [=process a shortcut=] with
-              |entry| and |manifest|["scope"].
+              |entry|, |manifest URL|, |manifest|["scope"], and
+              |manifest|["dir"].
               </li>
               <li>If |shortcut| is failure, continue.
               </li>
@@ -1266,6 +1262,289 @@
           shortcuts presented in order to remain consistent with the
           conventions or limitations of the host operating system.
         </p>
+      </section>
+      <section>
+        <h3>
+          `*_localized` members
+        </h3>
+        <p>
+          A <dfn>localizable member</dfn> is a [=manifest=] member that can be
+          localized. Each [=localizable member=] of the [=manifest=] has a
+          corresponding <code><dfn data-export="" data-dfn-for=
+          "manifest">*_localized</dfn></code> member, where `*` represents the
+          member name.
+        </p>
+        <aside class="example">
+          <p>
+            For example, the [=manifest/name=] member is a [=localizable
+            member=], and `name_localized` is its corresponding
+            [=manifest/*_localized=] member.
+          </p>
+        </aside>
+        <p>
+          A <dfn>language map</dfn> is an [=ordered map=] whose key is a
+          [=language tag=] and whose value is a [=localized value=]. The
+          <dfn>localized value</dfn> is content localized in the language given
+          by the key.
+        </p>
+        <aside class="example" title="Localizing the application name">
+          <pre class="json">
+            {
+              "lang": "en-US",
+              "dir": "ltr",
+              "name": "Color Picker",
+              "name_localized": {
+                "de":    "Farbwähler",
+                "en":    {"value": "Color Picker"},
+                "en-GB": {"value": "Colour Picker", "dir": "ltr"},
+                "fr":    {"value": "Sélecteur de Couleur", "lang": "fr-CA", "dir": "ltr"},
+                "ar":    {"value": "منتقي الألوان", "dir": "rtl"}
+              }
+            }
+          </pre>
+        </aside>
+        <p>
+          The value assigned to the [=localizable member=] is the
+          <dfn data-dfn-for="localizable member">default representation</dfn>.
+          [=manifest/*_localized=] members contain a [=language map=] that
+          defines [=localized values=] for the given [=localizable member=] in
+          the application. The user agent SHOULD use the user's localization
+          settings to select the [=localized value=] whose [=language tag=] key
+          best matches the user's preference. When no such [=localized value=]
+          is available, the [=localizable member/default representation=] is
+          used.
+        </p>
+        <section>
+          <h3>
+            Localizing text values
+          </h3>
+          <p>
+            A <dfn>localized text object</dfn> is an [=ordered map=] with the
+            following properties:
+          </p>
+          <dl>
+            <dt>
+              <code><dfn data-export="" data-dfn-for=
+              "manifest/*_localized">value</dfn></code>
+            </dt>
+            <dd>
+              The localized [=string=].
+            </dd>
+            <dt>
+              <code><dfn data-export="" data-dfn-for=
+              "manifest/*_localized">dir</dfn></code> (optional)
+            </dt>
+            <dd>
+              The [=text-direction=].
+            </dd>
+            <dt>
+              <code><dfn data-export="" data-dfn-for=
+              "manifest/*_localized">lang</dfn></code> (optional)
+            </dt>
+            <dd>
+              A [=language tag=].
+            </dd>
+          </dl>
+          <p>
+            For [=localizable members=] that accept [=strings=], the
+            [=manifest/*_localized=] member's [=language map=] accepts either a
+            [=string=] or a [=localized text object=] as the [=localized
+            value=].
+          </p>
+          <p>
+            When a [=string=] is used, or when the [=manifest/*_localized/dir=]
+            member of the [=localized text object=] is missing, the [=default
+            direction=] ([=manifest/dir=] member of the [=manifest=]) is
+            applied.
+          </p>
+          <aside class="note">
+            The [=manifest/*_localized/dir=] member of the [=localized text
+            object=] needs to be present if the direction of a localized string
+            differs from the [=default direction=] set in the manifest
+            ([=manifest/dir=] member). Right-to-left text will require specific
+            direction settings if the manifest's default direction is
+            left-to-right, and vice versa.
+          </aside>
+          <p>
+            When a [=string=] is used, or when the
+            [=manifest/*_localized/lang=] member of the [=localized text
+            object=] is missing, the [=language tag=] of the [=language map=]
+            key is applied.
+          </p>
+          <aside class="note">
+            To support multilingual content and ensure optimal display and
+            accessibility, it is possible to specify a different language for a
+            [=localized text object=]. This is needed for situations where a
+            term or text needs to be presented in a language different from the
+            user's set locale.
+          </aside>
+          <aside class="example">
+            <p>
+              For example, brand names might need to be pronounced in a
+              language different from the user's locale:
+            </p>
+            <pre class="json">
+            {
+              "lang": "fr",
+              "name": "Superbes biscuits",
+              "name_localized": {
+                "de-DE": {"value": "Super Cookies", "lang": "en"}
+              }
+            }
+          </pre>
+          </aside>
+          <p>
+            To <dfn>process a `*_localized` text member</dfn>, given [=ordered
+            map=] |json:ordered map|, [=ordered map=] |map:ordered map|,
+            [=string=] |member:string|, and [=text-direction=]
+            |defaultDirection:string|:
+          </p>
+          <ol class="algorithm">
+            <li>If |member| does not [=map/exist=] in |json|, return.
+            </li>
+            <li>Let |languageMap| be |json|[|member|].
+            </li>
+            <li>If |languageMap| is not an [=ordered map=], return.
+            </li>
+            <li>Let |languageTags:ordered set| be the [=map/keys=] of
+            |languageMap:ordered map|.
+            </li>
+            <li>Set |map|[|member|] to a new [=ordered map=].
+            </li>
+            <li>[=Set/For each=] |languageTag:string| of |languageTags|, run
+            [=process a localized text object=], passing
+            |languageMap|[|languageTag|], |languageTag|, |map|, |member|, and
+            |defaultDirection|.
+            </li>
+          </ol>
+          <p>
+            To <dfn>process a localized text object</dfn>, given [=string=] or
+            [=ordered map=] |localizedValue|, [=string=]
+            |defaultLanguageTag:string|, [=ordered map=] |map:ordered map|,
+            [=string=] |member:string|, and [=text-direction=]
+            |defaultDirection:string|:
+          </p>
+          <ol class="algorithm">
+            <li>Let |normalizedValue:ordered map| be an [=ordered map=].
+            </li>
+            <li>If |localizedValue| is a [=string=], [=strip leading and
+            trailing ASCII whitespace=] from |localizedValue:string| and
+            [=map/set=] |normalizedValue|["value"] to |localizedValue|.
+            </li>
+            <li>If |localizedValue:ordered map| is an [=ordered map=]:
+              <ol>
+                <li>If "value" [=map/exists=] in |localizedValue| and
+                |localizedValue|["value"] is a [=string=], [=strip leading and
+                trailing ASCII whitespace=] from |localizedValue|["value"] and
+                [=map/set=] |normalizedValue|["value"] to
+                |localizedValue|["value"].
+                </li>
+                <li>If "lang" [=map/exists=] in |localizedValue| and
+                |localizedValue|["lang"] is a [=string=], [=strip leading and
+                trailing ASCII whitespace=] from |localizedValue|["lang"] and
+                [=map/set=] |normalizedValue|["lang"] to
+                |localizedValue|["lang"].
+                </li>
+                <li>If "dir" [=map/exists=] in |localizedValue| and
+                |localizedValue|["dir"] is a [=string=]:
+                  <ol>
+                    <li>[=Strip leading and trailing ASCII whitespace=] from
+                    |localizedValue|["dir"].
+                    </li>
+                    <li>If [=text-direction list=] [=list/contains=]
+                    |localizedValue|["dir"], [=map/set=]
+                    |normalizedValue|["dir"] to |localizedValue|["dir"].
+                    </li>
+                  </ol>
+                </li>
+              </ol>
+            </li>
+            <li>If "value" does not [=map/exist=] in |normalizedValue|, return.
+            </li>
+            <li>If "lang" does not [=map/exist=] in |normalizedValue|,
+            [=map/set=] |normalizedValue|["lang"] to |defaultLanguageTag|.
+            </li>
+            <li>If "dir" does not [=map/exist=] in |normalizedValue|,
+            [=map/set=] |normalizedValue|["dir"] to |defaultDirection|.
+            </li>
+            <li>If calling <a data-cite=
+            "ECMA-402#sec-isstructurallyvalidlanguagetag">IsStructurallyValidLanguageTag</a>
+            with |normalizedValue|["lang"] or calling <a data-cite=
+            "ECMA-402#sec-isstructurallyvalidlanguagetag">IsStructurallyValidLanguageTag</a>
+            with |defaultLanguageTag| returns `false`, return.
+            </li>
+            <li>[=Map/Set=] |map|[|member|][|defaultLanguageTag|] to
+            |normalizedValue|.
+            </li>
+          </ol>
+          <p class="note">
+            The [=process a localized text object=] algorithm takes both a
+            [=string=] or a [=localized text object=] for the [=localized
+            value=] parameter, but the processed result will be normalized into
+            a [=localized text object=] with the
+            [=manifest/*_localized/value=], [=manifest/*_localized/lang=], and
+            [=manifest/*_localized/dir=] members set.
+          </p>
+        </section>
+        <section>
+          <h3>
+            Localizing image resources
+          </h3>
+          <p>
+            For [=localizable members=] that accept a [=list=] of [=image
+            resources=], the [=manifest/*_localized=] member's [=language map=]
+            accepts a [=list=] of [=image resources=] as the [=localized
+            value=].
+          </p>
+          <aside class="example" title="Localizing the application icon">
+            <pre class="json">
+            {
+              "lang": "en-US",
+              "icons": [
+                { "src": "icon/lowres.png", "sizes": "64x64" },
+                { "src": "icon/hires.png", "sizes": "256x256" }
+              ],
+              "icons_localized": {
+                "fr": [
+                  { "src": "icon/lowres_fr.png", "sizes": "64x64" },
+                  { "src": "icon/hires_fr.png", "sizes": "256x256" }
+                ]
+              }
+            }
+          </pre>
+          </aside>
+          <p>
+            To <dfn>process a `*_localized` image resource member</dfn>, given
+            [=ordered map=] |json:ordered map|, [=ordered map=] |map:ordered
+            map|, [=string=] |member:string|, and [=URL=] |manifest URL:URL|:
+          </p>
+          <ol class="algorithm">
+            <li>If |member| does not [=map/exist=] in |json|, return.
+            </li>
+            <li>Let |languageMap| be |json|[|member|].
+            </li>
+            <li>If |languageMap| is not an [=ordered map=], return.
+            </li>
+            <li>Let |languageTags:ordered set| be the [=map/keys=] of
+            |languageMap:ordered map|.
+            </li>
+            <li>[=Map/Set=] |map|[|member|] to a new [=ordered map=].
+            </li>
+            <li>[=List/For each=] |languageTag:string| of |languageTags|:
+              <ol>
+                <li>If calling <a data-cite=
+                "ECMA-402#sec-isstructurallyvalidlanguagetag">IsStructurallyValidLanguageTag</a>
+                with |languageTag| returns `false`, [=iteration/continue=].
+                </li>
+                <li>Run [=process image resources=], passing
+                |languageMap|[|languageTag|] as the [=list=] of [=image
+                resources=], |map|[|member|], |manifest URL|, and |languageTag|
+                as the member.
+                </li>
+              </ol>
+            </li>
+          </ol>
+        </section>
       </section>
       <section>
         <h2>
@@ -1328,8 +1607,14 @@
             <li>[=Process a text member=] passing |json|, |manifest|, and
             "name".
             </li>
+            <li>[=Process a `*_localized` text member=] passing |json|,
+            |manifest|, "name_localized", and |manifest|["dir"].
+            </li>
             <li>[=Process a text member=] passing |json|, |manifest|, and
             "short_name".
+            </li>
+            <li>[=Process a `*_localized` text member=] passing |json|,
+            |manifest|, "short_name_localized", and |manifest|["dir"].
             </li>
             <li>[=Process the `start_url` member=] passing |json|, |manifest|,
             |manifest URL|, and |document URL|.
@@ -1353,6 +1638,9 @@
             </li>
             <li>[=Process image resources=] passing |json|["icons"],
             |manifest|, |manifest URL|, and "icons".
+            </li>
+            <li>[=Process a `*_localized` image resource member=] passing
+            |json|, |manifest|, "icons_localized", and |manifest URL|.
             </li>
             <li>[=Process the `orientation` member=] passing |json|,
             |manifest|.
@@ -1540,13 +1828,21 @@
             installation and on launch surfaces:
           </p>
           <ol>
-            <li>[=manifest/short_name=],
+            <li>[=manifest/short_name=] and localized representations in
+            `short_name_localized`,
             </li>
-            <li>[=manifest/icons=]
+            <li>[=manifest/icons=] and localized representations in
+            `icons_localized`,
             </li>
-            <li>[=manifest/name=],
+            <li>[=manifest/name=] and localized representations in
+            `name_localized`.
             </li>
           </ol>
+          <p>
+            [=Security-sensitive members=] SHOULD be displayed in a
+            bidirectionally isolated way as described in [[UTS55]], regardless
+            of their direction.
+          </p>
           <p data-cite="permissions">
             User agents SHOULD NOT automatically apply changes to
             [=security-sensitive members=] without [=express permission=] from
@@ -1560,6 +1856,13 @@
           <p>
             The user agent MAY automatically apply the changes if the update
             does not contain changes to [=security-sensitive members=].
+          </p>
+          <p>
+            If a user changes localization settings, the user agent MAY
+            automatically adjust the [=security-sensitive members=] visible on
+            launch surfaces to their localized representations specified in the
+            [=manifest/`*_localized`=] members. These changes SHOULD be
+            presented to users the next time they open the web application.
           </p>
           <aside class="note" title=
           "A user agent will not apply a partial update">
@@ -2037,6 +2340,9 @@
           represents the name of the shortcut as it is usually displayed to the
           user in a context menu.
         </p>
+        <p>
+          The [=shortcut item/name=] member is a [=localizable member=].
+        </p>
       </section>
       <section>
         <h3>
@@ -2049,6 +2355,9 @@
           intended to be used where there is insufficient space to display the
           full name of the shortcut.
         </p>
+        <p>
+          The [=shortcut item/short_name=] member is a [=localizable member=].
+        </p>
       </section>
       <section>
         <h3>
@@ -2059,6 +2368,9 @@
           "shortcut item">description</dfn></code> member is a <a>string</a>
           that allows the developer to describe the purpose of the shortcut.
           User agents MAY expose this information to assistive technology.
+        </p>
+        <p>
+          The [=shortcut item/description=] member is a [=localizable member=].
         </p>
       </section>
       <section>
@@ -2081,6 +2393,9 @@
           "shortcut item">icons</dfn></code> member lists images that serve as
           iconic representations of the shortcut in various contexts.
         </p>
+        <p>
+          The [=shortcut item/icons=] member is a [=localizable member=].
+        </p>
       </section>
       <section>
         <h3>
@@ -2098,7 +2413,8 @@
         </h2>
         <p>
           To <dfn>process a shortcut</dfn>, given [=ordered map=] |item:ordered
-          map| and |scope:URL|,:
+          map|, [=URL=] |manifest URL:URL|, [=URL=] |scope:URL|, and
+          [=text-direction=] |defaultDirection:string|:
         </p>
         <ol class="algorithm">
           <li>Return failure if it's the case that:
@@ -2126,8 +2442,28 @@
           <li>Let |shortcut:ordered map| be |ordered map| «[ "url" → |url|,
           "name" → |item|["name"] ]».
           </li>
+          <li>[=Process a `*_localized` text member=] passing |item|,
+          |shortcut|, "name_localized", and |defaultDirection|.
+          </li>
+          <li>If "short_name" [=map/exists=] in |item|, and
+          |item|["short_name"] is a [=string=], [=map/set=]
+          |shortcut|["short_name"] to |item|["short_name"].
+          </li>
+          <li>[=Process a `*_localized` text member=] passing |item|,
+          |shortcut|, "short_name_localized", and |defaultDirection|.
+          </li>
+          <li>If "description" [=map/exists=] in |item|, and
+          |item|["description"] is a [=string=], [=map/set=]
+          |shortcut|["description"] to |item|["description"].
+          </li>
+          <li>[=Process a `*_localized` text member=] passing |item|,
+          |shortcut|, "description_localized", and |defaultDirection|.
+          </li>
           <li>[=Process image resources=] passing |item|["icons"], |shortcut|,
           |manifest URL|, and "icons".
+          </li>
+          <li>[=Process a `*_localized` image resource member=] passing |item|,
+          |shortcut|, "icons_localized", and |manifest URL|.
           </li>
           <li>Return |shortcut|.
           </li>
@@ -2316,7 +2652,9 @@
         </h3>
         <p>
           The <dfn>application's name</dfn> is derived from either the
-          [=manifest/name=] member or [=manifest/short_name=] member.
+          [=manifest/name=] member or [=manifest/short_name=] member. The user
+          agent SHOULD first resolve the localized values from their
+          corresponding [=manifest/`*_localized`=] members.
         </p>
         <p>
           When either the [=manifest/name=] member or the
@@ -3303,7 +3641,8 @@
         </details>
       </aside>
     </section>
-    <section id="internationalization" class="appendix informative">
+    <section id="internationalization" class="appendix informative" data-cite=
+    "i18n-glossary">
       <h2>
         Internationalization
       </h2>
@@ -3312,6 +3651,16 @@
         using one of the following options:
       </p>
       <dl>
+        <dt>
+          Localized values in the manifest:
+        </dt>
+        <dd>
+          Authors can provide [=localized values=] for the [=localizable
+          members=] of the [=manifest=]. The user agent passes all localized
+          values to the host operating system. When the user changes the
+          [=locale=] at the OS level, the OS can present the updated localized
+          values of the [=installed web application=].
+        </dd>
         <dt>
           Dynamically setting the language:
         </dt>


aarongustafson:
I'd prefer adding the "_localized"  suffix across the board as it makes it explicit what the member's purpose is.

marcoscaceres:
Yeah, I guess it does make sense to make it "_localized" as this can only be used for that.

Let's just do the text ones initially.

dmurph:
From editors meeting:

- We want to also localize the icons, but because it already has an `s` that makes it a little weird
- Due to that, let's do `*_localized` for all of these fields.
- Icons_localized won't have any triplets, we'll just resuse the parsing algorithm for icons (same json structure). But, like the others, it will be lang-string dictionary.

marcoscaceres:
Instead of defining new members, let's instead define `*_localized` member pattern that is either a "text localizable member" or an "image-resource localizable member" (i.e., icons or for shortcuts). That way we don't need to define the algorithms over and over again.

marcoscaceres:
We already have defined [localizable members](https://www.w3.org/TR/appmanifest/#dfn-localizable-members) so we can already reuse that.

marcoscaceres:
Here is what an shortcuts member might look like with some _localized members sprinkled in:

```JSON
{
  "shortcuts": [
    {
      "name": "Play Later",
      "name_localized": {
        "fr": "Écouter plus tard"
      },
      "description": "View the list of podcasts you saved for later",
      "description_localized": {
        "fr": "Consultez la liste des podcasts que vous avez enregistrés pour plus tard.",
      },
      "url": "/play-later",
      "icons": [
        {
          "src": "/icons/play-later.svg",
          "type": "image/svg+xml"
        }
      ]
    },
    {
      "name": "Subscriptions",
      "description": "View the list of podcasts you listen to",
      "description_localized": {
        "fr": "Consultez la liste des podcasts que vous écoutez."
      },
      "url": "/subscriptions?sort=desc"
    }
  ]
}
```

marcoscaceres:
Question is if we should allow localizing various URLs... that gets a bit messy in places, but could be doable.

christianliebel:
@marcoscaceres @dmurph Thanks. Let's also localize `url`s. I'll resume working on this soon.

mgiuca:
LTTP... thanks for working on this Christian!

> Let's also localize urls. I'll resume working on this soon.

Wait, why are we localizing URLs? That seems undesirable to me. It means that shortcuts can go to different places depending on the language, and we need to update functionality that may be cached (that isn't just string/icon data). This may add non-trivial implementation complexity (it would require a deeper analysis to understand to what degree).

Would the use case for this be that you can have URLs with different `?lang=` query parameters to match the user's language? I think I'd prefer just letting the website use the `Accept-Language` HTTP feature for this.

christianliebel:
> Wait, why are we localizing URLs? That seems undesirable to me.

Maybe we can also loop in @aphillips to see whether it makes sense to localize URLs.

aphillips:
Icons, graphics, or remote content (help pages, for example) are sometimes varied by locale or by region (with locale serving as a poor proxy for region). This might be done, for example, because the icon contains some text or because a graphic shows a culturally-linked image (personal images, national costume, post box shapes, etc. etc.) that the user wishes to localize. Or it might be because functionality or defaults differ (sorting based on pronunciation instead of name for Chinese, for example)

We don't know why the user might want to localize the icon or shortcut (or whatever).

I agree that this can be abused and there might be reasons not to allow some fields to be localized, although I'd probably thinks about health warnings first?

aarongustafson:
> I think I'd prefer just letting the website use the `Accept-Language` HTTP feature for this.

Probably the most elegant, but it’s not within the realm of possibility for a lot of orgs and site types (thinking static sites, for example).

> Would the use case for this be that you can have URLs with different ?lang= query parameters to match the user's language?

See also MDN style links where the language code is embedded in the URL path.

> Icons, graphics, or remote content (help pages, for example) are sometimes varied by locale or by region (with locale serving as a poor proxy for region). This might be done, for example, because the icon contains some text or because a graphic shows a culturally-linked image (personal images, national costume, post box shapes, etc. etc.) that the user wishes to localize. Or it might be because functionality or defaults differ (sorting based on pronunciation instead of name for Chinese, for example)
> 
> We don't know why the user might want to localize the icon or shortcut (or whatever).

Agree on all of these.

benfrancis:
Just one data point, but as a past precedent the similar [Web of Things (WoT) Thing Description](https://www.w3.org/TR/wot-thing-description) specification landed on `titles` and `descriptions` members of [`Thing`](https://www.w3.org/TR/wot-thing-description/#thing) (each of type [`MultiLanguage`](https://www.w3.org/TR/wot-thing-description/#multilanguage)) in addition to `title` and `description` for this use case. This is the case for both Thing Description 1.0 (W3C Recommendation) and Thing Description 1.1 (Proposed Recommendation).

I personally don't like that solution because I prefer using HTTP content negotiation with an `Accept-Language` header as per the suggestion in the current Web Application Manifest Working Draft, rather than creating an extremely verbose manifest with (theoretically) up to thousands of different languages. However, as has been pointed out it's not always possible to use content negotiation (e.g. on static site hosting like GitHub Pages). The Thing Description specification therefore offers both as alternative approaches.

If consistency between W3C specifications is considered important, then `names`, `short_names` and `descriptions` would make sense. That doesn't work for `icons`, but that member is already different because it's an array of `ImageResource`s that the user agent can select from. Language could potentially just be another criteria for selecting an image.

I note that in HTML the `<link>` element has a `hreflang` attribute, so presumably `<link rel="icon" href="/icons/fr/play-later.svg" hreflang="fr">` is valid (though likely currently ignored by user agents). For the (slightly unusual) case of localising app icons, an equivalent might be to add a `lang` member to [`ImageResource`](https://www.w3.org/TR/image-resource).

Example:
```json
{
  "lang": "en",
  "dir": "ltr",
  "name": "Super Racer 3000",
  "names": {
    "fr-FR": "Super Coureur 3000",
    "es-ES": "Súper Corredor 3000"
  },
  "short_name": "Racer3K",
  "short_names": {
    "fr-FR": "Coureur3K",
    "es-ES": "Corredor3K"
  },
  "icons": [{
    "src": "icon.png",
    "sizes": "64x64",
    "type": "image/webp",
  }, {
    "src": "icon-fr.png",
    "sizes": "64x64",
    "lang": "fr-FR"
  },
  {
    "src": "icon-es.png",
    "sizes": "64x64",
    "lang": "es-ES"
  }],
  "scope": "/",
  "id": "superracer",
  "start_url": "/start.html",
  "display": "fullscreen",
  "orientation": "landscape",
  "theme_color": "aliceblue",
  "background_color": "red"
}
```

Note that `ImageResource` also has a `label` member which could then also be localised this way, if a localised accessible description of the icon is needed!

One question: How does `dir` interact with the localised members? Can it safely be derived from language? The Thing Description specification has a lot to say on that topic, which I can't say I fully understand, but they use the [Strings on the Web: Language and Direction Metadata](https://www.w3.org/TR/string-meta/) W3C Note for guidance.

Hope this helps.

mgiuca:
Perhaps we should clarify what "localize URLs" means.

Are we talking about:
* Localizing all fields that are URLs? (e.g. `start_url`, `scope`, potential future ones like the home scope of tabbed apps).
* Localizing all fields called "`url`", like the ones in icons and shortcuts?
* Something else?

I think we generally agree that we need to localize icons, but we're doing that at the `icons` level, not the `url` within icons (i.e. `icons_localized` with a local version of each icon dict, not `icons` with the icon dict including a `url_localized`. Allowing both of these creates two ways to do it which isn't ideal.)

Maybe it makes sense for URLs like shortcuts to be able to change based on language, but really the point of this initiative (I thought) was to be able to localize your app's metadata that gets displayed at the OS level, like name and icon, not to solve the problem of localizing all the content in the app. (If you can't configure your server to serve content based on headers, you can still make your service worker return localized content based on Accept-Language.) An app that relies purely on the manifest URLs to display content in the user's language is likely to be quite brittle.

I think we could run into major headaches if we allow `scope` to be localized. And by extension, `start_url`. (e.g. `scope` must be a superset of `start_url` - what happens if that's true in some languages but not others?)

I prefer if we just start with `name`, `short_name`, `description` and `icons` and go from there as needed.

> How does `dir` interact with the localised members? Can it safely be derived from language?

This was [discussed at TPAC](https://docs.google.com/document/d/1RIaeXT_-j8n4iGPI3odZyJTqtHCKqe3oWZTO-j21z9U/edit#heading=h.5i48hfw584ma) (search for "dir") - unfortunately when I asked this question, the answer was not recorded ("?"). From memory, @aphillips pointed out that we may not know the language's direction because languages are not set in stone. I think the overwhelmingly common case will be a known language, which means we should be able to derive `dir` from `lang` in all practical cases (and probably default to `ltr` if we don't recognize the language - the overwhelming majority of languages are LTR). We should have the `dir` member for being explicit, but I think in 99.9% of cases the site should not need to specify `dir` as it can be derived from language.

marcoscaceres:
Supportive of what @mgiuca said above... let's start small and go from there (and definitely let's not have multiple ways of doing the same thing, specially with URLs). And yes, let's keep the localizable members restricted to a small set (including image objects, not members within those objects).

dmurph:
Manifest Working session notes:

This seems to be the concluded format (copied from Marcos's comment above) with triple examples:
```
{
  "shortcuts": [
    {
      "name": "Play Later",
      "name_localized": {
        "fr": "Écouter plus tard"
      },
      "description": "View the list of podcasts you saved for later",
      "description_localized": {
         "en":  { "value": "My App, hey!", "dir": "ltr", "lang": "fr"},
         "en-GB": { "value": "My App, eh wut?", "dir": "ltr"},
         "fr": "string",
         "ar": { "value": "...", "dir": "rtl" }
      },
      "url": "/play-later",
      "icons": [
        {
          "src": "/icons/play-later.svg",
          "type": "image/svg+xml"
        }
      ],
      "icons_localized": {
        "fr": [
          {
            "src": "/icons/fr/play-later.svg",
            "type": "image/svg+xml"
          }
        ]
      } 
    },
    {
      "name": "Subscriptions",
      "description": "View the list of podcasts you listen to",
      "description_localized": {
        "fr": "Consultez la liste des podcasts que vous écoutez."
      },
      "url": "/subscriptions?sort=desc"
    }
  ]
}

```

TPAC notes from this are [here](https://docs.google.com/document/d/1RIaeXT_-j8n4iGPI3odZyJTqtHCKqe3oWZTO-j21z9U/edit#heading=h.5i48hfw584ma)

tomayac:
I suppose the stray `"lang": "fr"` in your [comment before](https://github.com/w3c/manifest/pull/1101#:~:text=%22ltr%22-,%2C%20%22lang%22%3A%20%22fr%22,-%7D%2C%0A%20%20%20%20%20%20%20%20%20%22en%2DGB) isn't intended:

```json
"en":  { "value": "My App, hey!", "dir": "ltr", "lang": "fr"},
```

I could edit your comment directly, but wanted to make sure it's indeed a copy/paste first.

dmurph:
> I suppose the stray `"lang": "fr"` in your [comment before](https://github.com/w3c/manifest/pull/1101#:~:text=%22ltr%22-,%2C%20%22lang%22%3A%20%22fr%22,-%7D%2C%0A%20%20%20%20%20%20%20%20%20%22en%2DGB) isn't intended:
> 
> ```json
> "en":  { "value": "My App, hey!", "dir": "ltr", "lang": "fr"},
> ```
> 
> I could edit your comment directly, but wanted to make sure it's indeed a copy/paste first.

This actually is intended - the example might not be great, but we need the ability for a string to be displayed in a different language A when showing for language B. For example, if a product name or logo etc was a character in a different language, the dev can make sure it renders in the desired language for the user's chosen display language.

marcoscaceres:
@christianliebel can you check https://github.com/w3c/manifest/issues/676 ... and update the description of this issue as closing that bug, as well as any other bugs this will close?

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
         </p>
       </section>
+      <section>
+        <h3>
+          `*_localized` members
+        </h3>
+        <p>
+          A <dfn>localizable member</dfn> is a [=manifest=] member that can be
+          localized. Each [=localizable member=] of the [=manifest=] has a
+          corresponding <dfn>localized member</dfn> with the suffix
```
aphillips, line 1280: This is really hard to read, in that `localizable member` and `localized member` are super similar looking. Consider separating the terms somewhat more?

> A `localizable member` is a _manifest_ member that can be localized.
> This can include translation or other locale-based customization.
> When present, the localized representations of a `localizable member`
> appear in a corresponding member with the suffix `_localized`. ...

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
+          [=localizable member=], where `name_localized` is the corresponding
+          [=localized member=].
+        </p>
+        <p>
+          [=Localized members=] contain a [=localization map=] that defines
+          [=localized values=] for different locales denoted by a [=language
+          tag=]. If present and if the [=localization map=] contains a
+          [=localized value=] that matches the user's localization settings,
+          the user agent SHOULD override the value of the [=localizable
+          member=].
```
aphillips, line 1291: I would probably explain this differently? For one thing, I think it is not helpful to talk about the localizations "overriding" the value. It's not really an override: it **_is_** the value. The `name` (etc.) is the _default_, used when no applicable localized value is available. The default has a locale associated with it (represented by the `lang` member), but the application does not do language matching against this value. 

Perhaps something like:

```suggestion
A [=localized member=] contains a [=localization map=] that provides
[=localized values=] for the given [=localizable member=] in the application.
Each [=localized value=] is identified by a [=language tag=].
The user agent SHOULD use the user's localization settings to select
the [=localized value=] whose language tag best matches the user's preference.
When no such [=localized value=] is available, the default representation is used.
```

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
+        </p>
+        <aside class="example" title=
+        "Overriding the application name with localized values">
```
aphillips, line 1294: As noted, I'd prefer not to talk about "overriding". Perhaps:

```suggestion
        "Localizing the application name">
```

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
+          <pre class="json">
+            {
+              "lang": "en-US",
+              "name": "Color Picker",
+              "name_localized": {
+                "en-GB": "Colour Picker",
+                "fr": "Sélecteur de Couleur"
```
aphillips, line 1301: I think you'd be better off showing the whole enchillada up front? (My localized example is stolen verbatim from [string-meta](https://w3c.github.io/string-meta/#language-maps) and is meant to be illustrative only)

```json
"lang": "en-US",
"dir": "ltr",
"name": "Color Picker",
"name_localized": {
    "en":    {"value": "This is English"},
    "en-GB": {"value": "This is UK English", "dir": "ltr"},
    "fr":    {"value": "C'est français", "lang": "fr-CA", "dir": "ltr"},
    "ar":    {"value": "هذه عربية", "dir": "rtl"}
}
```

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
+          its corresponding [=localizable member=]. For example, a [=localized
+          value=] for the [=manifest/name=] member could directly be a
+          [=string=].
+        </p>
+        <p class="note">
+          To support multilingual content and ensure proper display and
+          accessibility, it is possible to specify a different language tag and
+          text direction for a localized value. This is needed for situations
+          where a term or text must be presented in a language different from
+          the user's set language.
```
aphillips, line 1320: This covers the language tag not matching the language map key case (which is rare). However, the far more common requirement is that the direction of the translation does not match the default direction of the application. For example, every right-to-left value will need this when the default language of the application is a left-to-right language (or vice versa).

As a result, I would leave out "text direction" in the note and mention direction separately.

index.html
```diff
@@ -2089,6 +2231,10 @@ <h3>
           "shortcut item">icons</dfn></code> member lists images that serve as
           iconic representations of the shortcut in various contexts.
         </p>
+        <p>
+          This member is a [=localizable member=]. Setting a [=text-direction=]
+          is not supported and will be ignored.
```
aphillips, line 2236: see above comment

index.html
```diff
@@ -1280,6 +1270,158 @@ <h3>
...
+          <li>Let |localizedMemberName| be the [=string/concatenation=] of the
+          [=list=] « |localizableMemberName|, "_localized" ».
+          </li>
+          <li>If |localizedMemberName| [=map/exists=] in |json|:
+            <ol>
+              <li>Let |localizedMember| be |json|[|localizedMemberName|].
+              </li>
+              <li>If |localizedMember| is not an [=ordered map=], continue.
+              </li><!-- todo: any existing algorithm for this? -->
+              <!-- todo: also take manifest's own lang into account -->
```
aphillips, line 1406: What's wanted here is `Intl.LocaleMatcher`, which is not yet beyond the proposal stage.

Language tag matching is complicated. It is case _insensitive_ and there can be some quirks if one wants the "best fit" style of matching. If one is okay with BCP47 Lookup matching, things are somewhat simpler: one selects the _longest_ tag that is a prefix of the user's requested value (or the default if no tag is a prefix). BCP47 [describes this](https://www.rfc-editor.org/rfc/rfc4647.html#section-3.4) in terms of truncating the requested tag until a match is found (or use the default if no match is found):

```
   Range to match: zh-Hant-CN-x-private1-private2
   1. zh-Hant-CN-x-private1-private2
   2. zh-Hant-CN-x-private1
   3. zh-Hant-CN
   4. zh-Hant
   5. zh
   6. (default)
```

Note that the manifest's own language should not participate in this. It has a language associated with it, but it is the "root locale" default value for purposes of language negotiation.

This could be a bit weird, since the localized value might override the manifest's default language (e.g. the default language is never visible)

index.html
```diff
@@ -675,6 +661,10 @@ <h3>
           an <abbr title="Operating system">OS</abbr>'s task switcher and/or
           system preferences.
         </p>
+        <p>
+          This member is a [=localizable member=]. Setting a [=text-direction=]
+          is not supported and will be ignored.
```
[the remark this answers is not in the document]
christianliebel: I wonder whether this scenario should be supported at all:

```
{
  "icons_localized": {
    "fr": { "lang": "de", "dir": "rtl", value: [] }
  }
}
```

Alternatively, we could also distinguish between text localizations (where we would support changing `lang`/`dir`) and image resource localizations where we would *only* allow setting the usual member value (image resource array):

```
{
  "icons_localized": {
    "fr": []
  }
}
```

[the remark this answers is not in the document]
aphillips: There might be a use case for generating a tag like this:

```html
<img lang="de" dir="ltr" src="..." alt="Hier ist Deutsch">
```

Of course, the `alt` attribute has to come from a separate string somewhere. I guess the application icon generally will use the `name` as it's `alt`.

calidion:
hello, everyone

I would suggest that localization be a native feature for all strings. 

Hence we don't need extra fields for localization.

just to enhance the parser to parse new strings.

My suggestion is that strings can divided into two types: 

1. the normal primitive strings defined by different charsets
```
 "Hello world!"
or
"你好世界！“
```
2. the enhanced strings which include i18n features and be in json format, like this:
```
```json
```
lang: "en",   // fallback language if browser meets no locale strings listed
"name":  {
  "en":  "Web App",    // fallback for all en-* browsers or for general  en-* users
  "en-US":  "Web App",   // specific string for localized en
  "zh": "网站应用”， // fallback for all zh-* browsers  or for general  zh-* users
  "zh-CN": "网站应用”， // specific string for localized zh
  ...
}
```

**Comment by calidion** [comment]
@christianliebel 
I think `_localized` suffix can be a temporary solution to be backward compatible with existing browsers and applications.
while native localizable strings can also be a standard way for all web standards like html can be enhanced like this: 

```html
<p>
<string>
<locale lang="en"> Hello World!</locale>
<locale lang="en-US"> Hello World!</locale>
<i18n lang="zh"> 你好世界！</i18n>
<i18n lang="zh-CN"> 你好世界！</i18n>
</string>
</p>
````

**Comment by dmurph** [comment]
@calidion said:

> I think `_localized` suffix can be a temporary solution to be backward compatible with existing browsers and applications.

This is a good idea! This is what we settled on at TPAC - see my comment [above](https://github.com/w3c/manifest/pull/1101#issuecomment-2091849671).

To be compatible with what is needed for I18N, for each localized field, we need the local language lookup key, which resolves either  a simple string that is used, or a object with "value", "dir", and "lang". Likely optional for "dir" and "lang".

**Comment by calidion** [comment]
@christianliebel @dmurph 

There may also be a new way like extending manifest when included by the html file.

currently manifest can be included this way:
```html
<link rel="manifest" href="manifest.json" />
```
but for i18n support, it can be extended this way:

```html
<link rel="manifest"  type="locale" href="i18n/manifest.fr.json" />
<link rel="manifest"  type="locale" href="i18n/manifest.fr-ca.json" />
<link rel="manifest"  type="locale" href="i18n/manifest.en.json" />
<link rel="manifest"  type="locale" href="i18n/manifest.en-us.json" />
```

where locale typed manifest can override all features of the default one according to the user agent settings.

this method may be a better way for both compatibility and extensibility.

dmurph:
@calidion Unfortunately that doesn't solve the need to have per-field "dir" and "lang". Segmenting manifest files also means lots of inconsistency risk between them. Also, a single file allows the OS to have all info if the user changes their language offline etc. The existing solution solves all of the use cases very well, I'm not sure it's worth continuing to rehash.

calidion:
@dmurph 

per-field "dir" and "lang" seems not reasonable feature for a manifest file needing language consistency.
The dir problem can be solved easily use link tag like the following:
```html
<link rel="manifest"  type="locale" dir = "ltr" lang="fr" href="i18n/manifest.fr.json" />
<link rel="manifest"  type="locale" dir = "rtl" lang="fr-ca" href="i18n/manifest.fr-ca.json" />
<link rel="manifest"  type="locale" dir = "auto" lang="en" href="i18n/manifest.en.json" />
<link rel="manifest"  type="locale" dir = "ltr" lang="en-us" href="i18n/manifest.en-us.json" />
<link rel="manifest"  type="locale" dir = "ltr" lang="en-us" href="i18n/manifest.en-us.v1.json" />
```

christianliebel:
@calidion There are cases where you must specify a different directionality and language for a certain string within a manifest, so just doing it on the top level would not be enough. For example, you may always want to keep your _Change language…_ shortcut in English regardless of the manifest language.

calidion:
@christianliebel 
pwa manifest files are files providing descriptive information for operating systems and / or browsers, not a content rich files for end users.
I don't think such feature is needed while most w3c rfc files are english only as these files really sometimes needed by some end users.

dmurph:
@calidion 

Those are interesting ideas, but unfortunately that doesn't solve all of the use-cases and problems we need to solve with this feature. If you feel passionately about your ideas I encourage you to write an explainer outlining your use-cases and the solution that solves those use-cases well, including the use-cases we have outlined above. If you don't feel those use-cases are valid, then I'm afraid we will never reach consensus, and we'll have to agree to disagree :)

calidion:
@dmurph 
what use case need multiple language support for a manifest?

index.html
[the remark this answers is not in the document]
christianliebel: Resolution: Don’t allow image resource localizations.

index.html
christianliebel: Resolution: `<dfn>`*_localized` member</dfn>`

index.html
christianliebel: Resolution: Locale matching is left to the OS (e.g., needed for language change during runtime).

calidion:
@christianliebel  @dmurph 

I don't think the suffix `_localized` is a proper way for manifest localization and would like to suggestion use html tag.

## HTML Tag method for PWA manifest localization 

Steps: 
1. set a default/base manifest. for example use English based manifest.
2. keep previous manifest document unchanged.
3. add language extension by defining new localized manifest files.

pros: 
1. consistent manifest definition
2. clean and short attribute names
3. no need to add suffix
4. multiple language files for better versioning
5. can localize images, videos as well

cons:
 to be discovered.

## Localization for manifest

1. use tag <manifest> or  use <link>
2. support at lease `dir`, `lang` attributes and should be extensible

`manifest` tag:
```
 <!-- default manifest -->
<manifest href="i18n/manifest.json" />   
<manifest dir = "ltr" lang="fr" href="i18n/manifest.fr.json" />
<manifest dir = "rtl" lang="fr-ca" href="i18n/manifest.fr-ca.json" />
<manifest dir = "auto" lang="en" href="i18n/manifest.en.json" />
<manifest dir = "ltr" lang="en-us" href="i18n/manifest.en-us.json" />
<manifest dir = "ltr" lang="en-us" href="i18n/manifest.en-us.v1.json" />
```

`link` tag:
```
 <!-- default manifest -->
<link rel="manifest"  type="locale"  href="i18n/manifest.json" />
<link rel="manifest"  type="locale" dir = "ltr" lang="fr" href="i18n/manifest.fr.json" />
<link rel="manifest"  type="locale" dir = "rtl" lang="fr-ca" href="i18n/manifest.fr-ca.json" />
<link rel="manifest"  type="locale" dir = "auto" lang="en" href="i18n/manifest.en.json" />
<link rel="manifest"  type="locale" dir = "ltr" lang="en-us" href="i18n/manifest.en-us.json" />
<link rel="manifest"  type="locale" dir = "ltr" lang="en-us" href="i18n/manifest.en-us.v1.json" />
```
4. the default manifest would contain basic features required
5. the extended manifest can have extra features
6.  browsers/OSes parse the base manifest first then parse the extended one according to user's preferences.

christianliebel:
@calidion We’ve discussed the HTML method before (see https://github.com/w3c/manifest/issues/676#issuecomment-439747607). Unfortunately, this method does not support a critical use case: when the user changes the OS language setting, the OS must immediately apply the localized names, icons, etc. of the installed web application. Hence, the localized information must already be included in the manifest. Please also note that we make these decisions in close coordination with the i18n working group.

aphillips:
@calidion @christianliebel 

> Please also note that we make these decisions in close coordination with the i18n working group.

Note that I18N has pointed out that the choices here necessarily include tradeoffs.

Examples in the spec and generally in discussion of the spec have a small number of languages. However, many applications are localized into tens or (in the case of large or geographically diverse companies) even over a hundred languages. 

Using multiple files for manifest localization reduces application size for these cases, since only the languages in use have to be retrieved. However, the retrieval mechanism is more complex and runtime changes (particularly to the OS locale) become more complicated or require a network connection to effect.

Using a single manifest file resolves some of these these problems, but the manifest itself can become quite large. The file can be difficult to work with (who has the ability to edit Hindi _and_ Japanese _and_ Arabic at the same time?). It also means that adding a language means redistributing the entire manifest.

None of these choices are specifically right or wrong. See: https://www.w3.org/TR/localizable-manifests/

index.html
```diff
@@ -1543,6 +1764,13 @@ <h3>
             The user agent MAY automatically apply the changes if the update
             does not contain changes to [=security-sensitive members=].
           </p>
+          <p>
+            If a user changes localization settings, the user agent MAY
```
aphillips, line 1768: You might also want to mention that, regardless of the direction (`dir`) of the localized value, the security sensitive members should be displayed in a bidirectionally isolated way to prevent the kinds of display problems described in [Unicode Source Code Handling](https://www.unicode.org/reports/tr55) [UTR55].

index.html
```diff
@@ -3294,6 +3549,16 @@ <h2>
         using one of the following options:
       </p>
       <dl>
+        <dt>
+          Localized values in the manifest:
+        </dt>
+        <dd>
+          Authors can provide [=localized values=] for the [=localizable
+          members=] of the [=manifest=]. The user agent should pass all
```
aphillips, line 3557: should or SHOULD?

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+                "ar":    {"value": "منتقي الألوان", "dir": "rtl"}
+              }
+            }
+          </pre>
+        </aside>
+        <p>
+          [=manifest/*_localized=] members contain a [=language map=] that
+          defines [=localized values=] for the given [=localizable member=] in
+          the application. The user agent SHOULD use the user's localization
+          settings to select the [=localized value=] whose language tag best
```
aphillips, line 1286: Perhaps:

```suggestion
          the application. The user agent SHOULD use the user's localization
          settings ([=locale=]) to select the [=localized value=] whose language tag best
```

aphillips (approved):
Minor comments. Looks good! Thanks.

calidion:
@aphillips 

would it be possible for the web application installer to download all manifests for supported languages?
and keep os operation local without any connection to the network?

aphillips:
@calidion asked:

> would it be possible for the web application installer to download all manifests for supported languages?
> and keep os operation local without any connection to the network?

Sure. The cost, of course, is that now you have more files. If the manifest is packaged into the app (so that the app is a single download, e.g. like a zip file) this might not matter. If each file is separately retrieved, you'll have the usual problems associated with multiple files (if you have 100 languages, that's 100 requests on your server and 100 times that the client is waiting on the request)

calidion:
> @calidion asked:
> 
> > would it be possible for the web application installer to download all manifests for supported languages?
> > and keep os operation local without any connection to the network?
> 
> Sure. The cost, of course, is that now you have more files. If the manifest is packaged into the app (so that the app is a single download, e.g. like a zip file) this might not matter. If each file is separately retrieved, you'll have the usual problems associated with multiple files (if you have 100 languages, that's 100 requests on your server and 100 times that the client is waiting on the request)

I think pwa should be able to be zipped as a file and append an extension for example `index.pwa`
which containing all static files needed for the browser/os installer. The installer can download .pwa for acceleration.

calidion:
by zipping the web applications into .pwa files,  the pwas can be run both locally(desktop/mobile) and remotely(web), even without installing.

calidion:
And .pwa can also be a new universal executive file format for all platforms that support web standards .

index.html
christianliebel: This section is informative, so I assume we should avoid RFC keywords here.

index.html
christianliebel: I can't cite the i18n-glossary document here, as it is non-normative, but I will do it in the informative Internationalization section.

christianliebel:
@calidion The concept you mentioned was piloted under the name [Web Bundles](https://developer.chrome.com/docs/web-platform/web-bundles) and might re-emerge as [Isolated Web Apps](https://github.com/WICG/isolated-web-apps/tree/main). However, it's important that installable web apps continue to function without requiring any bundling, just as they do today.

calidion:
@christianliebel 

Thanks. 

As investigated web bundles and isolated web apps, I think my ideas may have some differences in that:

1. pwa should be packed by any zip tools without introducing new tools.
2. use the pwa concept  instead of introducing new concept wbn or iwa.
3. pwa is an application format can fully utilize the os abilities with the permission granted by the end users.
4. pwa has no security limitation as described in isolated web apps.

index.html
dmurph, line 1286:
```suggestion
          settings to select the [=localized value=] whose [=language tag=] key best
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              tag=] for a [=localized value=]. This is needed for situations
+              where a term or text must be presented in a language different
+              from the user's set language. For example, this helps ensure that
+              an application name is correctly pronounced by assistive
+              technology, even if it is in a foreign language.
+            </p>
+          </aside>
+          <p>
+            To <dfn>process a `*_localized` text member</dfn>, given [=ordered
+            map=] |json:ordered map|, [=ordered map=] |map:ordered map|,
```
dmurph, line 1346: optional:

```suggestion
            map=] |json:ordered map|, [=ordered map=] |outputMap:ordered map|,
```
(and change references below)

This helps readability for me, knowing this argument is populated as the output of this algo

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            <li>[=Set/For each=] |languageTag:string| of |languageTags|, run
+            [=process a localized text value=], passing
+            |languageMap|[|languageTag|] as the [=localized value=],
+            |languageTag|, |map|, |member|, and |manifest direction|.
+            </li>
+          </ol>
+          <p>
+            To <dfn>process a localized text value</dfn>, given [=string=] or
+            [=ordered map=] |localizedValue|, [=string=] |languageTag:string|,
+            [=ordered map=] |map:ordered map|, [=string=] |member:string|, and
```
dmurph, line 1371: optional:

```suggestion
            [=ordered map=] |outputMap:ordered map|, [=string=] |member:string|, and
```

(and change references below)

This helps readability for me, knowing this argument is populated as the output of this algo

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            </li>
+            <li>[=Set/For each=] |languageTag:string| of |languageTags|, run
+            [=process a localized text value=], passing
+            |languageMap|[|languageTag|] as the [=localized value=],
+            |languageTag|, |map|, |member|, and |manifest direction|.
+            </li>
+          </ol>
+          <p>
+            To <dfn>process a localized text value</dfn>, given [=string=] or
+            [=ordered map=] |localizedValue|, [=string=] |languageTag:string|,
```
dmurph, line 1370: optional, here and below:

```suggestion
            [=ordered map=] |localizedValue|, [=string=] |defaultLanguageTag:string|,
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            [=ordered map=] |map:ordered map|, [=string=] |member:string|, and
+            [=text-direction=] |manifest direction:string|:
```
dmurph, line 1372: optional, here and below:

```suggestion
            [=text-direction=] |defaultDirection:string|:
```

dmurph (commented):
This generally LGTM, a few minor nits.

index.html
christianliebel: I tried to be consistent with the other algos in the spec that use “map” currently. If we change this, should we change all uses of “map”?

index.html
christianliebel: This is not necessarily the manifest's default language tag but the language map's key. So I think “languageTag” is correct here.

index.html
christianliebel: I used “manifest direction“ for the other algos in this PR, do you think I should change it everywhere? Also, is “defaultDirection” or “default direction” correct? The algos in our spec do it both ways.

index.html
dmurph: Isn't this used if the value doesn't have a 'dir'? Hence it's the default one for parsing this value?

index.html
dmurph: I'm not sure - I don't know if it's so important to keep argument names consistent, especially for this patch. I think just update these ones.

index.html
dmurph: Whatever format you want here is fine. In this algorithm it is used if the value doesn't specify a direction, so that's why I called it the default.

I'm ok with "manifest direction" too.

dmurph (approved):
Changes LGTM - the final comments I left are optional changes.

marcoscaceres:
Filed WebKit tracking bug https://bugs.webkit.org/show_bug.cgi?id=276463 ...

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            For example, the [=manifest/name=] member is a [=localizable
+            member=], and `name_localized` is its corresponding
+            [=manifest/*_localized=] member.
+          </p>
+        </aside>
+        <p>
+          A <dfn>language map</dfn> is an [=ordered map=] whose key is a
+          [=language tag=] and whose value is a [=localized value=]. The
+          <dfn>localized value</dfn> represents the value applied for the
+          [=language tag=] specified as the key.
```
marcoscaceres, line 1264:
```suggestion
          <dfn>localized value</dfn> is content localized in the language given by the key.
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            }
+          </pre>
+        </aside>
+        <p>
+          [=manifest/*_localized=] members contain a [=language map=] that
+          defines [=localized values=] for the given [=localizable member=] in
+          the application. The user agent SHOULD use the user's localization
+          settings to select the [=localized value=] whose [=language tag=] key best
+          matches the user's preference. When no such [=localized value=] is
+          available, the default representation is used.
```
marcoscaceres, line 1288: "Default representation" should be defined (meaning the one at the root, if present).

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+        </p>
+        <section>
+          <h3>
+            Localizing text resources
+          </h3>
+          <p>
+            For [=manifest/*_localized=] members that accept a [=string=] as
+            their value, the [=localized value=] can either be a plain
+            localized [=string=] or an [=ordered map=] containing the following
+            members:
```
marcoscaceres, line 1298:
```suggestion
            their value, the [=localized value=] can either be 
            a [=string=] or an [=ordered map=] containing the following
            members:
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+          </p>
+          <dl>
```
marcoscaceres, line 1300: All this should move up to the where `<dfn>` language map is defined. The language map should have to have its own section (similar to how we define images).

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            </dd>
+            <dt>
+              `dir` (optional)
+            </dt>
+            <dd>
+              The [=text-direction=].
+            </dd>
+          </dl>
+          <p>
+            When a plain [=string=] is used, or when the `lang` or `dir`
```
marcoscaceres, line 1321:
```suggestion
            When a [=string=] is used, or when the `lang` or `dir`
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            members are omitted, `lang` falls back the [=language tag=] of the
```
marcoscaceres, line 1322:
```suggestion
            members are missing, `lang` falls back the [=language tag=] of the
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            [=language map=] key, and `dir` falls back to the [=manifest/dir=]
```
marcoscaceres, line 1323: I'm having trouble following this... I think you should make two paragraphs here, one for dir and one for lang. 

I was expecting the lang to fall back to to the lang defined at the root of the manifest.

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            member of the [=manifest=].
+          </p>
+          <aside class="note">
+            <p>
+              The `dir` member must be specified if the direction of a
```
marcoscaceres, line 1328: NB: avoid "must" etc. (i.e., RFC2119 keywords) in notes 🙏

marcoscaceres, line 1328:
```suggestion
              The `dir` member needs to be present if the direction of a
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              localized string differs from the default direction set in the
```
marcoscaceres, line 1329: We might need to formally define "default direction" and "default language"... or maybe call them "root language" and "root direction"? I don't mind which. 

 
```suggestion
              localized string differs from the [=default direction=] set in the
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              manifest ([=manifest/dir=] member). Right-to-left text will
+              require specific direction settings if the manifest's default
+              direction is left-to-right, and vice versa.
+            </p>
+            <p>
+              To support multilingual content and ensure proper display and
```
marcoscaceres, line 1335:
```suggestion
              To support multilingual content and ensure optimal display and
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              accessibility, it is possible to specify a different [=language
+              tag=] for a [=localized value=]. This is needed for situations
```
marcoscaceres, line 1337: Would be good to have a ". For example..." here.

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              where a term or text must be presented in a language different
```
marcoscaceres, line 1338:
```suggestion
              where a term or text needs to be presented in a language different
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+              from the user's set language. For example, this helps ensure that
+              an application name is correctly pronounced by assistive
+              technology, even if it is in a foreign language.
```
marcoscaceres, line 1341:
```suggestion
              from the user's set language. This helps ensure that
              an application name is correctly pronounced by assistive
              technology, even if it is in a language foreign to the user.
```

marcoscaceres: The intent of why this is great is good, but I think we need to rewrite this. Let's discuss in our next call. In particular, it would be great to rewrite this so "correctly pronounced" is at the start of the note. We should also provide an actual example.

index.html
marcoscaceres: oh, good point... I'm kinda regretting the way I ended specifying passing in the output map :( That's a very "gecko" thing to do, and I should really switch all these back to retiring a parser map... anyway, let's do "map" for now, and then we can fix all of them later.

index.html
marcoscaceres: yeah, or "root direction", but "default direction" and "default language" work.

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+            </li>
+            <li>[=Map/Set=] |map|[|member|][|languageTag|] to
+            |normalizedValue|.
+            </li>
+          </ol>
+          <p class="note">
+            The [=process a localized text value=] algorithm takes both a
+            [=string=] or an [=ordered map=] for the [=localized value=]
+            parameter, but the processed result will be normalized into an
+            [=ordered map=] with the `value`, `lang`, and `dir` members set.
```
marcoscaceres, line 1413: it should be possible to link them like this:

```suggestion
            [=ordered map=] with the [=manifest/*_localized/value=], [=manifest/*_localized/lang=], and [=manifest/*_localized/dir=] members set.
```

index.html
```diff
@@ -1255,6 +1239,231 @@ <h3>
...
+          </p>
+        </section>
+        <section>
+          <h3>
+            Localizing image resources
+          </h3>
+          <p>
+            For [=manifest/*_localized=] members that accept a [=list=] of
+            [=image resources=] as their value, the [=localized value=] must
+            also be a [=list=] of [=image resources=].
```
marcoscaceres, line 1423: Statement of fact: 

```suggestion
            For [=manifest/*_localized=] members that accept a [=list=] of
            [=image resources=] as their value, the [=localized value=] is
            a [=list=] of [=image resources=].
```

index.html
```diff
@@ -2298,7 +2556,9 @@ <h3>
         </h3>
         <p>
           The <dfn>application's name</dfn> is derived from either the
-          [=manifest/name=] member or [=manifest/short_name=] member.
+          [=manifest/name=] member or [=manifest/short_name=] member. The user
```
marcoscaceres, line 2559: This needs more clarification as to which one wins.

index.html
```diff
@@ -3294,6 +3555,16 @@ <h2>
         using one of the following options:
       </p>
       <dl>
+        <dt>
+          Localized values in the manifest:
+        </dt>
+        <dd>
+          Authors can provide [=localized values=] for the [=localizable
+          members=] of the [=manifest=]. The user agent should pass all
```
marcoscaceres, line 3563:
```suggestion
          members=] of the [=manifest=]. The user agent pass all
```

index.html
```diff
@@ -3294,6 +3555,16 @@ <h2>
...
+          localized values to the host operating system. When the user changes
+          the [=locale=] at the OS level, the OS can immediately display the
```
marcoscaceres, line 3565:
```suggestion
          the [=locale=] at the OS level, the OS can display the
```

index.html
```diff
@@ -549,13 +534,7 @@ <h3>
           <li>If |json|["lang"] doesn't [=map/exist=] or if |json|["lang"] is
           not a [=string=], return.
           </li>
-          <li>If calling <a data-cite=
```
marcoscaceres, line 552: Don't we still need this? We still want structurally valid tags, no?

christianliebel: I discussed this with @aphillips (see https://github.com/w3c/manifest/pull/1101#discussion_r1631528734). As parsing language tags is quite complicated, our proposed solution is to allow all strings and let the OS determine the appropriate language based on the user's locale. If an invalid language tag is specified, it will simply be ignored and not used.

index.html
jcayzac: I think this means that, for instance, in `"fr": {"value": "Sélecteur de Couleur", "dir": "ltr"}` the `lang` value is `"fr"`. But I agree it could be made clearer.

I'm not sure I understand why we need both the key of the map and this `lang` value. `"fr": {"value": "Sélecteur de Couleur", "lang": "fr-CA", "dir": "ltr"}` could just be `"fr-CA": {"value": "Sélecteur de Couleur", "dir": "ltr"}` no?

I would see some value in allowing multiple values, on the other hand. This would allow having a single entry for multiple variants of a language, e.g. `"fr-others": {"value": "Sélecteur de Couleur", "lang": "fr-CA, fr-BE", "dir": "ltr"}`

jcayzac: To make it even less ambiguous, I would just get rid of the language map and replace it with an array of objects:

```json
"name_localized": [
  { "lang": "de", "value": "Farbwähler" },
  { "lang": "en", "value": "Color Picker" },
  { "lang": "en-GB", "value": "Colour Picker" },
  { "lang": "fr", "value": "Sélecteur de Couleur" },
  { "lang": "fr-CA, fr-BE", "value": "Sélecteur des Couleurs" },
  { "lang": "ar", "value": "منتقي الألوان", "dir": "rtl" }
}
```

calidion:
I don't think put all languages into one file is a good idea.
And I never see compatibility work in web.
Almost all new created web pages can not be loaded in old browsers.
backward compatibility is meaningless in most cases.

calidion:
appendix `_localized` definitely is not a good choice for localization for a web application or a web site.
There should be a general localization schema for both a web application and/or a web site that can easily introduce multi-language support.
and reduce the current burden carried by most backend web servers and frontend libraries/frameworks.

index.html
christianliebel: @jcayzac @marcoscaceres This is needed in (probably rare) cases where you want the value in language A applied to locale B, for example, so that assistive technology can use the right voice for pronunciation.

I think a good example is if a brand is localized differently across countries ([actual example for such a brand](https://en.wikipedia.org/wiki/Just_Eat_Takeaway.com#Operations)):

```json
{
  "name": "Thuisbezorgd.nl",
  "lang": "nl-NL",
  "name_localized": {
    "de-DE": "Lieferando",
    "de-CH": { "lang": "en", "value": "Just eat" },
    "ro": { "lang": "en", "value": "Takeaway.com" }
  }
}
```

Or if you have a bilingual app and want to offer a “change language” shortcut in the target language:

```json
"name_localized": {
  "en": { "lang": "fr", "value": "Changer de langue" },
  "fr": { "lang": "en", "value": "Change language" }
}
```

This is why
1. we need a language map with a language tag key
1. the "lang" attribute, if not specified, must fall back to the language tag specified in the key

index.html
christianliebel: @dmurph This value is the language map's key. It is used if the localized value doesn't have a "lang" member (maybe related: https://github.com/w3c/manifest/pull/1101#discussion_r1676783850). I think the wording "default language tag" could be confusing if we also use "default direction" as the "default language tag" is the language map key (not the root/manifest language), but the "default direction" would be the root/manifest direction.

christianliebel:
> There should be a general localization schema for both a web application and/or a web site that can easily introduce multi-language support.

@calidion While a general localization schema for web applications and websites may be beneficial, this PR is solely focused on adding localization capabilities specifically to the Web Application Manifest rather than solving broader multi-language support for web applications or websites. I think the [WICG Proposals](https://github.com/WICG/proposals) repo would be the right spot to propose and discuss a more general solution.

index.html
christianliebel: The position is correct in my opinion. I will update the section to make it clearer that this is not the language map.

index.html
christianliebel: A `*_localized` value always takes a language map. The question is what the _value_ of the language map is. In this case, it can be either a string or a "localized text object" (with the `dir`, `lang`, and `value` members).

index.html
christianliebel: I added an example and changed the text a bit.

calidion:
@christianliebel 
Thanks for the information.

But still I hope this feature can be hold for a while before vast agreement would be made.

index.html
aphillips: Note that the comment is about _matching_ language tags. _Parsing_ language tags is more straightforward.

It's a good idea to check if the tag is what BCP47 calls "well formed". (BCP47 defined two levels of tag conformance. A tag can be grammatically correct ['well formed'] and it can contain only registered subtags with other restrictions ['valid']. Unfortunately, 402 named the "well-formed-checking" thing `IsStructurallyValidLanguageTag`). In general, few specifications require validity checking of the tags, because that requires access to (a copy of) the IANA Language Subtag Registry, but well-formed checking helps users avoid putting invalid strings in.

The canonicalization of language tags is into locale IDs is probably fine. So perhaps put this text back?

marcoscaceres: yeah, let's put this back. We want the make sure that only structurally valid language tags ever get to the OS. The manifest parser should be filtering out any invalid ones.

christianliebel: It's back. I'm leaving this conversation open as a reminder that I need to do the same for the language tag keys in the language map.

index.html
dmurph: Same idea as my other comment - from the perspective this algorithm, this argument is the 'default value' if one isn't specified. totally optional suggestion, you can keep it the same too.

index.html
```diff
@@ -1255,6 +1246,262 @@ <h3>
...
+        </aside>
+        <p>
+          A <dfn>language map</dfn> is an [=ordered map=] whose key is a
+          [=language tag=] and whose value is a [=localized value=]. The
+          <dfn>localized value</dfn> is content localized in the language given
+          by the key.
+        </p>
+        <aside class="example" title="Localizing the application name">
+          <pre class="json">
+            {
```
dmurph, line 1275: In order to make it obvious why one might want to have a different 'lang' for an entry in here, perhaps we should use a company that so it would have opinions about name pronunciations in other languages.

L'Occitane is a good example - using wikipedia you can see that sometimes is uses the french version, while other times it has a translated name. For cases like English & German, screen readers should read the name in a french pronunciation. 

```suggestion
            {
              "lang": "fr",
              "dir": "ltr",
              "name":  "L'Occitane",
              "name_localized": {
                "en":  { value: "L'Occitane", "lang": "fr" },
                "de":  { value:  "L'Occitane", "lang": "fr" },
                "zh":  "歐舒丹"
                "en-GB": {"value": "L'Occitane en Provence", "lang": "fr" },
                "fr":    "L'Occitane",
                "ar":    {value: "لوکسیتان",  "dir": "rtl"}
              }
            }
```

index.html
dmurph: In Chromium we use them depending on UX needs. It's not implemented, but I would like to follow similar rules for extensions - `short_name` is truncated to 12 characters and `name` is truncated to 75.

This sets expectations for devs, and allows the user agent to show an app name where there might not be much space.

dmurph (commented):
Still LGTM, one suggestion to change the example.

index.html
christianliebel: "Default representation" is already defined, and it refers to the member without the `*_localized` suffix.

It's a good question what should happen in this scenario:

```json
{
  // "name" (default representation) is missing
  "name_localized": {
    "de": "Guten Morgen",
    "fr": "Bonjour"
  }
}
```

We could simply ignore the `*_localized` member if there is no/an invalid default representation (case A).

Or, we could specify a fallback chain:
1. Match the `*_localized` against the user's locale settings and use the best match (case B).
2. If we have a valid `lang` specified at the root level, match the `*_localized` language map against it and use the best match (case C).
3. If there is still no match, ignore the member altogether (case D).

WDYT? @marcoscaceres @dmurph @aphillips 

<details>
<summary>Cases</summary>

## Case A

```json
{
  "name": 123, // invalid name
  "name_localized": {
    "de": "Guten Morgen",
    "fr": "Bonjour"
  }
}
```

| Locale | Result |
| --- | --- |
| `de` | — |
| `fr` | — |
| `en` | — |

## Case B

```json
{
  "name_localized": {
    "de": "Guten Morgen",
    "fr": "Bonjour"
  }
}
```

| Locale | Result |
| --- | --- |
| `de` | Guten Morgen |
| `fr` | Bonjour |
| `en` | — |

## Case C

```json
{
  "lang": "fr",
  "name_localized": {
    "de": "Guten Morgen",
    "fr": "Bonjour"
  }
}
```

| Locale | Result |
| --- | --- |
| `de` | Guten Morgen |
| `fr` | Bonjour |
| `en` | Bonjour |

## Case D

```json
{
  "lang": "en",
  "name_localized": {
    "de": "Guten Morgen",
    "fr": "Bonjour"
  }
}
```

| Locale | Result |
| --- | --- |
| `de` | Guten Morgen |
| `fr` | Bonjour |
| `en` | — |
</details>

index.html
christianliebel: @marcoscaceres I split this into [two paragraphs](https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1101.html#localizing-text-resources:~:text=When%20a%20string%20is%20used%2C%20or%20when%20the%20dir%20member%20of%20the%20localized%20text%20resource%20is%20missing%2C%20the%20default%20direction%20(dir%20member%20of%20the%20manifest)%20is%20applied.). Does that help?

index.html
christianliebel: @marcoscaceres [Here's](http://localhost:3002/#x_localized-members:~:text=key%20is%20applied.-,NOTE,be%20presented%20in%20a%20language%20different%20from%20the%20user%27s%20set%20locale.,-EXAMPLE%209) the new version.

index.html
christianliebel: Could we discuss that in a separate issue? It used to be like this and is not directly related to l10n.

index.html
christianliebel: If it's okay to use real-world brands/examples, that's fine with me. I also have the "Just Eat" example here, where the brand even differs between de-DE and de-CH: https://github.com/w3c/manifest/pull/1101#discussion_r1676783850

@marcoscaceres WDYT?

marcoscaceres: Sounds good, but don't use a real company name... they might not approve.

marcoscaceres: The example below seems fine to me... @dmurph?

index.html
```diff
@@ -1255,6 +1246,276 @@ <h3>
...
+          whose [=language tag=] key best matches the user's preference. When
+          no such [=localized value=] is available, the [=default
+          representation=] is used.
+        </p>
+        <section>
+          <h3>
+            Localizing text resources
+          </h3>
+          <p>
+            A <dfn>localized text resource</dfn> is an [=ordered map=] with the
```
marcoscaceres, line 1304: Let's call this "localized text object" or check what image objects are called

index.html
```diff
@@ -1255,6 +1246,276 @@ <h3>
...
+                "en":    {"value": "Color Picker"},
+                "en-GB": {"value": "Colour Picker", "dir": "ltr"},
+                "fr":    {"value": "Sélecteur de Couleur", "lang": "fr-CA", "dir": "ltr"},
+                "ar":    {"value": "منتقي الألوان", "dir": "rtl"}
+              }
+            }
+          </pre>
+        </aside>
+        <p>
+          The value assigned to the [=localizable member=] is the <dfn>default
```
marcoscaceres, line 1290:
```suggestion
          The value assigned to the [=localizable member=] is the <dfn data-for="localizable member">default
```

index.html
```diff
@@ -2063,6 +2349,9 @@ <h3>
           "shortcut item">icons</dfn></code> member lists images that serve as
           iconic representations of the shortcut in various contexts.
         </p>
+        <p>
+          This member is a [=localizable member=].
```
marcoscaceres, line 2353: Like this:

```suggestion
          This [=manifest/icons=] member is a [=localizable member=].
```

index.html
marcoscaceres: How they are treated by the OS is implementation specific. I don't think we need to say anything here.

marcoscaceres (approved):
This looks ok to me. 

Add the trimming any text strings.

index.html
dmurph: IDK if it's true that these are only used by the OS - the user agent uses these strings too. Happy for it to be a non-normative note - truncating characters is actually fraught with weirdness due to multi-character sequences now etc, so it would be helpful to document here limits to prevent those breakages.

User agents can't just allow any string here to be displayed in full, as it'll cause weird UX issues with dialogs / menu items / etc. This was the purpose of, say, `short_name`.

But - I certainly don't want to block this. And perhaps you are right here @christianliebel that this can be addressed in a separate issue / pull request.

aphillips: I'll second @dmurph's _truncating characters is actually fraught with weirdness_. 

Example. 🏴󠁧󠁢󠁳󠁣󠁴󠁿

The Scottish flag emoji is 7 code points. The JavaScript string length is 14. It's one grapheme (user perceived character, i.e. screen position). If the `short_name` limit is `short_name.length < 13`, the flag won't fit...

As a reminder, I18N's [guidelines about text truncation](https://www.w3.org/TR/international-specs/#char_truncation) apply to length limits. One has to be careful about what one is counting (bytes, code points, graphemes) to ensure that some languages are not disadvantaged because of how their writing system works (emoji makes for a great demo, but there are languages that use combining marks similar to the way emoji sequences work that have the same counting issue)

aphillips:
Why did you add trimming of localized string values? Leading or trailing whitespace is sometimes significant in localization. Often it is a signal of an I18N bug, but still... users sometimes count on the whitespace being present.

christianliebel:
@aphillips The browser engines trim the values for the default representation fields (example `name`: [Chromium](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/modules/manifest/manifest_parser.cc#L620), [Gecko](https://github.com/mozilla/gecko-dev/blob/master/dom/manifest/ManifestProcessor.sys.mjs#L146), [WebKit](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/applicationmanifest/ApplicationManifestParser.cpp#L541)), so we wanted to treat the localized values the same way. If you are still at TPAC today, we can also discuss this in person. The WebApps WG meets in Huntington (4 Concourse Level) today.

aphillips:
No, that answers the question. I'm "at" TPAC, but have a cold, so I'm staying away from giving it to you all. Happy to jump on your Zoom connection if there is more to discuss, though.

index.html
christianliebel: Created https://github.com/w3c/manifest/issues/1145 for further discussion.

mgiuca:
Amazing work, Christian! This has been an extremely long standing and complex issue, it's great that you worked your way through it.

aarongustafson:
Yes, congrats. Amazing work!

dmurph:
This is really awesome @christianliebel! Thanks for your work here and being so persistent!

jcayzac:
Thanks @christianliebel!

PR #1112, opened 2024-03-07
chore: Export the terms "manifest URL" and "start URL"

mgiuca:
This change (choose at least one, delete ones that don't apply):

* Is a "chore" (metadata, formatting, fixing warnings, etc).

Commit message:

These are generally useful concepts relevant to the manifest spec but
not internal details. I am exporting them because they need to be linked
from the manifest-incubations spec.

Also ran tidy.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/1112.html" title="Last updated on Apr 10, 2024, 5:28 AM UTC (fed0454)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1112/9d8ad08...mgiuca:fed0454.html" title="Last updated on Apr 10, 2024, 5:28 AM UTC (fed0454)">Diff</a>

PR #1114, opened 2024-03-07
Rewrite privacy considerations on fingerprinting in start_url

### LINKED ISSUE #1113: Privacy consideration specifies an unenforceable "MUST NOT" condition on developers
Under "Privacy consideration: start_url tracking", there is the following normative requirement:

> Additionally, developers _MUST NOT_ use the start URL to include information that uniquely identifies a user (e.g., "?user=123" or "/user/123/", or "https://user123.foo.bar").

Although it has good intentions, this is unenforceable. Specs do not give requirements to site developers, they give requirements to user agents, and information to developers. If you identify potentially bad developer behaviour, you can't simply say "MUST NOT" to the developer, you have to identify the behaviour and give recommendations to the user agents on how to mitigate it.

In this case, I would recommend:

1. Removing this requirement for developers.
2. Adding a non-normative note that tells developers it would be irresponsible to do this (but acknowledging that we can't practically prevent it).
3. A practical mitigation: Adding a MAY requirement for user agents to offer to uninstall apps associated with an origin when clearing site data. (I don't know of a browser that does this, but it's a reasonable UI and I think it's something that we've discussed before.)

mgiuca:
Closes #1113

(No implementation commitment required as it adds a MAY requirement.)

Commit message:

    Rewrite privacy considerations on fingerprinting in start_url.
    
    There is a "MUST NOT" requirement for developers about putting user data
    in the start_url. This is not enforceable, so rewriting the paragraph:
    
    1. Removed this requirement for developers.
    2. Added a non-normative note that tells developers it would be
       irresponsible to do this (but acknowledging that we can't practically
       prevent it).
    3. Added a MAY requirement for user agents to offer to uninstall apps
       associated with an origin when clearing site data.

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/1114.html" title="Last updated on May 1, 2024, 6:22 AM UTC (3351bdc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1114/a791201...mgiuca:3351bdc.html" title="Last updated on May 1, 2024, 6:22 AM UTC (3351bdc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -817,7 +817,7 @@
             This can be useful for analytics and possibly other customizations.
             However, it is also conceivable that developers could encode
             strings into the start_url that uniquely identify the user (e.g., a
-            server-assigned <abbr>UUID</abbr> such as `"?user=123"`,
+            server-assigned identifier, such as `"?user=123"`,
             `"/user/123/"`, or `"https://user123.foo.bar"`). This is
             fingerprinting/privacy sensitive information that the user might
             not be aware of.
--- a/index.html
+++ b/index.html
@@ -822,7 +822,7 @@
             fingerprinting/privacy sensitive information that the user might
             not be aware of.
           </p>
-          <p class="note">
+          <p class="note" title="Don't add identifiers to start URLs">
             It is bad practice for a developer to use the [=start URL=]
             to include information that uniquely identifies a user, as it would
             represent a fingerprint that is not cleared when the user clears
--- a/index.html
+++ b/index.html
@@ -823,7 +823,7 @@
             not be aware of.
           </p>
           <p class="note">
-            It would be irresponsible for a developer to use the [=start URL=]
+            It is bad practice for a developer to use the [=start URL=]
             to include information that uniquely identifies a user, as it would
             represent a fingerprint that is not cleared when the user clears
             site data. However, nothing in this specification can practically
--- a/index.html
+++ b/index.html
@@ -817,8 +817,17 @@
             This can be useful for analytics and possibly other customizations.
             However, it is also conceivable that developers could encode
             strings into the start_url that uniquely identify the user (e.g., a
-            server assigned <abbr>UUID</abbr>). This is fingerprinting/privacy
-            sensitive information that the user might not be aware of.
+            server-assigned <abbr>UUID</abbr> such as `"?user=123"`,
+            `"/user/123/"`, or `"https://user123.foo.bar"`). This is
+            fingerprinting/privacy sensitive information that the user might
+            not be aware of.
+          </p>
+          <p class="note">
+            It would be irresponsible for a developer to use the [=start URL=]
+            to include information that uniquely identifies a user, as it would
+            represent a fingerprint that is not cleared when the user clears
+            site data. However, nothing in this specification can practically
+            prevent developers from doing this.
           </p>
           <p>
             Given the above, it is RECOMMENDED that, upon installation, or any
@@ -826,9 +835,11 @@
             necessary, modify the [=start URL=] of an application.
           </p>
           <p>
-            Additionally, developers MUST NOT use the [=manifest/start URL=] to include
-            information that uniquely identifies a user (e.g., "?user=123" or
-            "/user/123/", or "https://user123.foo.bar").
+            A user agent MAY offer other protections against this form of
+            fingerprinting. For example, if a user clears data from an origin,
+            the user agent MAY offer to uninstall applications that are
+            [=manifest/within scope=] of that origin, thus removing the
+            potential fingerprint from the application's start URL.
           </p>
         </section>
       </section>


mgiuca:
@marcoscaceres the change is a bit hard to read because it's based off of #1112 which runs tidy and other changes. The relevant changed section is around line 810.

index.html
```diff
@@ -816,18 +817,29 @@ <h3>
             This can be useful for analytics and possibly other customizations.
             However, it is also conceivable that developers could encode
             strings into the start_url that uniquely identify the user (e.g., a
-            server assigned <abbr>UUID</abbr>). This is fingerprinting/privacy
-            sensitive information that the user might not be aware of.
+            server-assigned <abbr>UUID</abbr> such as `"?user=123"`,
```
marcoscaceres, line 820:
```suggestion
            server-assigned identifier, such as `"?user=123"`,
```

index.html
```diff
@@ -816,18 +817,29 @@ <h3>
...
+            `"/user/123/"`, or `"https://user123.foo.bar"`). This is
+            fingerprinting/privacy sensitive information that the user might
+            not be aware of.
+          </p>
+          <p class="note">
+            It would be irresponsible for a developer to use the [=start URL=]
```
marcoscaceres, line 826:
```suggestion
            It is irresponsible for a developer to use the [=start URL=]
```

index.html
```diff
@@ -816,18 +817,29 @@ <h3>
...
             However, it is also conceivable that developers could encode
             strings into the start_url that uniquely identify the user (e.g., a
-            server assigned <abbr>UUID</abbr>). This is fingerprinting/privacy
-            sensitive information that the user might not be aware of.
+            server-assigned <abbr>UUID</abbr> such as `"?user=123"`,
+            `"/user/123/"`, or `"https://user123.foo.bar"`). This is
+            fingerprinting/privacy sensitive information that the user might
+            not be aware of.
+          </p>
+          <p class="note">
```
marcoscaceres, line 825:
```suggestion
          <p class="note" title="Don't add identifiers to start URLs">
```

index.html
```diff
@@ -816,18 +817,29 @@ <h3>
...
+            It would be irresponsible for a developer to use the [=start URL=]
+            to include information that uniquely identifies a user, as it would
+            represent a fingerprint that is not cleared when the user clears
+            site data. However, nothing in this specification can practically
+            prevent developers from doing this.
```
marcoscaceres, line 830:
```suggestion
            prevent developers from doing this. Similarly, nothing in this specification prevents user agents from identifying 
            and stripping out unique identifiers start URLs.
```

index.html
```diff
@@ -759,10 +759,11 @@ <h3>
         <p>
           The [=manifest's=] <code><dfn data-export="" data-dfn-for=
           "manifest">start_url</dfn></code> member is a <a>string</a> that
-          represents the <dfn>start URL</dfn> , which is <a>URL</a> that the
-          developer would prefer the user agent load when the user launches the
-          web application (e.g., when the user clicks on the icon of the web
-          application from a device's application menu or homescreen).
+          represents the <dfn data-export="">start URL</dfn> , which is
```
marcoscaceres, line 762: I'm still a bit worried about exporting "start URL" without an associated thing... could it be:

```suggestion
          represents the <dfn class"export" data-dfn-for="Web Application">start URL</dfn> , which is
```

marcoscaceres: Actually, seems we have defined "installed web application" 🎉

index.html
marcoscaceres: Maybe we should just make the above some may... that the user agent MAY strip out any personally identifiable information or IDs from start URLs.

marcoscaceres (commented):
Sorry for the delay... some suggestions.

index.html
mgiuca: Hmm.... I can't exactly point to the bit that says that you can't do this, but it would be weird I think if user agents do this. (i.e. it would break sites - you can't just randomly delete any single character from a URL because you don't know if it's important.)

The top of this spec says "The start_url member is purely advisory" so I guess UAs can do whatever they want. But I think there is an expectation that UAs actually load up the app at the URL developers specify and they shouldn't be changing it. (I would perhaps suggest we remove the statement "The start_url member is purely advisory".)

So while this may technically be true, I wouldn't say it like this as if it's a good idea for UAs to do.

index.html
mgiuca: This is commenting on a PR that was already submitted (#1112). (It just got caught up in this PR's delta because it wasn't submitted at the time I uploaded this.) I've rebased now, which should make things clearer.

You may be right, but I would like you to put this up as a separate PR if you don't mind. It's a bit complex as I've now got text in [manifest-incubations](https://wicg.github.io/manifest-incubations/#home_tab-member) which links to this and will have to be updated when you make this change.

mgiuca:
Thanks for the review. I've rebased so it's now current against HEAD.

I accepted most of your suggestions, just pushing back against the new sentence about UAs messing with start URLs.

marcoscaceres (approved):

marcoscaceres:
Thanks @mgiuca. I agree that until such times that user agents strip things out (if ever) then there is probably no need to mention it. Just as an example, I think mail.app does strip out known identifiers... so there is some precedent.

PR #1118, opened 2024-05-02
fix: dir's auto definition

### LINKED ISSUE #1079: Value `auto` for `dir` probably shouldn't say "no explicit directionality"
`dir` member
https://w3c.github.io/manifest/#dir-member

> "auto" (default)
>    No explicit directionality. 

The value `auto` isn't exactly the same as "no explicit directionality". It means that the direction of strings with this value should be estimated using the first-strong part of the bidi algorithm. It is an explicit value. It also implies that the value should be bidi isolated when displayed. Perhaps say "Direction determined from content" instead?

Note that content which fools the "first-strong" detection in `auto` is common: an actual direction value such as `rtl` or `ltr` is preferred to just relying on `auto`.

Note that there was some previous discussion of this topic in https://github.com/w3c/manifest/pull/926

marcoscaceres:
Closes #1079

* Breaks existing normative behavior (please add label "breaking")
* Is a "chore" (metadata, formatting, fixing warnings, etc).

Commit message:

difference will be.)

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1118.html" title="Last updated on Jun 6, 2024, 11:13 PM UTC (c9f3254)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1118/2f54efe...c9f3254.html" title="Last updated on Jun 6, 2024, 11:13 PM UTC (c9f3254)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -479,8 +479,9 @@
             </p>
             <aside class="note">
               <p>
-                An actual direction value such as `rtl` or `ltr` is preferred
-                to relying on `auto`.
+                An actual direction value such as "[=text-direction/rtl=]" or
+                "[=text-direction/ltr=]" is preferred to relying on
+                "[=text-direction/auto=]".
               </p>
             </aside>
           </dd>
--- a/index.html
+++ b/index.html
@@ -473,7 +473,12 @@
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            No explicit directionality.
+            <p>Direction determined from content.</p>
+            <aside class="note">
+              <p>
+                An actual direction value such as `rtl` or `ltr` is preferred to relying on `auto`.
+              </p>
+            </aside>
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -488,7 +488,7 @@
           </dt>
           <dd>
             <p>
-              Direction determined from content using the first-strong part of
+              Direction determined from content using <a data-cite="bidi#P2">Rule P2</a> of
               the [[BIDI]] algorithm.
             </p>
             <aside class="note">
--- a/index.html
+++ b/index.html
@@ -473,7 +473,7 @@
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            <p>Direction determined from content.</p>
+            <p>Direction determined from content using the first-strong part of the [[BIDI]] algorithm.</p>
             <aside class="note">
               <p>
                 An actual direction value such as `rtl` or `ltr` is preferred to relying on `auto`.
--- a/index.html
+++ b/index.html
@@ -493,7 +493,7 @@
             </p>
             <aside class="note">
               <p>
-                An explicit direction value such as "[=text-direction/rtl=]" or
+[=text-direction/auto=] means that the directionality of each member is determined by its first strongly directional character (as per the <a data-cite="bidi#P2">Rule P2</a>). An explicit direction value such as "[=text-direction/rtl=]" or
                 "[=text-direction/ltr=]" is preferred to relying on the default of
                 "[=text-direction/auto=]".
               </p>
--- a/index.html
+++ b/index.html
@@ -473,10 +473,14 @@
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            <p>Direction determined from content using the first-strong part of the [[BIDI]] algorithm.</p>
+            <p>
+              Direction determined from content using the first-strong part of
+              the [[BIDI]] algorithm.
+            </p>
             <aside class="note">
               <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -473,7 +473,12 @@ <h3>
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            No explicit directionality.
+            <p>Direction determined from content.</p>
```
marcoscaceres, line 476:
```suggestion
            <p>Direction determined from content using the first-strong part of the [[BIDI]] algorithm.</p>
```

mgiuca:
I can review this when it's ready (are you still working on it)?

marcoscaceres:
Thanks @mgiuca... then we should get @aphillips's eye's on it for approval.

aphillips (approved):
Looks good to me! Thanks.

index.html
```diff
@@ -473,7 +473,17 @@ <h3>
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            No explicit directionality.
+            <p>
+              Direction determined from content using the first-strong part of
```
mgiuca, line 477: Replace "the first strong part" with "Rule P2". (Elsewhere, we link to those rules like this: `<a data-cite="bidi#P2">Rule P2</a>` -- not sure if there is a more Respeccy way to link that.)

Note: Even though it doesn't explicitly say this, I consider this to be non-normative text (since it is just giving a human-readable summary of what this value means). The "when displaying" algorithm below gives the actual normative statement. So it isn't necessary to have this be entirely precise, but I think it helps to link to the exact part of the algorithm.

index.html
```diff
@@ -473,7 +473,17 @@ <h3>
...
+              the [[BIDI]] algorithm.
+            </p>
+            <aside class="note">
+              <p>
+                An explicit direction value such as "[=text-direction/rtl=]" or
```
mgiuca, line 482: Since my above comment makes the meaning of auto less clear, add to the front here: "Using `[=text-direction/auto=]` means that the directionality of each member will be determined by its first strongly directional character. An explicit  ....."

mgiuca (changes requested):
I think we can fix up some other text in this section while we're here.

1. "implying that the value of the localizable members is by default:" -- I don't know how to parse this part of the sentence (it's right before the list of options). Delete it.

2. "the use agent SHOULD" - s/use/user/

3. Replace "When displaying the localizable members to an end-user..." with:

"The user agent SHOULD display the localizable members to the end-user using the [BIDI] algorithm. If the base direction is ltr or rtl, the user agent SHOULD override Rule P3 of [BIDI] ..."

(Remove the ordered list and drop list item 2 entirely - there is no need to explicitly state what happens in the "auto" case since it's covered by Bidi automatically. And we already have non-normative text to explain what happens in this case, above.)

(Also, it shouldn't say "apply Rule P1"; rule P2 is where the first strong character is chosen. But I would advise deleting this sentence entirely.)

index.html
```diff
@@ -487,7 +487,17 @@ <h3>
             "<dfn data-dfn-for="text-direction">auto</dfn>" (default)
           </dt>
           <dd>
-            No explicit directionality.
+            <p>
+              Direction determined from content using the first-strong part of
```
marcoscaceres, line 491: How about we do: 

```suggestion
              Direction determined from content using <a data-cite="bidi#P2">Rule P2</a> of
```

That has the

index.html
```diff
@@ -487,7 +487,17 @@ <h3>
...
           </dt>
           <dd>
-            No explicit directionality.
+            <p>
+              Direction determined from content using <a data-cite="bidi#P2">Rule P2</a> of
+              the [[BIDI]] algorithm.
+            </p>
+            <aside class="note">
+              <p>
+                An explicit direction value such as "[=text-direction/rtl=]" or
```
marcoscaceres, line 496:
```suggestion
Using [=text-direction/auto=] means that the directionality of each member will be determined by its first strongly directional character (as per the <a data-cite="bidi#P2">Rule P2</a>). An explicit direction value such as "[=text-direction/rtl=]" or
```

index.html
```diff
@@ -487,7 +487,17 @@ <h3>
...
+                "[=text-direction/ltr=]" is preferred to relying on
+                "[=text-direction/auto=]".
```
marcoscaceres, line 498:
```suggestion
                "[=text-direction/ltr=]" is preferred to relying on the default of
                "[=text-direction/auto=]".
```

dmurph (approved):

marcoscaceres:
@mgiuca, revised based on your feedback. Taking @dmurph review as the basis for merging. 

We can do any additional fixes as a follow up.

PR #1120, opened 2024-05-02
Remove problematic icon matching algorithm

### LINKED ISSUE #1094: Icon purpose "any" is confusing
From discussions on other bugs and elsewhere I learned that the "any" purpose does not in fact mean that "The user agent is free to display the icon in any context.", but rather that the icon can be used in any context for which no other purpose has been defined (i.e. icons without explicit purpose "maskable" should not be masked). 

While the ship has sailed as far as naming goes, I do think it would be good to be very explicit about this in the explanation of what the "any" purpose actually means and how purpose of an icon ties into icon selection logic.

If for example the "most appropriate" icon was actually defined in an algorithm (related  discussion in https://github.com/w3c/manifest/issues/925#issuecomment-1642799399) that would probably have avoided confusion here, as such an algorithm would presumably have included the purpose of icons, but if the spec doesn't want to define that, at least some non-normative text would be good I think.

marcoscaceres:
Closes #1094 

* Breaks existing normative behavior (please add label "breaking")
* Is a "chore" (metadata, formatting, fixing warnings, etc).

Commit message:

difference will be.)

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1120.html" title="Last updated on Jun 6, 2024, 11:01 PM UTC (b877a03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1120/bceb2ee...b877a03.html" title="Last updated on Jun 6, 2024, 11:01 PM UTC (b877a03)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1601,8 +1601,10 @@
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon where no [=manifest image resource/purpose=] is required.
-            For example, an "any" [=manifest image resource=] wouldn't be used in a context where "[=icon purpose/monochrome=]" is required.
+            The user agent is free to display the icon where no [=manifest
+            image resource/purpose=] is required. For example, an "any"
+            [=manifest image resource=] wouldn't be used in a context where
+            "[=icon purpose/monochrome=]" is required.
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -1602,6 +1602,7 @@
           </dt>
           <dd>
             The user agent is free to display the icon where no [=manifest image resource/purpose=] is required.
+            For example, an "any" [=manifest image resource=] wouldn't be used in a context where "[=icon purpose/monochrome=]" is required.
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -1602,7 +1602,7 @@
           </dt>
           <dd>
             The user agent is free to display the icon where a more specialized
-            purpose wouldn't be suitable.
+            purpose isn't suitable.
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -1601,8 +1601,7 @@
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon where a more specialized
-            purpose isn't suitable.
+            The user agent is free to display the icon where no specific purpose is required.
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -1601,7 +1601,8 @@
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where a more specialized
+            purpose wouldn't be suitable.
           </dd>
         </dl>
         <p>
--- a/index.html
+++ b/index.html
@@ -649,16 +649,6 @@
           an <abbr title="Operating system">OS</abbr>'s task switcher and/or
           system preferences.
         </p>
-        <p>
-          If there are multiple equally appropriate images in
-          [=manifest/icons=], a user agent MUST use the last one declared in
-          order at the time that the user agent collected the list of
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1611,7 +1601,8 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where a more specialized
+            purpose wouldn't be suitable.
```
marcoscaceres, line 1605:
```suggestion
            purpose isn't suitable.
```

index.html
```diff
@@ -1611,7 +1601,8 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where a more specialized
+            purpose isn't suitable.
```
marcoscaceres, line 1605:
```suggestion
            The user agent is free to display the icon where no specific purpose is required.
```

index.html
```diff
@@ -1611,7 +1601,7 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where no specific purpose is required.
```
marcoscaceres, line 1604:
```suggestion
            The user agent is free to display the icon where no [=manifest image resource/purpose=] is required.
```

index.html
```diff
@@ -1611,7 +1601,8 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where a more specialized
```
dmurph: Discussion idea: `The user agent can present this icon when a specific purpose is not required.`

index.html
```diff
@@ -1611,7 +1601,7 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where no [=manifest image resource/purpose=] is required.
```
marcoscaceres, line 1604:
```suggestion
            The user agent is free to display the icon where no [=manifest image resource/purpose=] is required.
            For example, an "any" [=manifest image resource=] wouldn't be used in a context where "[=icon purpose/monochrome=]" is required.
```

index.html
```diff
@@ -1611,7 +1601,10 @@ <h3>
             <dfn data-dfn-for="icon purpose">any</dfn> (default)
           </dt>
           <dd>
-            The user agent is free to display the icon in any context.
+            The user agent is free to display the icon where no [=manifest
+            image resource/purpose=] is required. For example, an "any"
+            [=manifest image resource=] wouldn't be used in a context where
+            "[=icon purpose/monochrome=]" is required.
```
marcoscaceres, line 1607:
```suggestion
            image resource/purpose=] is required. For example, a
            [=manifest image resource=] with a "any" purpose 
            wouldn't be used in a context where "[=icon purpose/monochrome=]" 
            is required.
```

index.html
```diff
@@ -649,16 +649,6 @@ <h3>
           an <abbr title="Operating system">OS</abbr>'s task switcher and/or
           system preferences.
         </p>
-        <p>
-          If there are multiple equally appropriate images in
```
dmurph, line 653: Some ideas:

```
When a user agent is choosing an icon, if multiple images are equally appropriate, a user agent SHOULD use the last one declared. This is intended to help the choice be deterministic across user agents. For example, if two images are declared with the same metadata, the last one should be chosen.
```

```
To make icon choice more deterministic between user agents, if a user agent determines that multiple icons are equivalent candidates for a given use, then the one defined last in [resources] should be used.
```

mgiuca:
+1 to the general idea of this: it's overly restrictive and UAs should be able to choose the best icon of all the icons (assuming they follow the requirements of the "purpose" field).

index.html
dmurph: After chatting, removing this, it's fine w/o it.

dmurph (approved):

PR #1122, opened 2024-05-06
Processing steps for id now removes the fragment

### LINKED ISSUE #1121: Manifest id should strip fragment at parse time
Currently, the `id` member does _not_ have its fragment stripped in the parser, but both places where the spec mentions _comparing_ identities, it uses [exclude fragments](https://url.spec.whatwg.org/#url-equals-exclude-fragments). This strongly suggests that the canonical `id` should not include a fragment, and it's creating implementation problems.

There's even an example of the `id` parsing that leaves the fragment in, which is confusing:

| json["id"] | manifest["start_url"] | manifest["id"] |
| ------------- | ------------- | ---------- |
| undefined | "https://example.com/my-app/#here" | "https://example.com/my-app/#here" |

Since for all intents and purposes, the "#here" is not part of the fragment (it would compare equal to any app id with the same origin, path and query, but a different fragment).

The problem with this is that it encourages user agents to _retain_ the fragment when storing the identity in a database, but then do comparisons ignoring fragment later. However, if the id is used as a database primary key, or hashed before being compared, it can be difficult to correctly recognize two ids as the same app if they differ only by fragment.

**The fragment should be stripped out at parse time, not at comparison time, so they represent the canonical id.**

## Historical analysis

I'm a bit confused because an earlier draft of this text in the original PR #988 included stripping out the fragment at parse time ([discussion](https://github.com/w3c/manifest/pull/988#pullrequestreview-761539027)). But it wasn't in the final version that was merged and I don't see where it got taken out. Note that Chromium [implemented stripping out the fragment at parse time](https://chromium-review.googlesource.com/c/chromium/src/+/3188751), in the same week as that PR, citing the then-draft of the manifest, and still does.

We recently ran into an issue in Chromium (CC @phoglenix) where some old app database entries mysteriously had a fragment in them, which led us to investigate this. Although we can't figure out where those old entries came from (since Chromium always stripped out the fragment), it has highlighted the dangers of having fragments in the URL used as the primary key in a database, when the fragments are supposed to be ignored for comparison purposes.

mgiuca:
Closes #1121

[Note: This technically changes normative behaviour, but does not break implementations. Any implementation that keeps the fragment and strips it out when comparing ids will behave the same as this. The new behaviour just encourages implementations to keep their local data model clean.]

Commit message:

The canonical id no longer includes the fragment. This was already effectively true (as all comparisons would exclude fragments), but now it makes it true in the processed manifest's canonical id.

Removed or made optional the exclude fragments in the comparisons. Added a lengthy note as to why user agents may still wish to exclude fragments when comparing against potentially old database entries.

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest/pull/1122.html" title="Last updated on Jun 6, 2024, 10:45 PM UTC (86c8d94)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1122/4b97432...mgiuca:86c8d94.html" title="Last updated on Jun 6, 2024, 10:45 PM UTC (86c8d94)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -858,9 +858,9 @@
           application, it SHOULD treat that manifest as a description of a
           distinct application, even if it is served from the same URL as that
           of another application. When the user agent sees a manifest where
-          |manifest|["id"] is [=url/equal=] (OPTIONAL: with
-          [=URL/equals/exclude fragments=] set to true) to the [=identity=] of
-          an already-installed application, it SHOULD be used as a signal that
+          |manifest|["id"] is [=url/equal=] (with [=URL/equals/exclude
+          fragments=] OPTIONALLY set to true) to the [=identity=] of an
+          already-installed application, it SHOULD be used as a signal that
           this manifest is a replacement for the already-installed
           application's manifest, and not a distinct application, even if it is
           served from a different URL than the one seen previously.
--- a/index.html
+++ b/index.html
@@ -626,8 +626,7 @@
           </li>
           <li>If |scope| is failure, return.
           </li>
-          <li>From |scope|, remove the [=url/query=] and [=url/fragment=]
-          components.
+          <li>Set |scope|'s [=URL/query=] and [=URL/fragment=] to null.
           </li>
           <li>If |manifest|["start_url"] is not [=URL/within scope=] of
           |scope|, return.
--- a/index.html
+++ b/index.html
@@ -865,6 +865,19 @@
           a distinct application, even if it is served from a different URL
           than the one seen previously.
         </p>
+        <p class="note" title="Exclude fragments for good measure">
+          Since the [=process the id member|processing algorithm=] removes the
+          [=URL/fragment=] from the <code>[=id=]</code> member, it is not
+          strictly necessary to [=URL/equals/exclude fragments=] when checking
+          for a matching application. However, a database of installed
+          applications can contain old or erroneous data. In particular, old
+          versions of this spec (and therefore possibly old versions of user
+          agents) did not remove the [=URL/fragment=] from the [=URL=] at parse
+          time, and relied only on [=URL/equals/exclude fragments|excluding
+          fragments=] during comparisons. Therefore, we ask that user agents
+          continue to [=URL/equals/exclude fragments=] when comparing against
+          their existing app database.
+        </p>
         <p class="note">
           The [=identity=] can be used by a service that collects lists of web
           applications to uniquely identify applications.
--- a/index.html
+++ b/index.html
@@ -879,7 +879,7 @@
           application's manifest, and not a distinct application, even if it is
           served from a different URL than the one seen previously.
         </p>
-        <p class="note" title="Excluding fragments is best practice">
+        <aside class="note" title="Excluding fragments is best practice">
           Since the [=process the id member|processing algorithm=] removes the
           [=URL/fragment=] from the <code>[=manifest/id=]</code> member, it is
           not strictly necessary to [=URL/equals/exclude fragments=] when
@@ -891,7 +891,7 @@
           the <code>[=manifest/id=]</code>. Therefore, it is best practice for
           user agents to [=URL/equals/exclude fragments=] even when comparing
           two [=URLs=] that ought not to have fragments.
-        </p>
+        </aside>
         <p class="note">
           The [=identity=] can be used by a service that collects lists of web
           applications to uniquely identify applications.
--- a/index.html
+++ b/index.html
@@ -874,7 +874,7 @@
           from the [=URL=] at parse time, and relied only on
           [=URL/equals/exclude fragments|excluding fragments=] during
           comparisons, there may be historical app data with [=URL/fragments=]
-          in the <code>[=id=]</code>. Therefore, it is best practice for user
+          in the <code>[=manifest/id=]</code>. Therefore, it is best practice for user
           agents to [=URL/equals/exclude fragments=] even when comparing two
           [=URLs=] that should not have any fragments.
         </p>
--- a/index.html
+++ b/index.html
@@ -867,7 +867,7 @@
         </p>
         <p class="note" title="Exclude fragments for good measure">
           Since the [=process the id member|processing algorithm=] removes the
-          [=URL/fragment=] from the <code>[=id=]</code> member, it is not
+          [=URL/fragment=] from the <code>[=manifest/id=]</code> member, it is not
           strictly necessary to [=URL/equals/exclude fragments=] when checking
           for a matching application. However, since old versions of this spec
           (and, possibly, old user agents) did not remove the [=URL/fragment=]
--- a/index.html
+++ b/index.html
@@ -895,6 +895,8 @@
           <li>If |id| is not [=same origin=] as |manifest|["start_url"],
           return.
           </li>
+          <li>Set |id|'s [=URL/fragment=] to null.
+          </li>
           <li>Set |manifest|["id"] to |id|.
           </li>
         </ol>
@@ -934,7 +936,7 @@
                 "https://example.com/my-app/#here"
               </td>
               <td>
-                "https://example.com/my-app/#here"
+                "https://example.com/my-app/"
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


dmurph (approved):
This is a good change - this has come up repeatedly with confusion from devs & implementers, so just removing the fragment from the beginning really helps to solve this.

marcoscaceres:
Hmm... looking at WebKit, it wasn't stripping the fragment etc. I'll need to check what the impact is.

index.html
```diff
@@ -859,13 +858,26 @@ <h3>
           application, it SHOULD treat that manifest as a description of a
           distinct application, even if it is served from the same URL as that
           of another application. When the user agent sees a manifest where
-          |manifest|["id"] is [=url/equal=] with [=URL serializer/exclude
-          fragment|exclude fragment true=] to the [=identity=] of an
-          already-installed application, it SHOULD be used as a signal that
+          |manifest|["id"] is [=url/equal=] (OPTIONAL: with
```
marcoscaceres, line 861: I think we should remove OPTIONAL... it's a bit confusing. Let's just assume a single model here.

index.html
```diff
@@ -859,13 +858,26 @@ <h3>
...
+          [=URL/equals/exclude fragments=] set to true) to the [=identity=] of
+          an already-installed application, it SHOULD be used as a signal that
           this manifest is a replacement for the already-installed
           application's manifest, and not a distinct application, even if it is
           served from a different URL than the one seen previously.
         </p>
+        <p class="note" title="Exclude fragments for good measure">
+          Since the [=process the id member|processing algorithm=] removes the
+          [=URL/fragment=] from the <code>[=id=]</code> member, it is not
```
marcoscaceres, line 870: Should this be? (I can't remember... I thought we had scoped them)

```suggestion
          [=URL/fragment=] from the <code>[=manifest/id=]</code> member, it is not
```

index.html
mgiuca: The purpose of OPTIONAL here is that some UAs may decide not to do it if they are sure the fragments were already stripped. (e.g. if they always strip fragments before writing them into the database, and there never was a time when they didn't do that, or they ran a migration at some point stripping fragments from the database, then they don't need to exclude fragments here).

Not doing so could actually significantly help those clients, for example, they could then compare the hashes of the app IDs rather than having to perform a URL equals operation (which can't be done using a SQL query). This doesn't only affect major browsers, but also any website that indexes apps -- we'd like it to be optional for them so they can simply assume the fragment-removed app IDs are primary keys in their database and do simple lookups. If we made this mandatory, then _every_ entity that indexes apps would need to have no canonical representation of an app ID, and always do the URL equals excluding fragments comparison.

So I did go back and forth on this a bit, but settled on making it OPTIONAL by way of saying "hey, if you've been storing app IDs with their fragments, you probably should exclude fragments when you compare them, but you don't have to do that if you've treated the data properly per the algorithm, and there are significant simplifications to the data model if you don't do this."

index.html
mgiuca: Yeah this is correct. Unfortunately the file now needs a `tidy` and I'm not at my PC until next week. Let's agree in principle and I'll fix it up and merge next week at my desk.

index.html
marcoscaceres: Right, yeah. that seems ok. I wonder if it could be better phrased though the "(OPTIONAL: ...)" looks a bit odd to me. Trying to think of a better way to phrase it.

index.html
```diff
@@ -859,13 +858,26 @@ <h3>
...
-          fragment|exclude fragment true=] to the [=identity=] of an
-          already-installed application, it SHOULD be used as a signal that
+          |manifest|["id"] is [=url/equal=] (OPTIONAL: with
+          [=URL/equals/exclude fragments=] set to true) to the [=identity=] of
+          an already-installed application, it SHOULD be used as a signal that
           this manifest is a replacement for the already-installed
           application's manifest, and not a distinct application, even if it is
           served from a different URL than the one seen previously.
         </p>
+        <p class="note" title="Exclude fragments for good measure">
```
marcoscaceres, line 868:
```suggestion
        <p class="note" title="Exclude fragments for comparison">
```

index.html
```diff
@@ -859,13 +858,26 @@ <h3>
...
+          [=URL/fragment=] from the <code>[=manifest/id=]</code> member, it is not
+          strictly necessary to [=URL/equals/exclude fragments=] when checking
+          for a matching application. However, since old versions of this spec
+          (and, possibly, old user agents) did not remove the [=URL/fragment=]
+          from the [=URL=] at parse time, and relied only on
+          [=URL/equals/exclude fragments|excluding fragments=] during
+          comparisons, there may be historical app data with [=URL/fragments=]
+          in the <code>[=manifest/id=]</code>. Therefore, it is best practice for user
+          agents to [=URL/equals/exclude fragments=] even when comparing two
+          [=URLs=] that should not have any fragments.
```
marcoscaceres, line 879: Removing the use of RFC2119 keywords: 

```suggestion
          for a matching application. However, since old versions of this spec
          (and, possibly, old user agents) did not remove the [=URL/fragment=]
          from the [=URL=] at parse time, and relied only on
          [=URL/equals/exclude fragments|excluding fragments=] during
          comparisons, historical app data could include [=URL/fragments=]
          in the <code>[=manifest/id=]</code>. Therefore, it is best practice for user
          agents to [=URL/equals/exclude fragments=] even when comparing two
          [=URLs=] that typically lack fragments.
```

marcoscaceres (approved):
Couple of suggestions. I can live with the (OPTINAL…) if it’s a pain to change. Otherwise looks good.

index.html
mgiuca: Rephrased as "is best practice". The "good measure" was supposed to convey the fact that you don't really need to do it, but if you can then just do it anyway to sanitize bad data.

I hope this also conveys that user agents _don't_ have to do this and it can be beneficial not to, because it affords optimizations like hashing.

index.html
mgiuca: Done. (Wow I was away from my desk for 2 weeks due to illness :sneezing_face: )

index.html
mgiuca: Thanks, good catch! I rephrased these a bit but got rid of the RFC words.

index.html
mgiuca: @marcoscaceres input needed: I rewrote it as "(with exclude fragments OPTIONALLY set to true)". Is this sufficient to capture RFC 2119's keyword "OPTIONAL"? (Note that Respec doesn't italicize the word "OPTIONALLY" nor does this adverbial form appear in RFC 2119)

Otherwise if it has to exactly use those words, I think the best phrasing is the somewhat verbose: "(the user agent MAY set exclude fragments to true)".

mgiuca:
@marcoscaceres Thanks, I finally got back on this one. Addressed the comments, PTAL.

index.html
marcoscaceres: Hmmm 🤔 let’s go with MAY as I just know people will complain otherwise. 

I’ll try to see if I can think of some other phrasing in the future. 

Really sorry about all the churn on this! 

It should be good to merge otherwise (and as a bonus 🍨, it this helped fix two bugs in WebKit, so massive thank you!!!)

index.html
```diff
@@ -873,13 +872,26 @@ <h3>
...
-          |manifest|["id"] is [=url/equal=] with [=URL serializer/exclude
-          fragment|exclude fragment true=] to the [=identity=] of an
+          |manifest|["id"] is [=url/equal=] (with [=URL/equals/exclude
+          fragments=] OPTIONALLY set to true) to the [=identity=] of an
           already-installed application, it SHOULD be used as a signal that
           this manifest is a replacement for the already-installed
           application's manifest, and not a distinct application, even if it is
           served from a different URL than the one seen previously.
         </p>
+        <p class="note" title="Excluding fragments is best practice">
```
marcoscaceres, line 882:
```suggestion
        <aside class="note" title="Excluding fragments is best practice">
```

index.html
```diff
@@ -873,13 +872,26 @@ <h3>
...
+          not strictly necessary to [=URL/equals/exclude fragments=] when
+          checking for a matching application. However, since old versions of
+          this spec (and, possibly, old user agents) did not remove the
+          [=URL/fragment=] from the [=URL=] at parse time, and relied only on
+          [=URL/equals/exclude fragments|excluding fragments=] during
+          comparisons, historical app data could contain [=URL/fragments=] in
+          the <code>[=manifest/id=]</code>. Therefore, it is best practice for
+          user agents to [=URL/equals/exclude fragments=] even when comparing
+          two [=URLs=] that ought not to have fragments.
+        </p>
```
marcoscaceres, line 894:
```suggestion
        </aside>
```

PR #1153, opened 2024-11-13
Link incubations

christianliebel:
This PR links the incubations repository instead of enumerating the incubated extensions. It's hard to keep that in sync, and the manifest incubations repository has way more incubations than just BIP. The share target incubation is also linked in manifest incubations.

Commit message:

Editorial: Link incubations

* chore:
* editorial:
* BREAKING CHANGE:

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1153.html" title="Last updated on Mar 20, 2025, 2:15 PM UTC (081d45a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1153/5088e7b...081d45a.html" title="Last updated on Mar 20, 2025, 2:15 PM UTC (081d45a)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@ the [Web Applications Working Group](https://www.w3.org/2019/webapps/).
 ## Useful links
 * [Explainer](https://github.com/w3c/manifest/blob/gh-pages/explainer.md)
 * [The Web Application Manifest specification](https://www.w3.org/TR/appmanifest/)
+* [Manifest incubations](https://github.com/WICG/manifest-incubations)
 * [App Information supplement](https://github.com/w3c/manifest-app-info)
 * [Use cases and requirements](https://w3c-webmob.github.io/installable-webapps/)
 * [The Web Applications WG homepage](https://www.w3.org/2019/webapps/)
--- a/index.html
+++ b/index.html
@@ -3227,11 +3227,10 @@
             for a second user agent on a different platform to make use of that
             member, even if no other user agent has expressed interest right
             now). If so, we ask authors to design the API in a vendor-neutral
-            way, and propose it as a standard (see [[[#incubations]]]). If the
-            new member is truly proprietary (i.e. will only ever make sense in
-            the context of a proprietary ecosystem), use this process, and
-            prefix it with the short name of that proprietary ecosystem to
-            avoid name collisions.
+            way, and propose it as a standard. If the new member is truly
+            proprietary (i.e. will only ever make sense in the context of a
+            proprietary ecosystem), use this process, and prefix it with the
+            short name of that proprietary ecosystem to avoid name collisions.
           </p>
           <p>
             Do not use vendor prefixes that you intend to later remove once it
--- a/index.html
+++ b/index.html
@@ -3265,45 +3265,6 @@
       </section>
     </section>
     <section class="appendix informative">
-      <h2 id="incubations">
-        Incubations
-      </h2>
-      <p>
-        Extensions to this specification are being incubated in parallel by the
-        Web Community, some of which are shipping in multiple browsers. If two
-        or more browser engines end up supporting an incubated feature, then
-        those features will become part of this specification in the future -
-        allowing them to become a standard the Web Platform:
-      </p>
-      <dl>
-        <dt>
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -3270,38 +3270,17 @@ <h2 id="incubations">
...
-        those features will become part of this specification in the future -
-        allowing them to become a standard the Web Platform:
+        Web Community, some of which are shipping in multiple browsers. All
+        incubations can be found in the <a href=
+        "https://github.com/WICG/manifest-incubations">Manifest Incubations
+        repository</a> at the <a href="https://wicg.io"><abbr title=
+        "Web Incubator Community Group">WICG</abbr></a>.
+      </p>
+      <p>
+        If two or more browser engines end up supporting an incubated feature,
```
marcoscaceres, line 3280:
```suggestion
        If two or more browser engines end up expressing support for an incubated feature,
```

index.html
```diff
@@ -3270,38 +3270,17 @@ <h2 id="incubations">
       </h2>
       <p>
         Extensions to this specification are being incubated in parallel by the
-        Web Community, some of which are shipping in multiple browsers. If two
-        or more browser engines end up supporting an incubated feature, then
-        those features will become part of this specification in the future -
-        allowing them to become a standard the Web Platform:
+        Web Community, some of which are shipping in multiple browsers. All
+        incubations can be found in the <a href=
+        "https://github.com/WICG/manifest-incubations">Manifest Incubations
```
dmurph, line 3275: I think this isn't quite true, we have a number of repositories, and this one hosts a bunch of stuff, but we also have:
- https://wicg.github.io/web-app-launch/
- https://wicg.github.io/web-share-target/
- maybe more? We can just list them though.

dmurph (dismissed):

index.html
```diff
@@ -3270,38 +3270,17 @@ <h2 id="incubations">
...
+        repository</a> at the <a href="https://wicg.io"><abbr title=
+        "Web Incubator Community Group">WICG</abbr></a>.
+      </p>
+      <p>
+        If two or more browser engines end up supporting an incubated feature,
+        then those features will become part of this specification in the
+        future - allowing them to become a standard of the Web Platform.
```
marcoscaceres, line 3282: Can you link too the charter, particularly to scope? 

```suggestion
        then those features can be eligible to become part of this specification in the
        future (so long as they are in scope of the Working Group's charter) - allowing them to become a standard of the Web Platform.
```

index.html
christianliebel: I think it would be better if we could link to Manifest Incubations as the single document that refers to all incubations. This way, we can avoid noise in this spec.

The [`share_target` member](https://wicg.github.io/manifest-incubations/#share_target-member) is already listed in Manifest Incubations and links to the Share Target spec. The Web App Launch Handler API is linked in the references section, but it seems like the document doesn’t use the reference right now. We could add the `launch_handler` member and link to its spec in the same way.

WDYT? @dmurph @marcoscaceres

dmurph: Sure, whatever works

marcoscaceres: Hmmm... I don't think this should be in the spec at all, to be honest. 

Incubations should really be linked from the README.md of the repo IMO.

christianliebel: Updated 👍

dmurph: It seems a little sad that we have to hide this link in the explainer, as it seems important. But not a big deal IMO - changes seem fine.

README.md
```diff
@@ -4,6 +4,7 @@ the [Web Applications Working Group](https://www.w3.org/2019/webapps/).
 ## Useful links
 * [Explainer](https://github.com/w3c/manifest/blob/gh-pages/explainer.md)
 * [The Web Application Manifest specification](https://www.w3.org/TR/appmanifest/)
+* [Manifest incubations](https://github.com/WICG/manifest-incubations)
```
dmurph, line 7: add https://wicg.github.io/web-share-target/ too?

marcoscaceres: It's linked to already from the incubations, so we probably don't need to add it here.

marcoscaceres (approved):

PR #1163, opened 2025-02-07
Editorial: Add installable web application definition

dmurph:
Closes #1097

This change (choose at least one, delete ones that don't apply):

* Makes editorial changes (changes informative sections, or changes normative sections without changing behavior)

Commit message:

Adds a formal definition for 'installable web application'.

Person merging, please make sure that commits are squashed with one of the following as a commit message prefix:

* chore:
* editorial:
* BREAKING CHANGE:
* And use none if it's a normative change

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest/pull/1163.html" title="Last updated on Feb 7, 2025, 12:28 AM UTC (9046a16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1163/ec8a23b...dmurph:9046a16.html" title="Last updated on Feb 7, 2025, 12:28 AM UTC (9046a16)">Diff</a>

Issue #1111, opened 2024-02-29
When generating a 'default' manifest (where there wasn't any linked), manifest_url doesn't exist


dmurph:
I don't think there is an explicit problem here, as when generating the default manifest, nothing that uses the manifest_url is actually used (we just create the start_url, id, and scope).  But many places reference this url in the algorithms, and in this case it doesn't exist.

marcoscaceres:
Ok... so we need to make it clear that when there is no link relationship, there's still the processed manifest.

marcoscaceres:
Closing in favor or https://github.com/w3c/manifest/issues/784

PR #1164, opened 2025-02-08
editorial: Refine the definition of an installable web application

benfrancis:
PR #1163 added a definition of installable web applications. After code review comments had been addressed the resulting text included the sentence "Any website is an installable web application." I don't think that the vast majority of web developers would agree with this statement, and to me it seems to directly conflict with the text that follows which defines an _installed_ web application as one which has (or will have) a manifest applied to its top level browsing context. To me this clearly indicates that a website can not be an installable web application unless it at least has a web application manifest, even if only an empty manifest with defaults applied.

This is a suggestion to refine that text to address my review comments in #1163.

Specifically:
- What differentiates an installable web application from other websites is that it links to a web application manifest
- A manifest is applied when an installed web application is launched

This change:

Commit message:

Refine the definition of an installable web application

* chore:
* editorial:
* BREAKING CHANGE:


christianliebel:
@benfrancis Thank you for your pull request. The updated definition is the result of a discussion at TPAC 2023 where various vendors [reached consensus](https://docs.google.com/document/d/1QzHTKGDxHol7KybdqY-rll1VGd1FIirvW-mtAFxPjC4/edit?tab=t.0#heading=h.5wzb241vwwm2) that all websites should be installable. On iPhone OS, since launch, it has been possible to add any website to the homescreen without requiring a manifest. Chrome has since removed the installability criteria and partially turned it into promotability criteria. As a web developer, I am glad that the manufacturers were able to agree on a common definition. Therefore, the current definition of an installable web application fits from my point of view.

benfrancis:
@christianliebel It's cool if the browser vendors at TPAC 2023 were able to reach a consensus of some kind (though I can't see a clear consensus or resolution in those meeting notes), but as an implementer what I'm saying is that the current text of the specification seems nonsensical because it appears to contradict itself.

If an "installed" web application is one where a web application manifest has been applied to a top level browsing context to create an application context, then how can a website without a manifest be "installable"?

Apart from this, in both theory and practice, there is a difference between creating a bookmark to a single URL which opens in a browser tab with an unbounded navigation scope, and creating an app launcher which launches a standalone application context (usually presented like a native app window) with a manifest applied to all URLs which fall within its navigation scope.

Putting aside the constraints of the current UI paradigms of Android and iOS and the terms "add to homescreen" vs. "install", under the current specification it's not possible to create an application context without a manifest, and it's not possible to process a manifest without both a document URL and a manifest URL.

So how do you get from a website with no manifest to an "installed" web application? What am I missing?

dmurph:
There is a default manifest for any page. This contains a start_url, manifest_id, and scope. This can be used, if desired by the user agent, to create an installed web application.

benfrancis:
@dmurph wrote:
> There is a default manifest for any page.

I've heard people talk about this but I can't find it defined anywhere. What are the defaults and how are they derived? What do you use as manifest URL when processing the default manifest?

Is this a proprietary Chromium thing? As far as I can tell a default manifest isn't part of any web standard.

benfrancis:
> What do you use as manifest URL when processing the default manifest?

Oh, I see you already filed an issue about this in #1111.

Putting aside the issue that the manifest processing algorithm requires a manifest URL as an input (which is a problem), I'm guessing the defaults you mentioned could come from:

- `start_url` is set to document URL (and returns before getting to the bit about manifest URL)
- `id` is set to `start_url`
- `scope` is set to `.` resolved against `start_url` as a base

So whilst I can see it's _possible_ to create a default manifest from just a document URL, as far as I can tell the concept of a default manifest is not actually defined in the specification. What the specification actually says is:

_"A manifest has an associated manifest URL, which is the [[URL](https://w3c.github.io/manifest/#bib-url)] from which the [manifest](https://w3c.github.io/manifest/#dfn-manifest) was fetched."_

The idea of a default manifest feels like an implementation detail which has become an assumption but not actually specified anywhere..?

I'm not necessarily opposed to the idea that all websites are installable web apps if that's the consensus of the Working Group. Whilst I don't think most web developers would currently see it that way, and that's not how it's documented in places like [MDN](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/js13kGames/Installable_PWAs) and [web.dev](https://web.dev/learn/pwa/installation), it is an idea we experimented with back in Firefox OS 3.0 and I don't think it's a terrible idea. I just don't think that's currently the common understanding.

I maintain that as the specification is currently written, a web site can only become an "installed" web application if it links to web application manifest. So if this definition is going to stick, a lot more changes are needed for it to make sense.

marcoscaceres:
@benfrancis, for example, if you go to https://example.com in Safari on Desktop, and you "Add to Dock", it creates an installed web application. 

So, you are not wrong that the technical aspects are not stitched up together at well as they could be, but that above shows the intent of the text in that *any* website, regardless of an explicit inclusion of a web manifest, can be an installed web application. That is precisely what the original intent was for that text. 

The bit that might be missing its that all websites get assigned a processed manifest, even if the manifest is nonexistent.

So, the actual bug is in the HTML spec section on the manifest link relationship type (or perhaps in our spec)... that somewhere it needs to say that if there is not link relationship, just assume a null manifest and apply the default (with the browser being allowed to override any defaults it chooses, the way that Safari does)

Would that make more sense?

marcoscaceres:
I think we should move this to an issue and deal with it there understanding what we need to fix here and in the HTML spec.

marcoscaceres:
Just want to reiterate that @benfrancis is on point about the spec being buggy with respect to the defaults... it's not clear in the spec, as the spec does indeed rely on there being a `<link>`... we need to break (by fixing in the spec) that assumption.

benfrancis:
@marcoscaceres wrote:
> I think we should move this to an issue

OK, it seems there isn't a consensus around the fix proposed in this PR so I have filed #1168 for further discussion.

benfrancis:
@marcoscaceres wrote:
> if you go to https://example.com/ in Safari on Desktop, and you "Add to Dock"

BTW this also assumes that "add to dock" and "add to homescreen" are the same thing as "install", which i think is debatable.

> Would that make more sense?

I think that would signal a significant shift in the common understanding of what an installable web application is, but perhaps it's OK for this Working Group to decide that...

Issue #1168, opened 2025-03-20
The new definition of "installable web application" is inconsistent with the normative text of the specification


benfrancis:
This is a follow-up to #1164.

PR #1163 added a definition of installable web application which says:
 
> Any website is an installable web application.

However, as the specification is currently written a web site can only become an "installed" web application if it links to a web application manifest. This is because an "installed" web application is one where a web application manifest has been [or will be] applied to a top level browsing context to create an application context. The algorithm for processing a manifest takes a manifest URL as a required input, and therefore a manifest can not be processed unless a website links to a manifest URL. Therefore I assert that not every website is an "installable web application".

I proposed a fix to this issue in #1164 by modifying the definition of an installable web application such that:
- What differentiates an installable web application from other websites is that it links to a web application manifest
- A manifest is applied when an installed web application is launched

Other members have pointed out that in some existing implementations any website can be "added to the homescreen" and that in order to do that a "default manifest" is assumed. Whilst this may have become an assumption for some implementers, the concept of a default manifest is not defined anywhere in the specification and (as also highlighted in #1111) the algorithm for processing a manifest requires a manifest URL.

My suggestion is to modify the (non-normative) definition of "installable web application" in order to resolve this issue, as above.

However, if the consensus of the Working Group is that all websites are "installable web applications" (which also implies that all websites are web applications BTW), then normative changes will be needed to the specification to deal with the situation where a website does not provide a manifest.

benfrancis:
FWIW Firefox Desktop currently appears to be going in the direction of "any website can be an installable web application" too https://connect.mozilla.org/t5/discussions/how-can-firefox-create-the-best-support-for-web-apps-on-the/m-p/60561/highlight/true

Mozilla may have views on how to specify this based on their recent experiments in Nightly and I'm curious to see how they deal with all the edge cases.

benfrancis:
As it stands I still disagree with the statement "Any website is an installable web application."

See https://tola.me.uk/blog/2026/02/12/how-i-think-web-apps-should-be-installed:
> I don’t agree with this and think it’s actually inconsistent with the normative text of the specification, which says a web site can only become an “installed” web application if it links to a web application manifest.

[...]

PR #1175, opened 2025-06-06
Add `navigator.install()`

christianliebel:
This is a first pass for introducing `navigator.install()`. The goal is to specify a method that allows same-origin installs (current and background documents).

This draft currently only covers same-document installs (zero parameters) and does not take manifest IDs into account yet. 

Relevant documents:
* [Web Install API Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/WebInstall/explainer.md#suite-of-web-apps)
* [TAG feedback](https://github.com/w3ctag/design-reviews/issues/1051)
* [Discussion about requiring manifest ID](https://docs.google.com/document/d/19dad0LnqdvEhK-3GmSaffSGHYLeM0kHQ_v4ZRNBFgWM/edit?tab=t.0#heading=h.koe6r7c5fhdg)
* [The cross-browser future of Installable Web Apps (TPAC 2023)](https://docs.google.com/document/d/1QzHTKGDxHol7KybdqY-rll1VGd1FIirvW-mtAFxPjC4/edit?tab=t.0#heading=h.5wzb241vwwm2)

Acknowledgements:
* The algorithm steps are largely inspired by the Web Share API.

---

Closes #???

Commit message:

Add API for installing web applications

* chore:
* editorial:
* BREAKING CHANGE:

### DIFF
--- a/index.html
+++ b/index.html
@@ -2434,6 +2434,15 @@
         [=applied=].
       </p>
       <p>
+        At the discretion of the operating system or user agent, web
+        applications may be [=installed=] once or multiple times.
+      </p>
+      <aside class="note" title="Multiple installations">
+        User agents may allow multiple installations of the same web
+        application, for example, so they can be used with different profiles,
+        such as "work" or "private" profiles.
+      </aside>
+      <p>
         Once a web application is [=installed=] it is known as a <dfn class=
         "export">installed web application</dfn>: That is, the manifest's
         members, or their defaults, are [=applied=] to the <a>top-level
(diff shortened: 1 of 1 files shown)


benfrancis:
I'm fascinated to see how this pans out given we now seem to have gone full circle from [`navigator.mozApps.install()`](https://hacks.mozilla.org/2013/02/getting-started-with-open-web-apps-why-and-how/) in ~2012, and in the intervening time browser vendors have gone to great lengths (installability criteria, install prompts etc.) to avoid doing this.

Some questions:
- Is the intention still to eventually [require that a manifest contains an `id` member](https://docs.google.com/document/d/19dad0LnqdvEhK-3GmSaffSGHYLeM0kHQ_v4ZRNBFgWM/edit?tab=t.0#heading=h.2ex5jpno4wi8) in order for installation to succeed?
- If `navigator.install()` only works if the document links to a valid web app manifest, does this not further reinforce the idea that [a web app needs a web app manifest in order to be installable](https://github.com/w3c/manifest/issues/1168) (contrary to the [definition recently added to the specification](https://github.com/w3c/manifest/pull/1163))?
- Are we OK with the idea that this will cause the web to be littered with install buttons (which as I understand it is the reason browser vendors have avoided doing this so far)?

I think on balance my advice (in the capacity of an Invited Expert) would be to *not* add this feature to the web platform, but to instead focus on how to improve the user interface that user agents provide to install web apps.

It is already possible for browser vendors to implement an "install app" feature inside user agents, which could be a more prominent and trusted user interface than an install button in each web app (e.g. an ambient indicator in a URL bar to indicate the current document is part of an installable web app, and a prominent "install app" button). Instead this feature currently tends to get buried inside multiple levels of menus or use obscure language like "add to home" rather than "install". Unfortunately I think this poor user experience is a result of misaligned incentives of the current incumbent "user agents" rather than a lack of creativity in user interface design.

If browser vendors continue to bury and obfuscate the install app feature then I can see why this API would be necessary, but I think the resulting user experience (spammy install buttons and prompts everywhere) would be worse for everyone.

---

As an aside, this is another example of something that could be much simpler to implement if a web app was identified by its manifest URL (`navigator.install({manifest URL})`).

christianliebel:
@benfrancis This is all in flux. Drafting this first specification text was just my initiative to have a basis for discussion. No vendor has committed to implementing this.

> I'm fascinated to see how this pans out given we now seem to have gone full circle from [`navigator.mozApps.install()`](https://hacks.mozilla.org/2013/02/getting-started-with-open-web-apps-why-and-how/) in ~2012, and in the intervening time browser vendors have gone to great lengths (installability criteria, install prompts etc.) to avoid doing this.

Wow, interesting historical insights. I only joined this effort in 2018…

> Is the intention still to eventually [require that a manifest contains an id member](https://docs.google.com/document/d/19dad0LnqdvEhK-3GmSaffSGHYLeM0kHQ_v4ZRNBFgWM/edit?tab=t.0#heading=h.2ex5jpno4wi8) in order for installation to succeed?

This is up to the Working Group to decide. Safari has traditionally allowed the installation of all websites regardless of any preconditions. This would also suggest `navigator.install(documentURL)` so a linked manifest can be loaded, or the default manifest can be created for the document at the given URL. As a web developer, it would be great to have a single way to trigger the installation of an app, regardless of the platform.

> If navigator.install() only works if the document links to a valid web app manifest, does this not further reinforce the idea that https://github.com/w3c/manifest/issues/1168 (contrary to the https://github.com/w3c/manifest/pull/1163)?

(see above)

> Are we OK with the idea that this will cause the web to be littered with install buttons (which as I understand it is the reason browser vendors have avoided doing this so far)?

We already have `beforeinstallprompt` in Chromium-based browsers (see the related discussion in https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/1055) to do this, and haven't seen a larger littering of install buttons yet.

> I think on balance my advice (in the capacity of an Invited Expert) would be to _not_ add this feature to the web platform, but to instead focus on how to improve the user interface that user agents provide to install web apps.

The idea of the [Web Install API](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/WebInstall/explainer.md) would be to allow same-origin installations as well, e.g. _example.com_ could suggest to install _example.com/docs_ or _example.com/sheets_. Same-origin installs also have TAG approval. I think improving the install UX is an orthogonal effort.

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+            represents a user's current intent to install an application. It is
+            initialized to `null`.
+          </dd>
+        </dl>
+        <h3>
+          <dfn>install()</dfn> method
+        </h3>
+        <p>
+          When the {{Navigator/install()}} method is called, run the following
+          steps to [=install=] the website.
```
marcoscaceres, line 2648:
```suggestion
          steps:
```

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+        </p>
+        <ol class="algorithm">
+          <li>Let |global:Window| be [=this=]'s [=relevant global object=].
+          </li>
+          <li>Let |document:Document| be |global|'s [=associated Document=].
+          </li>
+          <li>If |document| is not [=Document/fully active=], return [=a
+          promise rejected with=] an {{"InvalidStateError"}} {{DOMException}}.
+          </li><!-- todo: permission policy? -->
```
marcoscaceres, line 2657: Instead of a permission policy, this should throw with a NotAllowedError if the browsing context is not the top-level traversable.

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+          </li>
+          <li>Set [=this=].{{Navigator/[[installPromise]]}} to be <a>a new
+          promise</a>.
+          </li>
+          <li>Return [=this=].{{Navigator/[[installPromise]]}} and <a>in
+          parallel</a>:
+            <ol>
+              <li>[=Processing|Process=] the |document|'s [=manifest=].
+              </li>
+              <li>If the |document|'s [=Document/processed manifest=] is null,
```
marcoscaceres, line 2675: I don't think the processed manifest can ever be null. It will have the defaults.

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+                  <li>Set [=this=].{{Navigator/[[installPromise]]}} to `null`.
+                  </li>
+                  <li>Terminate this algorithm.
+                  </li>
+                </ol>
+              </li>
+              <li>Present an installation UI for the |document|'s
+              [=Document/processed manifest=].
+                <aside class="note">
+                  If a user agent only allows one instance of the application
```
marcoscaceres, line 2691: There's a lot of RFC2119 language here, which is not appropriate for a note.

marcoscaceres:
@benfrancis' outlines the core concerns, which remain... and I tend to share them (spammy install buttons). At the same time, it could lead to a smoother install experience and we haven't seen much movement with user agents improving the install experience... this might jolt them into action? Not sure. 

Also, the web has interop issues now because of BIP, which, if made predictable, would provide the same functionality - but with a much clunkier API. It's a hard balance to strike. 

And no, it shouldn't require an explicit manifest or the inclusion of an id in the manifest: it should just work (present the install UI of a UA when the method is called).

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+                </ol>
+              </li>
+              <li>Otherwise, [=queue a global task=] on the [=user interaction
+              task source=] using |global| to:
+                <ol>
+                  <li>[=Resolve=] [=this=].{{Navigator/[[installPromise]]}}
+                  with `undefined`.
+                  </li>
+                  <li>Set [=this=].{{Navigator/[[installPromise]]}} to `null`.
+                  </li>
```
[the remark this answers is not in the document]
christianliebel: If I set the promise to null, I can't resolve it in the next step.

index.html
christianliebel: Removed the note.

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+                </ol>
+              </li>
+              <li>Present an installation UI for the |document|'s
+              [=Document/processed manifest=].
+                <aside class="note">
+                  If a user agent only allows one instance of the application
+                  to be [=installed=], it may show a launch UI instead. In this
+                  case, the user agent should return the result of the launch
+                  prompt. For privacy reasons, this API should not reveal
```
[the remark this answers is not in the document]
christianliebel: Removed the note.

index.html
```diff
@@ -2599,6 +2608,124 @@ <h3>
...
+          </li>
+          <li>If |global| does not have [=transient activation=], return [=a
+          promise rejected with=] a {{"NotAllowedError"}} {{DOMException}}.
+          </li>
+          <li>[=Consume user activation=] of |global|.
+          </li>
+          <li>Set [=this=].{{Navigator/[[installPromise]]}} to be <a>a new
+          promise</a>.
+          </li>
+          <li>Return [=this=].{{Navigator/[[installPromise]]}} and <a>in
```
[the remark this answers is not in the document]
christianliebel: The step now says "Present an installation UI for the document's manifest."

PR #1202, opened 2026-01-29
Clarify applying manifest to an existing browsing context

### LINKED ISSUE #784: Clarify behaviour of a page linking to a manifest that is not within scope of that manifest

christianliebel:
Closes #784

This change:

Adds normative text to the [Applying the manifest](https://w3c.github.io/manifest/#applying) section covering the case where a user agent applies a manifest to an *existing* top-level browsing context (e.g., immediately after installation). The two cases are now specified:

- If the active document's URL is within scope: the browsing context becomes an application context.
- If the active document's URL is not within scope: the behavior is implementation-defined.

An informative note illustrates the implementation-defined case (e.g., Chromium's tab-promotion behavior).

The original note added by this PR has been moved from the navigation scope section to sit adjacent to the new normative text, and revised for consistency (replacing `[=launch=]` with `[=apply=]`, which correctly describes applying to an existing context rather than creating a new one).

Implementation commitment: no implementation changes required — the normative text formalizes existing behavior and explicitly leaves the out-of-scope case as implementation-defined.

Commit message:

Add normative text for applying manifest to existing browsing context

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1202.html" title="Last updated on Apr 9, 2026, 5:30 PM UTC (4d7e980)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1202/6731c8c...4d7e980.html" title="Last updated on Apr 9, 2026, 5:30 PM UTC (4d7e980)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2717,6 +2717,14 @@
           third-party authentication). See <a href=
           "https://github.com/w3c/manifest/issues/646">Issue 646</a>.
         </p>
+        <p>
+          Some user agents might directly [=launch a web application|launch=]
+          the [=installed web application=] using the [=Document/URL=] of the
+          [=browsing context=] that initiated the installation. If the
+          [=Document/URL=] of this page is not [=manifest/within scope=] of the
+          manifest's [=manifest/navigation scope=], the [=application context=]
+          is initially instantiated with an out-of-scope [=Document/URL=].
+        </p>
       </aside>
       <section class="informative">
         <h3 id="navigation-scope-security-considerations">
--- a/index.html
+++ b/index.html
@@ -1733,6 +1733,25 @@
             [=top-level browsing context=] is created, the user agent MAY
             [=apply=] a manifest to it before [=navigate|navigation=] begins.
           </p>
+          <p>
+            A user agent MAY also [=apply=] a manifest to an existing [=top-level
+            browsing context=]. If the [=navigable/active document=]'s
+            [=Document/URL=] is [=manifest/within scope=] of the manifest, the
+            existing [=top-level browsing context=] becomes an [=application
+            context=]. If the [=Document/URL=] is not [=manifest/within scope=],
+            the resulting behavior is implementation-defined.
+          </p>
+          <aside class="note">
+            <p>
+              Some user agents might [=apply=] a manifest to the existing
+              [=top-level browsing context=] from which installation was
+              initiated. If the [=navigable/active document=]'s [=Document/URL=]
+              is not [=manifest/within scope=] of the manifest's
+              [=manifest/navigation scope=], one possible behavior is that the
+              [=application context=] initially exists at an out-of-scope
+              [=Document/URL=].
+            </p>
+          </aside>
           <aside class="note">
             Whether to [=apply=] a manifest, and which manifest to apply, is at
             the discretion of the user agent, based on the user's actions. For
@@ -2716,14 +2735,6 @@
           some sites that navigate to an off-scope URL (e.g., to perform
           third-party authentication). See <a href=
           "https://github.com/w3c/manifest/issues/646">Issue 646</a>.
-        </p>
-        <p>
-          Some user agents might directly [=launch a web application|launch=]
-          the [=installed web application=] using the [=Document/URL=] of the
-          [=browsing context=] that initiated the installation. If the
-          [=Document/URL=] of this page is not [=manifest/within scope=] of the
-          manifest's [=manifest/navigation scope=], the [=application context=]
-          is initially instantiated with an out-of-scope [=Document/URL=].
         </p>
       </aside>
       <section class="informative">


dmurph (approved):
LGTM

index.html
```diff
@@ -2717,6 +2717,14 @@ <h2>
           third-party authentication). See <a href=
           "https://github.com/w3c/manifest/issues/646">Issue 646</a>.
         </p>
+        <p>
+          Some user agents may directly [=launch a web application|launch=] the
+          [=installed web application=] using the [=Document/URL=] of the
+          [=browsing context=] that initiated the installation. If the
```
marcoscaceres: Just be mindful that the Document/URL and  [=browsing context=] URL could be different, maybe? ... or be more precise here. Find the right chain to the browsing context -> document -> document URL.

marcoscaceres:
After discussion with @marcoscaceres, we think this PR's note needs to be paired with normative text, and the note itself revised to be consistent with it.

## Proposed normative text

Add the following to the ["Applying the manifest"](https://w3c.github.io/manifest/#applying) section, after the sentence "the user agent MAY [=apply=] a manifest to it before [=navigate|navigation=] begins":

```html
<p>
  A user agent MAY also [=apply=] a manifest to an existing [=top-level
  browsing context=]. If the [=navigable/active document=]'s
  [=Document/URL=] is [=manifest/within scope=] of the manifest, the
  existing [=top-level browsing context=] becomes an [=application
  context=]. If the [=Document/URL=] is not [=manifest/within scope=],
  the resulting behavior is implementation-defined.
</p>
```

## Revised note

Replace the current `<aside>` added by this PR with the following (and move it to sit directly after the normative text above):

```html
<aside class="note">
  <p>
    Some user agents might [=apply=] a manifest to the existing
    [=top-level browsing context=] from which installation was
    initiated. If the [=navigable/active document=]'s [=Document/URL=]
    is not [=manifest/within scope=] of the manifest's
    [=manifest/navigation scope=], one possible behavior is that the
    [=application context=] initially exists at an out-of-scope
    [=Document/URL=].
  </p>
</aside>
```

## Rationale

- The note previously used `[=launch=]`, which per the spec creates a *new* application context — inconsistent with describing the existing-tab-promotion case.
- The normative text clarifies both branches: in-scope (browsing context becomes an application context) and out-of-scope (implementation-defined).
- Together they close #784 properly: the informative note illustrates one concrete instance of the implementation-defined behavior.

marcoscaceres (approved):
LGTM! Nice one.

christianliebel:
@dmurph @saschanaz Could you please review this change? It should align with the installation behavior of Chromium and Gecko. We wanted to clarify this in the spec to address https://github.com/w3c/manifest/issues/784.

dmurph (commented):
LGTM

dmurph:
@saschanaz Given that there is alignment between webkit and chromium folks here I'm going to approve & merge to keep velocity up (as we meet to chat about this stuff infrequently)

However! This should also come with the understanding that we should revert if there are other issues or objections where we have made mistakes. So please let us know if you have feedback that we can continue iterating on or revert this change.

PR #1204, opened 2026-01-29
[Editorial] Fixes "default display mode" terminology

### LINKED ISSUE #1024: "Default display mode" section is super confusing
The concept of "default display mode" is very confusing. Is it the mode requested by the app manifest? The mode the UA would have defaulted to if there is no request from an app manifest? The display mode that is actually in effect at the end of the day?

It would be much easier to follow this section if there were two named concepts neither of which use the word "default":
* requested display mode - the mode requested by app manifest
* effective display mode - the mode actually effective on the document after consideration of the app manifest, user preferences/overrides, and UA constraints

(It also needs to be clarified whether the `display-mode` media query reflects the effective display mode, requested display mode, or something else.)

diekus:
Closes #1024

This change:

Commit message:

Clarifies the concept of default display mode. Introduces the requested and effective display mode concept to better understand how the display mode and the media query behave.

* chore:
* editorial: Diego Gonzalez
* BREAKING CHANGE:


christianliebel (changes requested):
This change leads to conflicts with `main`, can you rebase it, please?

marcoscaceres:
this seems out of date.

PR #1218, opened 2026-05-11
Add `application-context` CSS media feature

### LINKED ISSUE #1092: Way to detect if a web app is installed
The `navigator.standalone` attribute has been in iOS since forever and has the nice semantics in that, as a boolean value, it distinguishes between installed web applications and those running in the regular browser tab. 

It's slightly unfortunate that "standalone" will mean "installed" (as per the spec's definition), and there is also "display-mode: standalone", but I think that's ok. 

What do folks here think? @dmurph, @aarongustafson, would love to hear your thoughts! 

See also this [webkit bug](https://bugs.webkit.org/show_bug.cgi?id=257806#c3).

marcoscaceres:
Closes #1092

Adds the `application-context` CSS media feature, a discrete media feature that reflects the application context in which the current top-level traversable is running. Values are `application` and `none` (default). Same-origin iframes inherit the parent value; cross-origin iframes evaluate to `none`. The value can transition from `none` to `application` when the user agent promotes a browsing context into an installed application window.

Aligned with Microsoft's [Application Context Media Feature](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ApplicationContextMediaFeature/explainer.md) explainer. Agreed in principle at TPAC 2025 (#1092). Values renamed from `installed`/`browser` to `application`/`none` per working group discussion, to remove the ambiguity between installation lifecycle state and the current running context.

* [ ] [WebKit](https://bugs.webkit.org/show_bug.cgi?id=314609)
* [ ] [Chromium](https://crbug.com/382114911)

Commit message:

Add the `application-context` CSS media feature

Documents can query `(application-context: application)` to detect whether the current top-level traversable is running as an installed web application; the value is `none` in a regular browser tab and in cross-origin nested documents.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/pull/1218.html" title="Last updated on Sep 10, 2026, 3:56 PM UTC (65bb071)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/manifest/1218/8ae3046...65bb071.html" title="Last updated on Sep 10, 2026, 3:56 PM UTC (65bb071)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3079,6 +3079,184 @@
         data-xref-for="@media">`display-mode`</a> media feature.
       </p>
     </section>
+    <section id="installed-media-feature">
+      <h2>
+        Installed context: the `installed` media feature
+      </h2>
+      <p>
+        The <dfn id="mf-installed" data-dfn-type="css-descriptor" data-dfn-for=
+        "@media" class="export">`installed`</dfn> media feature reflects
+        whether the current [=browsing context=] is running as an [=installed
+        web application=].
+      </p>
+      <table class="data">
+        <thead>
+          <tr>
+            <th>
+              Value
+            </th>
+            <th>
+              Description
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <td>
+              <dfn id="mf-installed-none" class="export" data-dfn-type="value"
+              data-dfn-for="@media/installed">none</dfn>
+            </td>
+            <td>
+              The document is not running as an [=installed web application=].
+              This is the value in a regular browser tab, inside an embedded
+              browser in a third-party application, or in any context where the
+              manifest is not [=applied=].
+            </td>
+          </tr>
+          <tr>
+            <td>
+              <dfn id="mf-installed-installed" class="export" data-dfn-type=
+              "value" data-dfn-for="@media/installed">installed</dfn>
+            </td>
+            <td>
+              The document is running as an [=installed web application=]. The
+              user agent launched the document from an operating system
+              integration surface (such as a home screen, dock, or taskbar) and
+              the document is running in an [=application context=].
+            </td>
+          </tr>
+        </tbody>
+      </table>
+      <p>
+        The <a href="#mf-installed">`installed`</a> media feature matches
+        <a href="#mf-installed-installed">`installed`</a> when the [=browsing
+        context=] is an [=application context=] with a manifest [=applied=].
+        Otherwise, it matches <a href="#mf-installed-none">`none`</a>.
+      </p>
+      <p>
+        In cross-origin child [=browsing contexts=] (e.g., cross-origin iframes
+        and fenced frames), the value MUST be <a href=
+        "#mf-installed-none">`none`</a>, regardless of the value in the
+        [=top-level browsing context=].
+      </p>
+      <p class="note">
+        Same-origin child [=browsing contexts=] inherit the value from their
+        parent, since they can already access this information via
+        <code>window.top.matchMedia("(installed)")</code>.
+      </p>
+      <p>
+        The <a href="#mf-installed">`installed`</a> media feature is orthogonal
+        to the <a data-xref-type="css-descriptor" data-xref-for=
+        "@media">`display-mode`</a> media feature. An [=installed web
+        application=] can use any [=display mode=], including [=display
+        mode/fullscreen=], [=display mode/standalone=], [=display
+        mode/minimal-ui=], or [=display mode/browser=]. The <a href=
+        "#mf-installed">`installed`</a> media feature answers "is this document
+        running as an installed application?" while `display-mode` answers
+        "what presentation mode is the browsing context using?"
+      </p>
+      <p>
+        If the user agent transitions a [=browsing context=] into an
+        [=application context=] (for example, by [=applying=] a manifest after
+        the user [=installs=] the application while the document is already
+        open), the value of <a href="#mf-installed">`installed`</a> MUST update
+        accordingly and the user agent MUST [=fire an event=] named
+        <code>change</code> at the corresponding <code>MediaQueryList</code>.
+      </p>
+      <p>
+        The value of <a href="#mf-installed">`installed`</a> in a print context
+        MUST be <a href="#mf-installed-none">`none`</a>. Installation launch
+        context is not meaningful in print output.
+      </p>
+      <p class="note">
+        The <a href="#mf-installed">`installed`</a> media feature is designed
+        to be used in a boolean context. <code>@media (installed)</code>
+        evaluates to true when the value is <a href=
+        "#mf-installed-installed">`installed`</a> and false when the value is
+        <a href="#mf-installed-none">`none`</a>. The explicit form <code>@media
+        (installed: installed)</code> is valid but the idiomatic form is the
+        boolean shorthand <code>@media (installed)</code>.
+      </p>
+      <aside class="example">
+        <p>
+          Suppress install prompts when the app is already running as
+          installed:
+        </p>
+        <pre class="css">
+@media (installed) {
+  .install-banner { display: none; }
+}
+</pre>
+        <p>
+          Adapt UI for the installed app context:
+        </p>
+        <pre class="css">
+@media (installed) {
+  .browser-nav { display: none; }
+  .share-via-os { display: block; }
+}
+</pre>
+        <p>
+          JavaScript access via <code>matchMedia()</code>:
+        </p>
+        <pre class="js">
+const installed = window.matchMedia("(installed)").matches;
+if (installed) {
+  // Running as an installed app - suppress install UI
+  document.querySelector(".install-banner")?.remove();
+}
+</pre>
+      </aside>
+      <p class="note">
+        Conformance testing for the <a href=
+        "#mf-installed-installed">`installed`</a> value requires launching the
+        document from an OS integration surface, which may not be automatable
+        in all testing environments. User agents SHOULD provide a WebDriver
+        extension or testing mode that allows setting the <a href=
+        "#mf-installed">`installed`</a> value for conformance testing purposes.
+        The <a href="#mf-installed-none">`none`</a> value can be verified
+        automatically in any regular browsing context.
+      </p>
+      <section>
+        <h3>
+          Privacy considerations for the `installed` media feature
+        </h3>
+        <p>
+          The <a href="#mf-installed">`installed`</a> media feature exposes one
+          bit of information: whether the current document is running as an
+          [=installed web application=]. A document running in a regular
+          browser tab sees <a href="#mf-installed-none">`none`</a>, even if the
+          user previously [=installed=] the application on the same device.
+        </p>
+        <p>
+          The media feature does not reveal:
+        </p>
+        <ul>
+          <li>Whether the user has previously [=installed=] this or any other
+          web application.
+          </li>
+          <li>Any user history, preferences, or OS-level state.
+          </li>
+          <li>Whether this specific document meets any user-agent-specific
+          [=installation=] criteria.
+          </li>
+        </ul>
+        <p>
+          In cross-origin child [=browsing contexts=], the value is always
+          <a href="#mf-installed-none">`none`</a>, preventing cross-origin
+          iframes from learning the [=top-level browsing context=]'s
+          installation state.
+        </p>
+        <p>
+          As with all CSS media features, a document's own stylesheets may use
+          conditional <code>@import</code> rules or resource URLs within
+          <code>@media (installed)</code> blocks, causing network requests that
+          reveal the media feature's value to third-party servers. This is a
+          general property of the CSS media query mechanism and is not unique
+          to this feature.
+        </p>
+      </section>
+    </section>
     <section id="priv-sec">
       <h2>
         Privacy and security considerations
--- a/index.html
+++ b/index.html
@@ -3079,183 +3079,183 @@
         data-xref-for="@media">`display-mode`</a> media feature.
       </p>
     </section>
-    <section id="installed-media-feature">
-      <h2>
-        Installed context: the `installed` media feature
-      </h2>
+
+    <h2 id="installed-media-feature">
+      Installed context: the `installed` media feature
+    </h2>
+    <p>
+      The <dfn id="mf-installed" data-dfn-type="css-descriptor" data-dfn-for=
+      "@media" class="export">`installed`</dfn> media feature reflects whether
+      the current [=browsing context=]'s [=top-level browsing context=] is
+      running as an [=installed web application=].
+    </p>
+    <table class="data">
+      <thead>
+        <tr>
+          <th>
+            Value
+          </th>
+          <th>
+            Description
+          </th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>
+            <dfn id="mf-installed-none" class="export" data-dfn-type="value"
+            data-dfn-for="@media/installed">none</dfn>
+          </td>
+          <td>
+            The document is not running as an [=installed web application=].
+            This is the value in a regular browser tab, inside an embedded
+            browser in a third-party application, or in any context where the
+            manifest is not [=applied=].
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <dfn id="mf-installed-installed" class="export" data-dfn-type=
+            "value" data-dfn-for="@media/installed">installed</dfn>
+          </td>
+          <td>
+            The document is running as an [=installed web application=], in an
+            [=application context=], with a manifest [=applied=]. This can
+            occur regardless of how the user agent entered that application
+            context.
+          </td>
+        </tr>
+      </tbody>
+    </table>
+    <p>
+      The <a href="#mf-installed">`installed`</a> media feature matches
+      <a href="#mf-installed-installed">`installed`</a> when the
+      [=top-level browsing context=] of the current [=browsing context=] is an
+      [=application context=]. Otherwise, it matches <a href=
+      "#mf-installed-none">`none`</a>.
+    </p>
+    <p>
+      In cross-origin child [=browsing contexts=] (e.g., cross-origin iframes
+      and fenced frames), the value MUST be <a href=
+      "#mf-installed-none">`none`</a>, regardless of the value in the
+      [=top-level browsing context=].
+    </p>
+    <p class="note">
+      Same-origin child [=browsing contexts=] inherit the value from their
+      parent, since they can already access this information via
+      <code>window.parent.matchMedia("(installed)")</code>.
+    </p>
+    <p>
+      The <a href="#mf-installed">`installed`</a> media feature is orthogonal
+      to the <a data-xref-type="css-descriptor" data-xref-for=
+      "@media">`display-mode`</a> media feature. An [=installed web
+      application=] can use any [=display mode=], including [=display
+      mode/fullscreen=], [=display mode/standalone=], [=display
+      mode/minimal-ui=], or [=display mode/browser=]. The <a href=
+      "#mf-installed">`installed`</a> media feature answers "is this document
+      running as an installed application?" while `display-mode` answers "what
+      presentation mode is the browsing context using?"
+    </p>
+    <p>
+      If the user agent transitions a [=browsing context=] into an
+      [=application context=] (for example, by [=applying=] a manifest after
+      the user [=installs=] the application while the document is already
+      open), the value of <a href="#mf-installed">`installed`</a> MUST update
+      accordingly and the user agent MUST [=fire an event=] named
+      <code>change</code> at the corresponding <code>MediaQueryList</code>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3079,6 +3079,184 @@ <h2>
...
+      <aside class="example">
+        <p>
+          Suppress install prompts when the app is already running as
+          installed:
+        </p>
+        <pre class="css">
+@media (installed) {
+  .install-banner { display: none; }
+}
+</pre>
```
```suggestion
        <pre class="css">
          @media (installed) {
            .install-banner { display: none; }
          }
        </pre>
```

index.html
```diff
@@ -3079,6 +3079,184 @@ <h2>
         data-xref-for="@media">`display-mode`</a> media feature.
       </p>
     </section>
+    <section id="installed-media-feature">
+      <h2>
+        Installed context: the `installed` media feature
+      </h2>
```
marcoscaceres: Nit... don't need this section. 
 
```suggestion
      <h2  id="installed-media-feature">
        Installed context: the `installed` media feature
      </h2>
```

marcoscaceres:
@copilot, can you fix the indenting on the `<pre>`s, including the examples. And remove the section that's not needed (see my suggestions)

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      "#mf-installed">`installed`</a> media feature answers "is this document
+      running as an installed application?" while `display-mode` answers "what
+      presentation mode is the browsing context using?"
+    </p>
+    <p data-cite="CSSOM-VIEW-1">
+      If the user agent transitions a [=browsing context=] into an
+      [=application context=] (for example, by [=applying=] a manifest after
+      the user [=installs=] the application while the document is already
+      open), the value of <a href="#mf-installed">`installed`</a> MUST update
+      accordingly and the user agent MUST [=fire an event=] named `change` at
```
marcoscaceres: need to queue a task here....

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      The <a href="#mf-installed">`installed`</a> media feature is designed to
+      be used in a boolean context. `@media (installed)` evaluates to true when
+      the value is <a href="#mf-installed-installed">`installed`</a> and false
+      when the value is <a href="#mf-installed-none">`none`</a>. The explicit
+      form `@media (installed: installed)` is valid but the idiomatic form is
+      the boolean shorthand `@media (installed)`.
+    </p>
+    <aside class="example">
+      <p>
+        Suppress install prompts when the app is already running as installed:
```
```suggestion
        Hide install prompts when the app is already running as installed:
```

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+            The document is running as an [=installed web application=], in an
+            [=application context=], with a manifest [=applied=]. This can
+            occur regardless of how the user agent entered that application
+            context.
+          </td>
+        </tr>
+      </tbody>
+    </table>
+    <p>
+      The <a href="#mf-installed">`installed`</a> media feature matches
```
marcoscaceres: This seems to duplicate the definition above... same issue with third party iframes.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      <a href="#mf-installed-installed">`installed`</a> when the [=top-level
+      browsing context=] of the current [=browsing context=] is an
+      [=application context=]. Otherwise, it matches <a href=
+      "#mf-installed-none">`none`</a>.
+    </p>
+    <p>
+      In cross-origin child [=browsing contexts=] (e.g., cross-origin iframes
```
marcoscaceres: this should be blended into the definition, not carve out an exception.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      and fenced frames), the value MUST be <a href=
+      "#mf-installed-none">`none`</a>, regardless of the value in the
+      [=top-level browsing context=].
+    </p>
+    <p class="note">
+      Same-origin child [=browsing contexts=] inherit the value from their
+      parent, since they can already access this information via
+      `window.parent.matchMedia("(installed)")`.
+    </p>
+    <p>
```
marcoscaceres: I think this is a note: 

```suggestion
    <p class="note">
```

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
         data-xref-for="@media">`display-mode`</a> media feature.
       </p>
     </section>
+    <h2 id="installed-media-feature">
+      Installed context: the `installed` media feature
+    </h2>
+    <p>
+      The <dfn id="mf-installed" data-dfn-type="css-descriptor" data-dfn-for=
+      "@media" class="export">`installed`</dfn> media feature reflects whether
```
marcoscaceres: The integration between this and "installed" or "installed web application" is not great... this section shouldn't mention browsing contexts here.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>
+            <dfn id="mf-installed-none" class="export" data-dfn-type="value"
+            data-dfn-for="@media/installed">none</dfn>
+          </td>
+          <td>
+            The document is not running as an [=installed web application=].
```
marcoscaceres: Document is probably not the right thing to check here... it's just if it's an [=installed web application=] I think.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+          </td>
+        </tr>
+        <tr>
+          <td>
+            <dfn id="mf-installed-installed" class="export" data-dfn-type=
+            "value" data-dfn-for="@media/installed">installed</dfn>
+          </td>
+          <td>
+            The document is running as an [=installed web application=], in an
+            [=application context=], with a manifest [=applied=]. This can
```
marcoscaceres: Does [=installed web application=] already imply "[=application context=], with a manifest [=applied=]"?

dmurph (commented):
Based on manifest editor meeting chat, we can remove the re-statements of manifest application on browsing contexts, as the definition of an installed web application imlpies that - e.g. it must be true if installed web application is true.

Other than that, the intention here lgtm and I'm sure there are small style things that can be fixed. but other than queueing an event (which I think marcos has in a comment) directionally lgtm

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      </p>
+      <pre class="js">
+        const installed = window.matchMedia("(installed)").matches;
+        if (installed) {
+          // Running as an installed app - suppress install UI
+          document.querySelector(".install-banner")?.remove();
+        }
+      </pre>
+    </aside>
+    <p class="note">
```
marcoscaceres, line 3279: we probably don't need this note... we are yet to figure out how to automate testing.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+      </h3>
+      <p>
+        The <a href="#mf-installed">`installed`</a> media feature exposes one
+        bit of information: whether the current document is running as an
+        [=installed web application=]. A document running in a regular browser
+        tab sees <a href="#mf-installed-none">`none`</a>, even if the user
+        previously [=installed=] the application on the same device.
+      </p>
+      <p>
+        The media feature does not reveal:
```
```suggestion
        The media feature <strong>does not</strong> reveal:
```

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
...
+        application.
+        </li>
+        <li>Any user history, preferences, or OS-level state.
+        </li>
+        <li>Whether this specific document meets any user-agent-specific
+        [=installation=] criteria.
+        </li>
+      </ul>
+      <p>
+        In cross-origin child [=browsing contexts=], the value is always
```
marcoscaceres: We should find the proper definition of "cross-origin child [=browsing contexts=]" in HTML... this lacks precision. Also, this should probably be part of the unordered list above.

index.html
```diff
@@ -3079,6 +3079,180 @@ <h2>
         data-xref-for="@media">`display-mode`</a> media feature.
       </p>
     </section>
+    <h2 id="installed-media-feature">
+      Installed context: the `installed` media feature
```
```suggestion
      The `installed` media feature
```

kyerebo:
Awesome to see that we're adding normative spec text for `installed` media feature, thank you for the work here!

Without being aware we were looking to merge this; I recently had taken a crack at an explainer/WICG proposal for this at https://github.com/WICG/proposals/issues/288

My current proposal is that we shift from a boolean media feature to an enumerated one, for reasons outlined in the [Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/AppContextMediaQuery/explainer.md) (namely context clarity and extensibility). Do you think that would make sense to reflect in the spec text here, or do we have compelling reasoning for having a boolean feature?

marcoscaceres:
I don’t mind either approach, and like the disambiguation argument raised in the explainer, which may address some concerns raised by @benfrancis in previous discussions (at least in a far that clears the “am I installed?” ambiguity while retaining user privacy). 

The proposed model more or less exactly matches what is being standardized here. We would just need to tweak the names.

kyerebo:
Agreed the model doesn't differ much besides the actual naming convention, and I wasn't aware that the disambiguation argument had already come up prior!
Considering that, I'm in favor of the `@media (app-context: installed)` name rather than the `@media (installed)` approach.

marcoscaceres:
@kyerebo, we should avoid abbreviations like "app", so maybe we change to "application-context"? Also, in your explainer you had that all iframes can't see the value... I think it's fine for same origin iframes to see the value, as they can reach into their parent or `top` regardless). Does that make sense, or am I missing something?

kyerebo:
Agreed, let's switch to `application-context`.

Ah, you're not missing anything with the iframe access, I was a little over eager here with the privacy measures 😅.

I'll update the explainer to reflect avoiding the abbreviation and only preventing cross-origin iframe access.

marcoscaceres:
@kyerebo, made the changes. Please take a look. 

@kyerebo, also, with my Working Group Chair hat on, it would be great if you could join the working group (Microsoft is already a participant... you just need to click a button): https://www.w3.org/groups/wg/webapps/

benfrancis:
@marcoscaceres wrote:
> I don’t mind either approach, and like the disambiguation argument raised in the explainer, which may address some concerns raised by @benfrancis in previous discussions (at least in a far that clears the “am I installed?” ambiguity while retaining user privacy).

Yes, I like this direction. Thank you for further highlighting the ambiguity issue @kyerebo.

In https://github.com/w3c/manifest/issues/1092#issuecomment-3952165303 I highlighted the following as three orthogonal states:
> 1. display-mode - What display mode the current document is being displayed in, already provided by the display-mode media query
> 2. installed - Whether the current page falls within the navigation scope of an app installed on this system, could be exposed using a DOM property or media query but Apple doesn't want to expose this in Safari due to privacy reasons
> 3. context - Whether the current document is being displayed in a browsing context or an application context - presumably less sensitive and still distinct from the two above

No. 1 is already implemented, Apple would prefer not to expose no. 2, this proposal specifies no. 3.

My only feedback would be that the naming is still slightly ambiguous because under the current specification:
- A web app can technically be both "installed" and using the "browser" display mode
- A web app can be "installed" on the current device whilst also being displayed in a browser tab (if navigation scope is not always captured by the installed app)

It may not be clear whether `application-context: browser` means:
- The web app is not installed
- The web is installed but currently loaded in a browser tab
- The web app is installed and loaded in an application context using the browser display mode

Fundamentally "browser" (a container and display mode) and "installed" (a lifecycle state) are not opposites, they are orthogonal.

~~I would suggest that a less ambiguous naming scheme could be:~~

- ~~`display-context: application`~~
- ~~`display-context: browser`~~

~~Reasons for the term `display-context`:~~
- ~~It is a nice complement to `display-mode`~~
- ~~Used instead of "browsing-context" to avoid confusion after the terminology change proposed in https://github.com/w3c/manifest/pull/1219~~

Edit: See comment below.

---

Another way to reduce the first part of the remaining ambiguity would be to simply retire the "browser" display mode, which I'm not sure is actually implemented anywhere anyway? That still wouldn't solve the second part though.

benfrancis:
Thinking about this more, `display-context: browser` is potentially still ambiguous as long as the "browser" display mode of a web app manifest exists.

Maybe `application-context` _should_ just have a boolean value? Then the following would be possible:

```
@media (application-context) and (display-mode: browser) {

}
```

marcoscaceres:
Yeah, I’m kinda landing in the same place.

kyerebo:
@marcoscaceres Just got approved to join! Thanks for the reminder. 

I think the main gap in retaining it as a Boolean value would be the loss of potential extensibility if it turns out we want to represent other application contexts (perhaps embedded apps?)

That being said, I do think keeping the axis intentionally narrow here with a Boolean value feels a lot more precise. I'm happy with switching back to a Boolean, and keeping the rename to `application-context` to avoid ambiguity.

marcoscaceres:
Yeah... I also kinda what you wrote said or eluded to in the explainer, about developers explicitly having to opt into this (and hopefully think a bit about what it means). I'm not super convinced on the extensibility part, but can see an "I told you so" situation in the future 🤭. 

Ok, let's let this cook over the weekend. But if no one objects, this feels pretty good to me. 

Let's shop this around a little more and see what other WG members think.

kyerebo:
Thank you for making those changes, this is shaping up nicely!
I had a question about the spec placement. Since this is introducing a CSS media feature, is there a reason this is being defined in the manifest spec rather than in the CSS Media Queries spec?
It seems like this might potentially span both areas. If that’s the case, would it make sense to also reflect changes (or at least reference this) in the CSS spec?
Curious if there’s prior discussion or rationale here that I missed.

marcoscaceres:
@kyerebo we generally do it here because we understand the context/usage better than the CSS WG - plus a lot of us who work on this are not members of the CSS WG. 

However, once it stabilizes, we can certainly upstream it to the CSS WG for maintenance.  

At the same time, we would get the CSS WG members to review this before merging it.

dmurph:
Generally LGTM. I would say we could mention that the information it reveals isn't 'new' as a dev can still infer this via the `display_mode` media query, although that information is more imperfect and quite a bit more annoying to clean up.

dmurph (approved):

benfrancis:
@marcoscaceres wrote:
> @benfrancis wrote:
>> Maybe application-context should just have a boolean value?

> Yeah, I’m kinda landing in the same place.

Am I missing something or does the version being reviewed still have two string values rather than a boolean?

@dmurph wrote:
> I would say we could mention that the information it reveals isn't 'new' as a dev can still infer this via the display_mode media query, although that information is more imperfect and quite a bit more annoying to clean up.

It is new in the sense that you will now be able to differentiate between `fullscreen` in a browsing context and `fullscreen` in an application context, and (in theory) between a browsing context and an application context with the `browser` display mode "browser" (using the current "browsing context" terminology). I think the current "Note: Independent of `display-mode`" explains this well, though I would maybe say "what context is this running in?" rather than "what application context is this running in?".

kyerebo:
> @kyerebo we generally do it here because we understand the context/usage better than the CSS WG - plus a lot of us who work on this are not members of the CSS WG.
> 
> However, once it stabilizes, we can certainly upstream it to the CSS WG for maintenance.
> 
> At the same time, we would get the CSS WG members to review this before merging it.

Thanks for the clarification, that makes sense to me. It'd be great to have a CSS WG member review this to avoid future friction when trying to upstream it.

> @marcoscaceres wrote:
> 
> > @benfrancis wrote:
> > > Maybe application-context should just have a boolean value?
> > > Yeah, I’m kinda landing in the same place.
> 
> Am I missing something or does the version being reviewed still have two string values rather than a boolean?

I'm also still seeing the enumerated media feature, could we have the spec text reflect the boolean media feature version (e.g. `@media (application-context)` or `@media not (application-context)`)?

dmurph:
I was told at some point that enum media queries are preferred over booleans, for future proofing, but idk where I was told that I could be misremembering. We can also define the boolean evaluation of this media query, I believe, while keeping it an enum

dmurph:
> Generally LGTM. I would say we could mention that the information it reveals isn't 'new' as a dev can still infer this via the `display_mode` media query, although that information is more imperfect and quite a bit more annoying to clean up.

this was in reference to the privacy/security considerations

benfrancis:
@dmurph wrote:
> I was told at some point that enum media queries are preferred over booleans, for future proofing, but idk where I was told that I could be misremembering. We can also define the boolean evaluation of this media query, I believe, while keeping it an enum

I see. I think the challenge is just what strings to use that don't just add to the confusion!

I think `display-context: application` and `display-context: browser` would make sense _if_ the browser display mode was deprecated in app manifests. (I was originally going to suggest `browsing-context: application` and `browsing-context: browser`, but given the proposed changes in terminology to refer to "top-level traversable" instead of "browsing context" I'm not sure if that might be confusing.)

Otherwise, looking at other existing media queries I see "none" is often used as an off state so:

`application-context: none` and `application-context: application` could work (evaluating to `true` if `application` and `false` if `none`)?

I agree with @kyerebo that it might be a good idea to get input from the CSS WG so it's something they're happy with from a consistency and extensibility point of view etc.

> this was in reference to the privacy/security considerations

Ah OK I understand. Yes it doesn't expose much more information than was already available from a privacy point of view, since `standalone` and `minimal-ui` display modes are currently only possible inside an application context so the application context is detectable most of the time.

christianliebel (approved):

dmurph:
Manifest spec eu/us meeting notes:

  * Dan: having a bool that could be an enum often is something I regret when coding, as in the future for established APIs this hinders adding new values, and often the 'enum-label: enum value' is a more readable description than a bool.  
  * Dan  
    * Don't care THAT much. Dan has slight preference for enum and no bool, unless.. maybe 'installed'  
    * application-context with an enum feels the most future proof.  
  * Christian  
    * agreed on enums over bools can be nice  
    * also don't care that much  
    * why not installed?  
  * explanation from [explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ApplicationContextMediaFeature/explainer.md#alternatives-considered)  
    * The name installed suggests a statement about global installation state. A developer might reasonably expect (installed) to be true if the app is installed on the device, even when viewed in a browser tab. In reality, the feature only matches when running *inside* an installed app window. The name application-context makes this distinction explicit, and describes the current context, not a global property.  
    * Dan: seems reasonable?  
      * this also means that it is guaranteed to be interop, as apple didn't want to reveal installed in the browser tab, and chrome.. could? but we likely want to align, and teh getInstalledRelatedApps() api is still a way to detect if the app is installed.  
    * Christian \- sounds good

marcoscaceres:
Requested CSSWG review of the `application-context` media feature: w3c/csswg-drafts#14125

benfrancis:
@marcoscaceres Why does that proposal still use the term `installed`? I thought we had established that term is confusing in this context since whether or not the application is "installed" is an orthogonal state to whether it is currently being displayed in an application context.

From above:
> The name installed suggests a statement about global installation state. A developer might reasonably expect (installed) to be true if the app is installed on the device, even when viewed in a browser tab. In reality, the feature only matches when running inside an installed app window. 

I would again suggest the following would be clearer:
- application-context: application
- application-context: none

dmurph:
@benfrancis said:
> ....
> I would again suggest the following would be clearer:
> 
> * application-context: application
> * application-context: none

Manifest editor meeting notes:
- @dmurph : This is fine with me
- @christianliebel : This is fine with me
- I guess feedback for @kyerebo and FYI for @robpaveza - wdyt? @marcoscaceres WDYT?

kyerebo:
I'm fine with 

> application-context: application
> application-context: none

being the agreed upon enum values. It addresses the concern of lack of extensibility while removing the ambiguity of "installed" that switching to an enum brought back.
Looks good to me!

(I'll update Explainer text to match once changes are made here so that we don't have conflicting documentation.)

dmurph:
Requested feedback on the css mailing list here: https://lists.w3.org/Archives/Public/www-style/2026Jul/0014.html

marcoscaceres:
Following a security review during WebKit implementation, I've moved `application-context` to an ancestor-chain origin check: a document observes the application context only if it and every ancestor up to the top-level traversable are same origin. This closes the case where a frame same-origin with the top, but nested inside a cross-origin frame (A→B→A′), would otherwise be observable. The spec text and WebKit implementation now reflect this.

Separately: `display-mode` currently exposes the `standalone` bit to cross-origin frames with no origin gating. I'll file a follow-up to consider tightening it for consistency — flagging here since it came up in the same review. @benfrancis @dmurph @kyerebo — shout if the ancestor-chain model looks wrong to you.

kyerebo:
The review request in the CSSWG thread seems to have landed on a `display-mode: application` proposal. Wanted to get a pulse check on if we're okay with this becoming the normative solution rather than `application-context` MQ.

While I do have concerns around continuing to add non-uniform values to `display-mode`, there are also advantages in not having to create a new media query and working within a space developers are already familiar with. We'd also have to make sure multiple `display-mode` values can match at the same time if we go this route.

